PR #4, opened 2018-07-16
AMDVLK issue#39 - dxvk: F.E.A.R. 3 black screen in menu

linqunzju:
root cause: we widen the bool type for all global variables, but we didn't widen the bool value in the initializer.


amdvlk-admin:

PR #6, opened 2018-07-30
Fix LLPC issue#5 - Structs with no members segfault

linqunzju:


amdvlk-admin:

PR #7, opened 2018-07-31
Fix crash in CTS dEQP-VK.spirv_assembly.instruction.graphics.indexing.input.struct.opaccesschain_u64_vert

linqunzju:
root cause: Constantexpr isn't handled in pass SpirvLowerAccessChain, and it causes some chained getelementptr isn't lowered
solution: Check constantexpr in SpirvLowerAccessChain::TryToCoalesceChain, and translate constantexpr to normal instruction if found.


amdvlk-admin:

PR #45, opened 2019-04-01
Rationalize Builder methods, opcodes and call names

trenouf:
Previously, the opcodes and call names for recorded builder calls were
inconsistent with the builder method names. The method names do not
have a prepended category, which is right, because the division of
BuilderImpl into categories is an internal implementation detail.

This commit similarly changes the opcodes and call names to match --
there is no need for a prepended category, because the categories are an
internal implementation detail of BuilderImpl.

So, for example, to "create a load of a buffer descriptor", the method
name already used that natural order of words:
  Builder::CreateLoadBufferDesc.
I have now changed the opcode to match:
  BuilderRecorder::Opcode::LoadBufferDesc
and I have changed the call name to match:
  llpc.call.load.buffer.desc.

Change-Id: Id37bb3a022d9d8a4c737d27d60d4e69cc3ff6902

### DIFF
--- a/builder/llpcBuilderRecorder.cpp
+++ b/builder/llpcBuilderRecorder.cpp
@@ -46,26 +46,26 @@ StringRef BuilderRecorder::GetCallName(
     {
     case Opcode::Nop:
         return "nop";
-    case Opcode::DescLoadBuffer:
-        return "desc.load.buffer";
-    case Opcode::DescLoadSampler:
-        return "desc.load.sampler";
-    case Opcode::DescLoadResource:
-        return "desc.load.resource";
-    case Opcode::DescLoadTexelBuffer:
-        return "desc.load.texel.buffer";
-    case Opcode::DescLoadFmask:
-        return "desc.load.fmask";
-    case Opcode::DescLoadSpillTablePtr:
-        return "desc.load.spill.table.ptr";
-    case Opcode::MiscKill:
-        return "misc.kill";
-    case Opcode::MiscReadClock:
-        return "misc.read.clock";
-    case Opcode::DescWaterfallLoop:
-        return "desc.waterfall.loop";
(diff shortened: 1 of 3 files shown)


JacobHeAMD:
Hi @trenouf , sorry that there is still some problem with the parser script and jenkins pipeline, the failed test cases are:
Failing Tests (11):
    LLPC_SHADERTEST :: shaderdb/ObjPushConst_TestSpillToMemory_lit.vert
    LLPC_SHADERTEST :: shaderdb/ObjResource_TestAlias_lit.frag
    LLPC_SHADERTEST :: shaderdb/ObjStorageBlock_TestDirectIndex_lit.frag
    LLPC_SHADERTEST :: shaderdb/ObjStorageBlock_TestIndirectIndex_lit.frag
    LLPC_SHADERTEST :: shaderdb/ObjUniformBlock_TestDirectIndex_lit.frag
    LLPC_SHADERTEST :: shaderdb/ObjUniformBlock_TestIndirectIndex_lit.frag
    LLPC_SHADERTEST :: shaderdb/OpKill_TestFunctionDynamic_lit.frag
    LLPC_SHADERTEST :: shaderdb/OpKill_TestGeneral_lit.frag
    LLPC_SHADERTEST :: shaderdb/PipelineCs_TestDynDescNoSpill_lit.pipe
    LLPC_SHADERTEST :: shaderdb/PipelineCs_TestDynDescSpill_lit.pipe
    LLPC_SHADERTEST :: shaderdb/PipelineCs_TestInlineConstDirect_lit.pipe

amdrexu:
Since you change the call name, the lit tests must have to be modified to update the reference results.

JacobHeAMD:
This change requires to change the shader tests in xgl, has to go through internal git branch. The shader tests will be moved into llpc from xgl soon.

PR #83, opened 2019-05-23
Fix MGPU regression with scratch bounds enabled

perlfu:
In a system with a mixture of GFX9/8/7 parts the scratch bounds
option is only enabled for GFX9 parts.  This causes an error
initialising the compiler as the options are inconsistent.
Fix this by ignoring the scratch bounds options when generating
the options hash.

### DIFF
--- a/context/llpcCompiler.cpp
+++ b/context/llpcCompiler.cpp
@@ -1489,6 +1489,7 @@ MetroHash::Hash Compiler::GenerateHashForCompileOptions(
         cl::LogFileOuts.ArgStr,
         cl::EnableShadowDescriptorTable.ArgStr,
         cl::ShadowDescTablePtrHigh.ArgStr,
+        "amdgpu-scratch-bounds-checking",
     };
 
     std::set<StringRef> effectingOptions;


amdvlk-admin:
### Test summary for commit 138fc5e7e514eb79bf9c6ea1cefcc67c134dbc2e

<li>PAL: dev 7438ae21a2e8e3c782f82745249da539a19c193e</li>
<li>SPVGEN: dev 5378c7618a2a3bf60ba01a86db980930e8581146</li>
<li>XGL: dev 9490055b71ec89be68c1ed4a3a847569c0ac3cbf</li>
<li>LLVM: amd-vulkan-dev f41e1a873108a371ae5574d518c1ee6eb3814cee</li>
</ul></details>

<li>Passed: 893/893 (100.0%)</li>
<li>Failed: 0/893 (0.0%)</li>
</ul></details>

<li>Passed: 366444/366444 (100.0%)</li>
<li>Failed: 0/366444 (0.0%)</li>
</ul></details>


</ul></details>

context/llpcCompiler.cpp
```diff
@@ -1489,6 +1489,7 @@ MetroHash::Hash Compiler::GenerateHashForCompileOptions(
         cl::LogFileOuts.ArgStr,
         cl::EnableShadowDescriptorTable.ArgStr,
         cl::ShadowDescTablePtrHigh.ArgStr,
+        "amdgpu-scratch-bounds-checking",
```
linqun: I don't think we should ignore option "amdgpu-scratch-bounds-checking" when generate hash code, because it will affect final GPU ISA code. I think we should enable this option in general and check the gfxip within LLVM back-end.

another choice is implement amdgpu-scratch-bounds-checking as an target option, then we can choose it in run time, instead of initial time.

perlfu: The LLVM backend code is target agnostic, it can be applied to any GFX revision.  So I don't think that's the correct place for the option.

A target option might work, I will investigate this.
However I wonder if another solution is to relax the check in ICompiler::Create to only compare option hashes for compilers of the same gfxIp?
I notice that GenerateHashForCompileOptions is only used in this context, so the question is whether the intention is to have compilers that produce identical output (which they won't for different GFX revision) or simply compatible output.

linqun: - This hash is used in shader cache. so we can relex the check for same gfxip.
- but I am afraid we can't set different value for same option.

perlfu: I can see the problem in detail now
I hadn't appreciated the static reuse of the compiler options between all compiler instances.
I think this issue will need to be fixed by further changes to the feature implementation in the backend and frontend.
I'll close this merge request.

PR #97, opened 2019-06-21
Wrong compile in ObjOutput_TestVsMatrix*.

xuechen417:
- Add a global llvm option llvm::cl::DisableNullFragShader to control whether we need to add a null fragment shader for standalone compiler.For .pipe file, this option is always set to FALSE, otherwise its value can be set by user or be TRUE by default.
- Add testing pattern in ObjOutput_TextVsMatrix*.vert

### DIFF
--- a/patch/llpcPatchNullFragShader.cpp
+++ b/patch/llpcPatchNullFragShader.cpp
@@ -45,6 +45,20 @@
 using namespace Llpc;
 using namespace llvm;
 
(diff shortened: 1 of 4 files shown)


amdvlk-admin:

JacobHeAMD:
ok to test

JacobHeAMD:
test this please

PR #111, opened 2019-07-11
Separated descriptor load into three parts

trenouf:
This commit separates a descriptor load into three parts:
1. Get pointer to descriptor
2. Optionally index it
3. Load the descriptor from the pointer.

This separation is done both in the builder interface and in the
llpc.descriptor.* calls lowered by PatchDescriptorLoad.

This only affects image-related descriptors (image, fmask, texel buffer,
fmask), not buffer descriptors.

I have done this to allow a future commit to change the SPIR-V reader to
use builder calls directly, instead of needing the LowerImageOp pass
afterwards. In most cases the SPIR-V reader would be able to trace back
from an image op to its descriptor's set, binding and index. But, in the
corner case of a pointer to (array of) descriptor being passed as a
function parameter, it is not possible for the SPIR-V reader to do that.

The implementation in PatchDescriptorLoad currently assumes that
everything is inlined by then, so it can trace back from the "descriptor
load from pointer" back to the "get pointer to descriptor", and
implement them together, including the case of optimizing a "dynamic
descriptor" (one that is in the top-level user data) or an immutable
sampler (constant data provided to the compiler). The plan is to remove
this assumption of inlining everything in a future commit.

Change-Id: I1275ecc1774afca4d7361479445cee0b900852f5

### DIFF
--- a/builder/llpcBuilder.cpp
+++ b/builder/llpcBuilder.cpp
@@ -357,23 +357,62 @@ PointerType* Builder::GetBufferDescTy(
 }
 
 // =====================================================================================================================
-// Get the type returned by CreateLoadResourceDesc.
-VectorType* Builder::GetResourceDescTy()
+// Get the type of an image descriptor
+VectorType* Builder::GetImageDescTy()
 {
     return VectorType::get(getInt32Ty(), 8);
 }
 
 // =====================================================================================================================
-// Get the type returned by CreateLoadTexelBufferDesc.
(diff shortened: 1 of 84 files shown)


trenouf:
This is based on pr110. I will rebase once pr110 is merged.

JaxLinAMD:
@trenouf , please resolve conflicts.

JaxLinAMD:
@trenouf , looks like there are new conflicts.

PR #120, opened 2019-07-21
Refine LLPC

amdrexu:
- Set Api, Type, LdsSize metadata.
- Conditionally report non-zero metadata.
- Conditionally set stream-out table in user data register.


JaxLinAMD:
@amdrexu , also conficts

amdrexu:
Rebased. I think you have already ported my internal changes to GitHub. So the commit actually has no effective code changes.

PR #147, opened 2019-08-19
Builder::CreateImageStore: moved pTexel to the first arg

trenouf:
Two reasons:
1. Having the value-to-write as the first arg brings it into line with
   the forthcoming shader output write methods.
2. Having the optional value (pMipLevel) at the end of the args helps
   the forthcoming change to autogenerate BuilderRecorder and
   BuilderReplayer.

Change-Id: I13938725b771b4855ad8ab2ccf00c1ba28193253


JacobHeAMD:
retest this please

JacobHeAMD:
retest this please

trenouf:
I will land this at the same time as PR #149 (when PR #150 testing passes).

PR #189, opened 2019-09-09
Consistently set FMF in builder

trenouf:
This changes the SPIR-V reader to consistently set the fast math flags
in Builder when generating an instruction.

Code that uses non-builder functions to generate IR, or uses a builder
function that does not honor FMF (e.g. CreateMinNum), needs to call
SPIRVToLLVM::setFastMathFlags to set the FMF on the instruction from the
builder.

However, in this commit, I did not want to introduce any actual
generated code changes, so in several places I also added a
clearFastMathFlags and a TODO: that we need to remove it and fix up the
tests. That should happen in future commits.

Change-Id: I0a2d7180de101eee2c6f8ee424a7976a1ccda68e


translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -2551,6 +2558,9 @@ Value* SPIRVToLLVM::addLoadInstRecursively(
 {
     LLPC_ASSERT(pLoadPointer->getType()->isPointerTy());
 
+    // TODO: Remove this and fix lit tests to cope with fast math flags.
+    getBuilder()->clearFastMathFlags();
```
linqun: I think we should hide it in BuilderImp instead of set it explicitly in SPIRV Reader.
If you expose more and more state, builder will be difficult to use for client.

trenouf: IRBuilder already has FMF state, which it uses when generating an FP operation. I want our Builder to work consistently with that. These TODOs are temporary; once we remove them, then the parts of the spir-v reader that use Builder will consistently get the FMF that the spir-v reader sets up to correspond to the overall mode of the shader and any decorations on the instruction. Then, once the spir-v reader uses Builder everywhere, it will all work consistently, and be easy to use for the Builder client.

linqun: I think FMF state should only affect FP operation, so it should not affect non-FP operations regardless it is from IRBuilder or LLPC Builder.
I think it is better if LLPC builder can ignore the fast math flag for non-FP operations.
I am OK to approve this change first if it is not easy to implement it in builder.

trenouf: I believe our builder does ignore FMF for non-FP operations. Maybe I inserted this TODO and clearFastMathFlags in more places than I needed to. Is that what you're seeing? (I just inserted it everywhere that uses builder to be safe, rather than only places that use builder for an FP operation.)

trenouf:
That looks like a spurious failure.

trenouf:
retest this please

trenouf:
retest this please

trenouf:
Rebased.

translator/lib/SPIRV/SPIRVReader.cpp
trenouf: Hi Qun. Please see subsequent commits #192 and #193, which I want to land straight after this one. Can you now approve this one?

amdrexu: Qun is on vacation. You can make updates if he has some suggestions. I approve this now. You can continue your work.

trenouf: Thanks Rex.

JacobHeAMD:
retest this please

trenouf:
V2: Set FMF for BuilderImpl in BuilderReplayer.

trenouf:
This needs to wait for #223.

trenouf:
V3: Don't set most FMF if DenormFlushToZero is on.

trenouf:
This looks like a spurious failure:
```
Test case 'dEQP-VK.renderpass.suballocation.attachment_allocation.grow.67'..
  ResourceError (Failed to allocate memory during test execution)

DONE!

Test run totals:
  Passed:        165992/282837 (58.7%)
  Failed:        1/282837 (0.0%)
  Not supported: 116844/282837 (41.3%)
  Warnings:      0/282837 (0.0%)
Test run was ABORTED!
```

trenouf:
retest this please

PR #157, opened 2019-08-23
Derivative operations in Builder

trenouf:
This commit adds the derivative operations to Builder, and switches the
SPIR-V reader to use Builder for them.

I also removed an unnecessary wqm intrinsic call in the middle of the
calculation.

Change-Id: Ie71ba1d9025ffc9e0c55509c80c47b4bcbcdb401

PR #219, opened 2019-09-16
Reinstated extra llvm.amdgcn.wqm in derivative code

trenouf:
In PR #157 "Derivative operations in Builder", I also removed what I
believe is an extra unnecessary llvm.amdgcn.wqm call in the middle of a
derivative calculation. However that caused much worse performance in
one pipeline in a particular game because it started doing lots of sgpr
spilling. Possibly having fewer wqm intrinsics and fewer bitcasts
allowed some pass more freedom to move other code around, which
accidetntally increased sgpr pressure.

This commit reinstates the extra llvm.amdgcn.wqm, to get the same final
ISA as before the above change.

Change-Id: I83467c4f05f42ab6626f13f5c586b43c2244ef12

### DIFF
--- a/builder/llpcBuilderImplMisc.cpp
+++ b/builder/llpcBuilderImplMisc.cpp
@@ -174,6 +174,7 @@ Value* BuilderImplMisc::CreateDerivative(
                                                                      getInt32(15),
                                                                      getTrue()
                                                                   });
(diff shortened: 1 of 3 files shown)


trenouf:
V2: Fixed lit tests.

PR #223, opened 2019-09-18
Builder trigonometric, exp, log, inverse sqrt

trenouf:
Reinstate the change that was reverted in PR218.

This commit switches the SPIR-V translator to use the above Builder
operations.

Change-Id: I46a50f1e1c61a9e28c7fd9408a3a4506d2e20f2e


trenouf:
This cannot be landed yet because we're still having some internal test problems with it.

trenouf:
V2: Do not remove the tan functions, still used when instcombine turns
        sin/cos into a call to the @tan library function.

trenouf:
It looks like this does fix our internal test problem. I will land it once review and test are complete.

PR #255, opened 2019-10-03
Make the LLPC_ASSERT macro variadic

kuhar:
This patch introduces type punning utils. They replace a few most glaring strict aliasing violations.

In C++, memory is roughly 'dynamically strongly typed', i.e., each memory read must be of type compatible with the type of the last write (if any) -- known as the strict aliasing rules. An exception from this is the `char` type that can be used access any valid chunk of memory.

As a consequence, we have to be careful not to read memory using incompatible types. For example, we cannot perform pointer casts to load an int from a float variable.

### DIFF
--- a/util/llpcDebug.h
+++ b/util/llpcDebug.h
@@ -38,7 +38,7 @@
 #include <cassert>
 
 // Debug assertion: generic
-#define LLPC_ASSERT(_expr)      assert(_expr)
+#define LLPC_ASSERT(...)        assert(__VA_ARGS__)
 // Debug assertion: not implemented
 #define LLPC_NOT_IMPLEMENTED()  assert(0 && "Not implemented!")
 // Debug assertion: should never be called
@@ -48,7 +48,7 @@
 
 #else
 
-#define LLPC_ASSERT(_expr)            ((void)0)
+#define LLPC_ASSERT(...)              ((void)0)
 #define LLPC_NOT_IMPLEMENTED()        ((void)0)
 #define LLPC_NEVER_CALLED()           ((void)0)
 #define LLPC_NOT_TESTED()             ((void)0)


amdvlk-admin:

util/llpcUtil.h
```diff
@@ -158,42 +163,125 @@ inline uint32_t Log2(
...
-        if (pStructHeader->type == type)
-        {
-            break;
-        }
-        else
-        {
-            pStructHeader = pStructHeader->pNext;
-        }
+      VkStructureType foundType;
+      pCurrent = FromRawBytes(pCurrent, pCurrent + sizeof(VkStructureType), foundType);
```
linqun: VkStructHeader isn't defined by LLPC, we should not assume the layout in it. i.e. please use original code.

util/llpcInternal.cpp
```diff
@@ -453,8 +453,11 @@ bool IsSpirvBinary(
     bool isSpvBinary = false;
     if (pShaderBin->codeSize > sizeof(SpirvHeader))
     {
-        const SpirvHeader* pHeader = reinterpret_cast<const SpirvHeader*>(pShaderBin->pCode);
-        if ((pHeader->magicNumber == MagicNumber) && (pHeader->spvVersion <= spv::Version) && (pHeader->reserved == 0))
+        const char* pCode = static_cast<const char*>(pShaderBin->pCode);
+        SpirvHeader maybeHeader;
+        FromRawBytes(pCode, pCode + pShaderBin->codeSize, maybeHeader);
```
linqun: Please initialize maybeHeader to 0

util/llpcInternal.cpp
```diff
@@ -470,10 +473,15 @@ bool IsLlvmBitcode(
 {
     static uint32_t BitcodeMagicNumber = 0xDEC04342; // 0x42, 0x43, 0xC0, 0xDE
     bool isLlvmBitcode = false;
-    if ((pShaderBin->codeSize > 4) &&
-        (*reinterpret_cast<const uint32_t*>(pShaderBin->pCode) == BitcodeMagicNumber))
+    if (pShaderBin->codeSize > 4)
     {
-        isLlvmBitcode = true;
+        const char *pCode = static_cast<const char*>(pShaderBin->pCode);
+        uint32_t code;
```
linqun: please initialize code to 0.

util/llpcInternal.cpp
```diff
@@ -616,18 +624,22 @@ Result VerifySpirvBinary(
...
+    const char* pCodePos = pCodeBegin + sizeof(SpirvHeader);
+    constexpr size_t wordSize = sizeof(uint32_t);
 
-    while (pCodePos < pEnd)
+    while (pCodePos < pCodeEnd)
     {
-        Op opCode = static_cast<Op>(pCodePos[0] & OpCodeMask);
-        uint32_t wordCount = (pCodePos[0] >> WordCountShift);
+        uint32_t currWord;
+        FromRawBytes(pCodePos, pCodeEnd, currWord);
```
linqun: please check whether it generated additional call in generated code in release build.
I don't want this change causes any compiler performance drop.

util/llpcDebug.h
```diff
@@ -38,7 +38,7 @@
 #include <cassert>
 
 // Debug assertion: generic
-#define LLPC_ASSERT(_expr)      assert(_expr)
+#define LLPC_ASSERT(...)        assert(__VA_ARGS__)
```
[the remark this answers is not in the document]
kuhar: C preprocessor doesn't really know about C++ constructs and treats multiple template arguments as multiple preprocessor parameters.
For example:
`LLPC_ASSERT(pair<int, float>::something)` would be treated as 2 parameters to the assert macro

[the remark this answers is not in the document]
kuhar: Using variadic macros is a standard way to handle macros like this from what I know.

patch/llpcConfigBuilderBase.cpp
```diff
@@ -376,13 +376,22 @@ void ConfigBuilderBase::WritePalMetadata()
...
     msgpack::MapDocNode registers = m_pipelineNode[".registers"].getMap(true);
-    size_t sizeInDword = m_configSize / sizeof(uint32_t);
+    const size_t sizeInDwords = m_configSize / sizeof(uint32_t);
     // Configs are composed of DWORD key/value pair, the size should be even
-    LLPC_ASSERT((sizeInDword % 2) == 0);
-    for (size_t i = 0; i < sizeInDword; i += 2)
+    LLPC_ASSERT((sizeInDwords % 2) == 0);
+    static_assert(sizeof(*m_pConfig) == 1, "Expected a byte buffer");
+    const uint8_t* pCurrentConfig = m_pConfig;
+    const uint8_t* const pConfigEnd = m_pConfig + m_configSize;
```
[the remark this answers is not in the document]
kuhar: Is `m_pConfig` a byte buffer or a word buffer?
Changing the pointer type here is a very confusing thing to me for the following reasons:
- It raises questions about legality of the code -- reading memory as a different type than the last written can exhibits undefined behavior in plain C++, and is legal in certain scenarios. In the case of LLPC, it seems legal only because it is compiled with `-fno-strict-aliasing`, but there is no way of telling that without analyzing the build scripts.
- It raises doubts about other low-level issues such as alignment, type sizes, etc.

I think that making type punning more explicit here solves the above issues and reduces mental overhead required to understand it convince yourself that the code is doing what was it was intended to.

util/llpcInternal.cpp
kuhar: All the calls are gone on my machine.

util/llpcDebug.h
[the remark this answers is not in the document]
linqun: got it, thank you for your explain.

patch/llpcConfigBuilderBase.cpp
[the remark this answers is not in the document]
linqun: m_pConfig is a word buffer. but we need change multiple files/functions when change the type from uint8_t* to uint32_t*

util/llpcInternal.cpp
linqun: got it. thanks!

kuhar:
As a high-level comment, I find the C/C++ strict aliasing to have many benefits, mainly making code easier to understand.
Here are a couple of explanations better than mine that talk about it:
- https://blog.regehr.org/archives/1307
- https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8#file-whatisstrictaliasingandwhydowecare-md

Reading the source, I understand that LLPC was developed for compiling with strict aliasing disabled.   If the intention is to keep it this way and do not plan transitioning then you can close this PR.

amdrexu:
> Reading the source, I understand that LLPC was developed for compiling with strict aliasing disabled. If the intention is to keep it this way and do not plan transitioning then you can close this PR.

Please drop other changes involve FromRawBytes and TypePunTo. We do develop LLPC with non-strict aliasing. We don't want to make things complex by introducing additional handlers. After internal discussion, we will change pConfig from uint_8* to uint32_t* to avoid code confusion. The register configuration is acutally a DWORD pari<key, value>. We will make changes to correct this and make codes clear.

kuhar:
Thanks for the explanation.

I reverted the changes related to strict aliasing, leaving only the LLPC_ASSERT macro changes.

amdvlk-admin:
ok to test

amdvlk-admin:
test this please

amdvlk-admin:
### Test summary for commit c3f298e35e5728b5127f5aa0508e95c3d2fa82a5

<li>PAL: dev 3e28ea331cf8e7b0e5733e5bd7f7d16582c603c5</li>
<li>SPVGEN: dev 2f31d1170e8a12a66168b23235638c4bbc43ecdc</li>
<li>XGL: dev 3252b6223947f9fc67399e0798b1062983925fce</li>
<li>LLVM: amd-vulkan-dev 21b028dc97ee9b0c6629ff83c2924b3b80c5d6e7</li>
</ul></details>

<li>Passed: 927/927 (100.0%)</li>
<li>Failed: 0/927 (0.0%)</li>
</ul></details>

<li>Passed: 68331/68331 (100.0%)</li>
<li>Failed: 0/68331 (0.0%)</li>
</ul></details>

<li>Passed: 68363/68363 (100.0%)</li>
<li>Failed: 0/68363 (0.0%)</li>
</ul></details>


</ul></details>

PR #257, opened 2019-10-10
Reinstate -enable-dynamic-loop-unroll option

tcorringham:
Bumped LLPC revision to 37, and reinstated the deprecated
-enable-dynamic-loop-unroll option for previous versions to
provide backwards compatibilty with older versions of xgl.

Note only the option is reinstated, and is a no-op.

### DIFF
--- a/context/llpcCompiler.cpp
+++ b/context/llpcCompiler.cpp
@@ -127,6 +127,11 @@ opt<uint32_t> ShadowDescTablePtrHigh("shadow-desc-table-ptr-high",
                                      desc("High part of VA for shadow descriptor table pointer"),
                                      init(2));
 
+#if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 37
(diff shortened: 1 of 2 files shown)


context/llpcCompiler.cpp
```diff
@@ -127,6 +127,11 @@ opt<uint32_t> ShadowDescTablePtrHigh("shadow-desc-table-ptr-high",
                                      desc("High part of VA for shadow descriptor table pointer"),
                                      init(2));
 
+#if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 37
+// -enable-dynamic-loop-unroll: Enable dynamic loop unroll. (Deprecated)
+opt<bool> EnableDynamicLoopUnroll("enable-dynamic-loop-unroll", desc("Enable dynamic loop unroll (deprecated)"), init(false));
```
amdrexu: This line might exceed 120 characters.

PR #305, opened 2019-10-29
PatchInOutImportExport: increase code readability

nhaehnle:
The original code wasted 60% of vertical space, plus returning immediately
from all the case statements makes the logic of the function clearer.

### DIFF
--- a/patch/llpcPatchInOutImportExport.cpp
+++ b/patch/llpcPatchInOutImportExport.cpp
@@ -2363,66 +2363,36 @@ Value* PatchInOutImportExport::PatchVsBuiltInInputImport(
     uint32_t     builtInId,     // ID of the built-in variable
     Instruction* pInsertPos)    // [in] Where to insert the patch instruction
 {
-    Value* pInput = UndefValue::get(pInputTy);
-
     auto& entryArgIdxs = m_pContext->GetShaderInterfaceData(ShaderStageVertex)->entryArgIdxs.vs;
 
     switch (builtInId)
     {
     case BuiltInVertexIndex:
-        {
-            pInput = m_pVertexFetch->GetVertexIndex();
-            break;
-        }
+        return m_pVertexFetch->GetVertexIndex();
     case BuiltInInstanceIndex:
-        {
-            pInput = m_pVertexFetch->GetInstanceIndex();
-            break;
-        }
+        return m_pVertexFetch->GetInstanceIndex();
     case BuiltInBaseVertex:
-        {
-            pInput = GetFunctionArgument(m_pEntryPoint, entryArgIdxs.baseVertex);
-            break;
-        }
+        return GetFunctionArgument(m_pEntryPoint, entryArgIdxs.baseVertex);
     case BuiltInBaseInstance:
-        {
-            pInput = GetFunctionArgument(m_pEntryPoint, entryArgIdxs.baseInstance);
-            break;
-        }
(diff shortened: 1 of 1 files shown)


amdvlk-admin:

patch/llpcPatchInOutImportExport.cpp
```diff
@@ -2363,66 +2363,35 @@ Value* PatchInOutImportExport::PatchVsBuiltInInputImport(
...
-            break;
-        }
-    case BuiltInDeviceIndex:
-        {
-            auto pPipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo*>(m_pContext->GetPipelineBuildInfo());
-            pInput = ConstantInt::get(m_pContext->Int32Ty(), pPipelineInfo->iaState.deviceIndex);
-            break;
-        }
+        return GetSubgroupLocalInvocationId(pInsertPos);
+    case BuiltInDeviceIndex: {
```
amdrexu: I think {}  is not necessary.

patch/llpcPatchInOutImportExport.cpp
```diff
@@ -2363,66 +2363,35 @@ Value* PatchInOutImportExport::PatchVsBuiltInInputImport(
...
+        auto pPipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo*>(m_pContext->GetPipelineBuildInfo());
+        return ConstantInt::get(m_pContext->Int32Ty(), pPipelineInfo->iaState.deviceIndex);
+        break;
```
amdrexu: This break is useless.

patch/llpcPatchInOutImportExport.cpp
nhaehnle: It's needed for the auto local variable.

patch/llpcPatchInOutImportExport.cpp
nhaehnle: Yes, going to fix this.

nhaehnle:
Thanks for taking a look. PAL doesn't consistently follow the single-exit either, and quite frankly, if you need a breakpoint here at the end of the function, you may as well set it at the function entry.

patch/llpcPatchInOutImportExport.cpp
amdrexu: Could you follow the switch-case style when adding {} for this case branch? Thanks.

```
case BuiltInDeviceIndex:
    {
         ...
    }
```

amdrexu:
Initially, PAL strictly follows "no early return" rule. Later, admittedly, some functions have several special cases that need early returns, so this rule is relaxed. In my opinion, I prefer early return in those cases:
(1) Early returns are grouped and are placed at the beginning of a function. Thus, we can see those special cases.
(2) Early returns are in each case of switch or if-else branches uniformly and consistently. And each code piece is simple.

For debugging, such case of using early return is annoying:
```
func()
{
    <large codes>
    early return
    <large codes>
     early return
    <large codes>
    <breakpoint> <----debug
    return
}
```
If someone is unfamiliar with this function, the breakpoint is possibly not hit. If he sets the breakpoint at the beginning of this function, he might walk a long way to reach the code point he expects.

patch/llpcPatchInOutImportExport.cpp
nhaehnle: It makes the code worse, and one day you should change your stance on this, but for now I guess it's the lesser of two evils...

JaxLinAMD:
retest this please

PR #310, opened 2019-10-30
llpcInternal: Refactor EmitCall in terms of the IRBuilder<> interface

nhaehnle:
Add a new variant of the EmitCall helper that takes an IRBuilder<>
reference, which is conveniently usable in all contexts that use a builder
and avoids the need to manually distinguish between an insertion point
in the middle vs. at the end of (an incomplete) basic block.

Further benefits of this refactoring include:

- Use a SmallVector for argTys to avoid a heap allocation in most cases.
- Use the builder interface for creating a function call instead of the
  manual Create method; this is the more idiomatic way.
- Do not set call instruction attributes redundantly.

### DIFF
--- a/util/llpcInternal.cpp
+++ b/util/llpcInternal.cpp
@@ -77,8 +77,52 @@ Function* GetEntryPoint(
 }
 
 // =====================================================================================================================
+/// Emits a LLVM function call using the given builder. The callee is built automically based on return
+/// type and its parameters.
+///
+/// @oaram [in] funcName Name of the callee
+/// @param [in] pRetTy Return type of the callee
+/// @param [in] args Arguments to pass to the callee
+/// @param [in] attribs Function attributes
+/// @param [in] builder The IRBuilder to use for creating the call
+CallInst* EmitCall(
+    StringRef                     funcName,
+    Type*                         pRetTy,
+    ArrayRef<Value *>             args,
+    ArrayRef<Attribute::AttrKind> attribs,
+    IRBuilder<>&                  builder)
+{
+    Module* pModule = builder.GetInsertBlock()->getParent()->getParent();
+    Function* pFunc = dyn_cast_or_null<Function>(pModule->getFunction(funcName));
+    if (!pFunc)
+    {
+        SmallVector<Type*, 8> argTys;
+        argTys.reserve(args.size());
+        for (auto arg : args)
+        {
+            argTys.push_back(arg->getType());
+        }
+
+        auto pFuncTy = FunctionType::get(pRetTy, argTys, false);
+        pFunc = Function::Create(pFuncTy, GlobalValue::ExternalLinkage, funcName, pModule);
+
+        pFunc->setCallingConv(CallingConv::C);
+        pFunc->addFnAttr(Attribute::NoUnwind);
+
+        for (auto attrib : attribs)
+        {
+            pFunc->addFnAttr(attrib);
+        }
+    }
+
+    return builder.CreateCall(pFunc, args);
+}
+
+// =====================================================================================================================
 // Emits a LLVM function call (inserted before the specified instruction), builds it automically based on return type
 // and its parameters.
+//
+// NOTE: Prefer the IRBuilder overload of this function where possible.
 CallInst* EmitCall(
     StringRef                     funcName,         // Name string of the function
     Type*                         pRetTy,           // [in] Return type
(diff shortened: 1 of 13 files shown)


nhaehnle:
Note that this commit mostly makes sense in context of at least one other commit (and for now internal development). It would be helpful to change policies and start reviewing commit series instead of individual patches.

util/llpcInternal.cpp
```diff
@@ -77,20 +77,27 @@ Function* GetEntryPoint(
...
+    Type*                               pRetTy,
+    ArrayRef<llvm::Value *>             args,
+    ArrayRef<llvm::Attribute::AttrKind> attribs,
+    IRBuilder<>&                        builder)
 {
-    Module* pModule = pInsertPos->getModule();
+    Module* pModule = builder.GetInsertBlock()->getParent()->getParent();
     Function* pFunc = dyn_cast_or_null<Function>(pModule->getFunction(funcName));
-    if (pFunc == nullptr)
+    if (!pFunc)
```
amdrexu: pFunc -> pFunc != nullptr

util/llpcInternal.cpp
```diff
@@ -77,20 +77,27 @@ Function* GetEntryPoint(
...
 // =====================================================================================================================
-// Emits a LLVM function call (inserted before the specified instruction), builds it automically based on return type
-// and its parameters.
-CallInst* EmitCall(
-    StringRef                     funcName,         // Name string of the function
-    Type*                         pRetTy,           // [in] Return type
-    ArrayRef<Value *>             args,             // [in] Parameters
-    ArrayRef<Attribute::AttrKind> attribs,          // Attributes
-    Instruction*                  pInsertPos)       // [in] Where to insert this call
+/// Emits a LLVM function call using the given builder. The callee is built automically based on return
```
amdrexu: Please use existing comment style.

util/llpcInternal.cpp
```diff
@@ -77,20 +77,27 @@ Function* GetEntryPoint(
...
+/// type and its parameters.
+///
+/// @oaram [in] funcName Name of the callee
+/// @param [in] pRetTy Return type of the callee
+/// @param [in] args Arguments to pass to the callee
+/// @param [in] attribs Function attributes
+/// @param [in] builder The IRBuilder to use for creating the call
+llvm::CallInst* EmitCall(
```
amdrexu: Please remove "llvm::" prefix. All source files in LLPC don't use commonly-used namespace unless there is ambiguity. Instead, place "using namespace XXX;" at the beginning of the function.

util/llpcInternal.cpp
```diff
@@ -77,20 +77,27 @@ Function* GetEntryPoint(
...
+    StringRef                           funcName,
```
amdrexu: Please copy comments to the function parameters like other EmitCall.

util/llpcInternal.cpp
nhaehnle: PAL uses this comment style in a few places, so I think it is okay. It is preferable because it avoids having to re-align comments on changes.

util/llpcInternal.cpp
nhaehnle: You're right, I'm going to fix this.

nhaehnle:
Addressed review comments and added missing lit test changes.

util/llpcInternal.cpp
amdrexu: PAL often uses this style for stable interface files and utility container files. So I think for component  internal files, we do not need doxygen style. For interface, such as llpc.h, llpcBuilder.h as the interface boundary, we can adopt this.

amdrexu: Also,  in PAL, you can find doxygen style comments often stay in interface header files to describe virtual base class methods. They are interfaces used by client drivers. So the documentation is preferable.

util/llpcInternal.cpp
amdrexu: I think `!boolean `is acceptable because the meaning is "not this". For a pointer `!ptr`, we have to understand it as "ptr not exist". Under such case, `ptr == nullptr` is very clear in human brain. This was widely discussed within 3D driver team many years ago. For pointer, always explicitly compare with `nullptr`.

nhaehnle: The 3D driver team's opinion is frankly not relevant when working on a compiler.

Using `!ptr` to check for pointer-nullness has been established canonical practice in C and C++ for **decades**. I'm not going to change this.

JaxLinAMD:
retest this please

PR #313, opened 2019-10-31
spirv: atomic loads need at least monotonic ordering

nhaehnle:
Unordered atomic loads are (incorrectly) hoisted out of loops. This is
a partial fix towards a test case under discussion for forward progress.


JaxLinAMD:
retest this please

PR #315, opened 2019-11-03
patch: Refactor the building of hardware_mapping PAL metadata

nhaehnle:
Use a common method for adding the mapping of a single API shader stage.
This helps us to avoid relying on the previously magic argument ordering
and can thereby streamline future changes.


JaxLinAMD:
retest this please

PR #316, opened 2019-11-03
PatchEntryPointMutate: simplify argument collection by removing redundant state

nhaehnle:
Remove the argIdx variable, since it is de facto a copy of argTys.size().
Not having to keep both in sync will simplify subsequent changes.


JaxLinAMD:
retest this please

PR #319, opened 2019-11-05
Add amdgpu-flat-work-group-size attribute

jayfoad:
For compute shaders, the frontend has to use this attribute to
communicate the workgroup size to the backend, because it may limit the
number of registers available to the backend code generator.

### DIFF
--- a/patch/llpcCodeGenManager.cpp
+++ b/patch/llpcCodeGenManager.cpp
@@ -206,6 +206,16 @@ void CodeGenManager::SetupTargetFeatures(
                 // Force s_barrier to be present (ignore optimization)
                 builder.addAttribute("amdgpu-flat-work-group-size", "128,128");
             }
+            if (pFunc->getCallingConv() == CallingConv::AMDGPU_CS)
+            {
+                // Set the work group size
+                const auto& csBuiltInUsage = pContext->GetShaderResourceUsage(ShaderStageCompute)->builtInUsage.cs;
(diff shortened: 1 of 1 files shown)


patch/llpcCodeGenManager.cpp
```diff
@@ -206,6 +206,14 @@ void CodeGenManager::SetupTargetFeatures(
                 // Force s_barrier to be present (ignore optimization)
                 builder.addAttribute("amdgpu-flat-work-group-size", "128,128");
             }
+            if (pFunc->getCallingConv() == CallingConv::AMDGPU_CS)
+            {
+                // Set the work group size
+                const auto& csBuiltInUsage = pContext->GetShaderResourceUsage(ShaderStageCompute)->builtInUsage.cs;
+                uint32_t flatWorkGroupSize = csBuiltInUsage.workgroupSizeX * csBuiltInUsage.workgroupSizeY * csBuiltInUsage.workgroupSizeZ;
```
amdrexu: This line might exceed 120 characters

patch/llpcCodeGenManager.cpp
```diff
@@ -206,6 +206,14 @@ void CodeGenManager::SetupTargetFeatures(
...
+                auto s = std::to_string(flatWorkGroupSize);
```
amdrexu: s -> flatWorkGroupSizeString

PR #320, opened 2019-11-06
Fix compiling and tests regression in some corner tests.

jiaolu:


util/llpcInternal.cpp
```diff
@@ -115,7 +115,11 @@ CallInst* EmitCall(
         }
     }
 
-    return builder.CreateCall(pFunc, args);
+    auto pCall = builder.CreateCall(pFunc, args);
+    pCall->setCallingConv(CallingConv::C);
+    pCall->setAttributes(pFunc->getAttributes());
```
nhaehnle: Maybe you can provide more detail internally, but I don't understand how this can be necessary. The effective attributes of a function call should always be the union of the attributes of the function and those of the call, so it's surprising that having them on the call itself presumably makes a difference.

translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5237,7 +5237,7 @@ template<> Value* SPIRVToLLVM::transValueWithOpcode<OpVariable>(
         }
     }
 
-    const uint32_t addrSpace = pPtrType->getPointerAddressSpace();
+    uint32_t addrSpace = pPtrType->getPointerAddressSpace();
```
nhaehnle: The pointer type returned by `transType` should already have the correct address space. Otherwise, I would worry that there could be inconsistencies in the applied LLVM types later on, e.g. when passing the OpVariable's value around. So if an address space fixup is required, it should better happen in `transType`.

PR #324, opened 2019-11-07
Add option -enable-timer-profile to skip time-pass itself overhear wh…

yukli-amd:


yukli-amd:
Add option -enable-timer-profile to skip time-pass itself overhear when calculate the compile time.
Test one CS,
                               shaderbench    time-passes   timer-profile
compile time(ms)       15.65                    17.3          14.6
It's reasonable that the result of shaderbench is 1ms more than timer-profile due to shaderbench will caculate create pal pipeline time extra.

PR #326, opened 2019-11-08
Add NGG support to GS

amdrexu:
Phase 3: Clear CTS issues for non culling mode.
- Remove the use of esArgIdx and gsArgIdx. The variables are redundant and
  are error-prone if we don't set them correctly.
- Fix an issue of RunEsOrEsVarient. When GS is present, ES should have a
  extra argument to store EsGsOffset for this primitive. It is missing.
- Also, correct another issue of RunEsOrEsVarient. If TES acts as ES, the
  argument orders of offChipLdsBase and isOffChip is different when GS is
  present or not.


amdrexu:
retest this please

amdrexu:
retest this please

JaxLinAMD:
retest this please

PR #327, opened 2019-11-11
Fix VFX parser issue (fails to read .pipe file).

amdrexu:
- The field option.allowVaryWaveSize is not dumped and hash update is
  missing as well.
- Simplify the calculation of each section member count. The nested
  macros will cause codes messy and are error-prone.

### DIFF
--- a/tool/vfx/vfxSection.h
+++ b/tool/vfx/vfxSection.h
@@ -1139,11 +1139,7 @@ public:
     void GetSubState(SubState& state) { state = m_state; };
 
 private:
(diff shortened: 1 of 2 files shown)


amdrexu:
retest this please

JaxLinAMD:
@amdrexu, dEQP-VK.api.object_management.multithreaded_shared_resources.graphics_pipeline failed, please have a look.

amdrexu:
Thanks. I will have a check.

amdrexu:
retest this please

amdrexu:
Thank you for this suggestion. Admittedly, I need two commits to resolve both issues.

PR #328, opened 2019-11-11
llpc: refactor ShaderModuleData structure

amingriyue:
No functionality changes.
This change is to easily export resource usage and fragnment shader output type.
The following patch will do these.

v2:
  1.New ShaderModuleEntry includes public and private parts.
  2.Rename struct OptShaderModuleData to ShaderModuleOptData.
  3.Rename struct ShaderModuleCaps to ShaderModuleAttribute.
v3:
  Fix Jenkins error.
v4:
  1.Improve coding style.
  2.Always evaluate entryCount for counting offset of data points,
     especially for data got from cache. Which fixes remaining Jenkins issue.
v5:
     make ShaderModuleExtraData inline in ShaderModuleDataEx.

### DIFF
--- a/include/llpc.h
+++ b/include/llpc.h
@@ -239,16 +239,76 @@ struct ShaderModuleBuildInfo
 #endif
 };
 
-/// Represents the header part of LLPC shader module data
-struct ShaderModuleDataHeader
+/// Represents the base data type
+enum class BasicType : uint32_t
 {
-    uint32_t hash[4];       // Shader hash code
+    Unknown = 0,          ///< Unknown
+    Float,                ///< Float
+    Double,               ///< Double
+    Int,                  ///< Signed integer
+    Uint,                 ///< Unsigned integer
+    Int64,                ///< 64-bit signed integer
+    Uint64,               ///< 64-bit unsigned integer
+    Float16,              ///< 16-bit floating-point
+    Int16,                ///< 16-bit signed integer
+    Uint16,               ///< 16-bit unsigned integer
+    Int8,                 ///< 8-bit signed integer
+    Uint8,                ///< 8-bit unsigned integer
+};
+
+/// Enumerates types of shader binary.
+enum class BinaryType : uint32_t
+{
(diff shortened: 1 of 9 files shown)


nhaehnle:
Thanks. Would it be possible to see the followup commit? It's not really possible to do a good review of this kind of change as a standalone change.

amingriyue:
Yes, I think so. That I purpose this change first is to get valuable suggestion for code reorganization. As you see, Qun and Rex already gave many valuable comments on that, especially we settlled down structure names, so that I can continue following work.
If without this and I directly send a large patch, any of renaming will take much effort on that.

include/llpc.h
```diff
@@ -239,16 +239,80 @@ struct ShaderModuleBuildInfo
...
+    uint32_t              entryCount;              ///< Shader entry count in the module
+    ShaderModuleEntryData entryDatas[1];           ///< Array of all shader entries in this module
+};
+
+
+/// Represents extended output of building a shader module (taking extra data info)
+struct ShaderModuleDataEx
+{
+    ShaderModuleData        common;        ///< Shader module common data
+    ShaderModuleExtraData   extra;         ///< Shader module extra data
```
nhaehnle: So if I'm understanding this right, the intention here is for `ShaderModuleExtraData::entryDatas` to be effectively a variable-sized array, which contains multiple array entries, one per entrypoint function in the module.

That seems fine, except by embedding `ShaderModuleExtraData` into `ShaderModuleDataEx`, we're now in the situation where `ShaderModuleDataEx` has a variable-length array embedded inside of it, but not at the top-level -- which is a risk to introducing bugs.

Since the `ShaderModuleExtraData` structure doesn't appear to be used standalone anywhere else, could you just inline its members into `ShaderModuleDataEx`? In that way, it is clear that there is a variable-length array at the end.

amingriyue: good idea. Will update it.

nhaehnle: Thank you, LGTM

PR #335, opened 2019-11-12
Fixed unused variable error missed in last change

trenouf:
Change-Id: I5a570d878a49c81c377b9a17a0f12d934d4a66a4

### DIFF
--- a/patch/gfx9/llpcNggPrimShader.cpp
+++ b/patch/gfx9/llpcNggPrimShader.cpp
@@ -3032,6 +3032,7 @@ void NggPrimShader::RunEsOrEsVariant(
 
     auto pEsArgBegin = pEsEntry->arg_begin();
     const uint32_t esArgCount = pEsEntry->arg_size();
+    LLPC_UNUSED(esArgCount);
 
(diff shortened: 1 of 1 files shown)


patch/gfx9/llpcNggPrimShader.cpp
```diff
@@ -3032,6 +3032,7 @@ void NggPrimShader::RunEsOrEsVariant(
 
     auto pEsArgBegin = pEsEntry->arg_begin();
     const uint32_t esArgCount = pEsEntry->arg_size();
+    LLPC_UNUSED(esArgCount);
```
amdrexu: I think we can replace esArgCount with pEsEntry->arg_size(). This variable is no longer actually used.
The same case is gsArgCount in NggPrimShader::RunGsVariant().
Sorry for that. I forget to clear them.

trenouf:
I'll just merge this though, because I think my last change probably causes a build breakage without it.

amdrexu:
It's OK. I think build-failure fix can be merged as soon as possible.

PR #339, opened 2019-11-12
Fix non gfx10 build

dnovillo:
This adds extra guards around code that is only valid when building with
GFX10 enabled.

Tested with GFX10 support disabled and GFX10 support enabled.

Note that this requires changes in the other drivers as well.

### DIFF
--- a/patch/gfx9/chip/llpcGfx9Chip.cpp
+++ b/patch/gfx9/chip/llpcGfx9Chip.cpp
@@ -30,6 +30,8 @@
  */
 #define DEBUG_TYPE "llpc-gfx9-chip"
 
+#include <stdio.h>
+
 #include "llpcGfx9Chip.h"
 
 namespace Llpc
--- a/patch/llpcPatchDescriptorLoad.cpp
+++ b/patch/llpcPatchDescriptorLoad.cpp
@@ -665,6 +665,7 @@ Value* PatchDescriptorLoad::BuildBufferCompactDesc(
                                             "",
                                             pInsertPoint);
     }
+#ifdef LLPC_BUILD_GFX10
     else if (gfxIp.major == 10)
     {
         SqBufRsrcWord3 sqBufRsrcWord3 = {};
@@ -683,6 +684,7 @@ Value* PatchDescriptorLoad::BuildBufferCompactDesc(
                                             "",
                                             pInsertPoint);
     }
+#endif
     else
     {
         LLPC_NOT_IMPLEMENTED();
--- a/patch/llpcPatchInOutImportExport.cpp
+++ b/patch/llpcPatchInOutImportExport.cpp
@@ -4235,8 +4235,8 @@ void PatchInOutImportExport::PatchGsBuiltInOutputExport(
         EmitCall(callName, m_pContext->VoidTy(), args, NoAttrib, pInsertPos);
     }
     else
-    {
 #endif
+    {
         switch (builtInId)
         {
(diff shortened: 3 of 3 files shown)


patch/gfx9/chip/llpcGfx9Chip.cpp
```diff
@@ -30,6 +30,8 @@
  */
 #define DEBUG_TYPE "llpc-gfx9-chip"
 
+#include <stdio.h>
+
```
nhaehnle: Why is that required?

dnovillo:
> Mostly LGTM, though we should probably just drop all those #ifdefs wholesale instead.

Yeah, I think most of the code is dynamically protected for GFX10, so it makes better sense to me.   I don't think I have a gfx9 part to test things on, however.

patch/gfx9/chip/llpcGfx9Chip.cpp
dnovillo: Oh, this one was for the calls to `snprintf`.  Interestingly, it only happens with our internal build system (clang-based).  I didn't really compare the pre-processed output to figure out why.

```
patch/gfx9/chip/llpcGfx9Chip.cpp:849:26: error: use of undeclared identifier 'snprintf'; did you mean 'vswprintf'?
        int32_t length = snprintf(unknownRegNameBuf, 256, "UNKNOWN(0x%08X)", regId);
                         ^~~~~~~~
                         vswprintf
/usr/include/wchar.h:628:12: note: 'vswprintf' declared here
extern int vswprintf (wchar_t *__restrict __s, size_t __n,
           ^
patch/gfx9/chip/llpcGfx9Chip.cpp:849:35: error: cannot initialize a parameter of type 'wchar_t *' with an lvalue of type 'char [256]'
        int32_t length = snprintf(unknownRegNameBuf, 256, "UNKNOWN(0x%08X)", regId);
                                  ^~~~~~~~~~~~~~~~~
/usr/include/wchar.h:628:43: note: passing argument to parameter '__s' here
extern int vswprintf (wchar_t *__restrict __s, size_t __n,
                                          ^
2 errors generated.
```

amdrexu:
We probably need to clear GFX10 macro in the near future. It is not required and recent code changes might break non-GFX10 build because developers tend to forget to add the macro.

patch/gfx9/chip/llpcGfx9Chip.cpp
nhaehnle: Huh, interesting. We do have one Clang-based builder internally that didn't show this. But yeah, it's obviously a reasonable change, I just got suspicious because it wasn't mentioned in the commit message.

dnovillo:
> We probably need to clear GFX10 macro in the near future. It is not required and recent code changes might break non-GFX10 build because developers tend to forget to add the macro.

It would be very useful to have a bot that built the common variants.  @kuhar was thinking of adding some bots for the configurations we are interested in.  Jakub, could you include a build that disables GFX10 (`-DXGL_BUILD_GFX10=OFF` in the `cmake` call).

patch/gfx9/chip/llpcGfx9Chip.cpp
dnovillo: Yeah, sorry about that.  I was in auto-pilot ;)

JaxLinAMD:
retest this please

dnovillo:
Thanks!

PR #342, opened 2019-11-13
Use array initializers instead of dynamic containers

jayfoad:
This usually makes the code simpler, and it removes dynamic memory
allocation which can be expensive.

### DIFF
--- a/builder/llpcBuilderImplInOut.cpp
+++ b/builder/llpcBuilderImplInOut.cpp
@@ -487,30 +487,31 @@ Value* BuilderImplInOut::ModifyAuxInterpValue(
     {
         // Add intrinsic to calculate I/J for interpolation function
         std::string evalInstName;
-        std::vector<Value*> evalArgs;
         auto pResUsage = getContext().GetShaderResourceUsage(ShaderStageFragment);
 
         if (inputInfo.GetInterpLoc() == InOutInfo::InterpLocCentroid)
         {
+            Value* evalArg = nullptr;
+
             evalInstName = LlpcName::InputImportBuiltIn;
             if (inputInfo.GetInterpMode() == InOutInfo::InterpModeNoPersp)
             {
                 evalInstName += "InterpLinearCentroid";
-                evalArgs.push_back(getInt32(spv::BuiltInInterpLinearCentroid));
+                evalArg = getInt32(spv::BuiltInInterpLinearCentroid);
                 pResUsage->builtInUsage.fs.noperspective = true;
                 pResUsage->builtInUsage.fs.centroid = true;
             }
             else
             {
                 evalInstName += "InterpPerspCentroid";
-                evalArgs.push_back(getInt32(spv::BuiltInInterpPerspCentroid));
+                evalArg = getInt32(spv::BuiltInInterpPerspCentroid);
                 pResUsage->builtInUsage.fs.smooth = true;
                 pResUsage->builtInUsage.fs.centroid = true;
             }
 
             pAuxInterpValue = EmitCall(evalInstName,
                                        VectorType::get(getFloatTy(), 2),
-                                       evalArgs,
+                                       { evalArg },
                                        Attribute::ReadOnly,
                                        &*GetInsertPoint());
         }
--- a/builder/llpcBuilderImplSubgroup.cpp
+++ b/builder/llpcBuilderImplSubgroup.cpp
@@ -1476,14 +1476,12 @@ Value* BuilderImplSubgroup::CreateGroupBallot(
     // The not equal predicate for the icmp intrinsic is 33.
     Constant* const pPredicateNE = getInt32(33);
 
-    SmallVector<Type*, 2> types;
-
     // icmp has a new signature (requiring the return type as the first type).
-    types.push_back(getIntNTy(GetShaderSubgroupSize()));
-    types.push_back(getInt32Ty());
-
     Value* pResult = CreateIntrinsic(Intrinsic::amdgcn_icmp,
-                                     types,
+                                     {
+                                          getIntNTy(GetShaderSubgroupSize()),
+                                          getInt32Ty()
+                                     },
                                      {
                                           pValueAsInt32,
                                           getInt32(0),
--- a/lower/llpcSpirvLowerLoopUnrollControl.cpp
+++ b/lower/llpcSpirvLowerLoopUnrollControl.cpp
@@ -147,11 +147,12 @@ bool SpirvLowerLoopUnrollControl::runOnModule(
                 // one operand pointing to itself, meaning that the SPIR-V did not
                 // have an unroll or don't-unroll directive, so we can add the force
                 // unroll count metadata.
-                SmallVector<llvm::Metadata*, 2> opUnrollCountMeta;
-                opUnrollCountMeta.push_back(MDString::get(*m_pContext, "llvm.loop.unroll.count"));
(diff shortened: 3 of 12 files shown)


patch/llpcFragColorExport.cpp
```diff
@@ -1376,13 +1374,15 @@ Value* FragColorExport::Run(
                     comps[i] = pUndefFloat;
                 }
 
-                std::vector<Attribute::AttrKind> attribs;
-                attribs.push_back(Attribute::ReadNone);
+                Attribute::AttrKind attribs[] = {
+                    Attribute::ReadNone
+                };
```
nhaehnle: Since the array is a singleton, this can presumably be simplified further?

patch/llpcPatchInOutImportExport.cpp
```diff
@@ -1904,10 +1896,9 @@ Value* PatchInOutImportExport::PatchFsGenericInputImport(
         pJ = ExtractElementInst::Create(pIJ, ConstantInt::get(m_pContext->Int32Ty(), 1), "", pInsertPos);
     }
 
-    std::vector<Value*> args;
-
-    std::vector<Attribute::AttrKind> attribs;
-    attribs.push_back(Attribute::ReadNone);
+    Attribute::AttrKind attribs[] = {
+        Attribute::ReadNone
+    };
```
nhaehnle: This can presumably be simplified further?

patch/llpcFragColorExport.cpp
jayfoad: You're right, I could just write `Attribute::AttrKind attribs = Attribute::ReadNone;`. I am slightly prejudiced against this simply because I didn't know that you can implicitly construct an ArrayRef from a single element.

patch/llpcPatchInOutImportExport.cpp
jayfoad: Likewise.

builder/llpcBuilderImplInOut.cpp
```diff
@@ -487,30 +487,31 @@ Value* BuilderImplInOut::ModifyAuxInterpValue(
     {
         // Add intrinsic to calculate I/J for interpolation function
         std::string evalInstName;
-        std::vector<Value*> evalArgs;
         auto pResUsage = getContext().GetShaderResourceUsage(ShaderStageFragment);
 
         if (inputInfo.GetInterpLoc() == InOutInfo::InterpLocCentroid)
         {
+            Value* evalArg;
```
amdrexu: evalArg -> evalArg = nullptr

builder/llpcBuilderImplSubgroup.cpp
```diff
@@ -1476,11 +1476,11 @@ Value* BuilderImplSubgroup::CreateGroupBallot(
...
     // icmp has a new signature (requiring the return type as the first type).
-    types.push_back(getIntNTy(GetShaderSubgroupSize()));
-    types.push_back(getInt32Ty());
+    Type* types[] = {
+        getIntNTy(GetShaderSubgroupSize()),
+        getInt32Ty()
+    };
 
     Value* pResult = CreateIntrinsic(Intrinsic::amdgcn_icmp,
                                      types,
```
amdrexu: Would it be possible to insert the elements to this function call direct without making a temp variable

```
CreateIntrinsic(Intrinsic::amdgcn_icmp,
                        {
                            getIntNTy(GetShaderSubgroupSize()),
                            getInt32Ty()
                        }
```

lower/llpcSpirvLowerLoopUnrollControl.cpp
```diff
@@ -147,10 +147,11 @@ bool SpirvLowerLoopUnrollControl::runOnModule(
                 // one operand pointing to itself, meaning that the SPIR-V did not
                 // have an unroll or don't-unroll directive, so we can add the force
                 // unroll count metadata.
-                SmallVector<llvm::Metadata*, 2> opUnrollCountMeta;
-                opUnrollCountMeta.push_back(MDString::get(*m_pContext, "llvm.loop.unroll.count"));
-                opUnrollCountMeta.push_back(ConstantAsMetadata::get(
-                    ConstantInt::get(Type::getInt32Ty(*m_pContext), m_forceLoopUnrollCount)));
+                llvm::Metadata* opUnrollCountMeta[] = {
```
amdrexu: opUnrollCountMeta -> unrollCountMeta

patch/llpcFragColorExport.cpp
amdrexu: I think you can place this in the function directly { Attribute::ReadNone } without defining this.

patch/llpcFragColorExport.cpp
```diff
@@ -1376,13 +1374,15 @@ Value* FragColorExport::Run(
...
-                args.clear();
-                args.push_back(comps[0]);
-                args.push_back(comps[1]);
+                Value* args[] = {
+                    comps[0],
+                    comps[1]
+                };
                 comps[0] = EmitCall("llvm.amdgcn.cvt.pkrtz",
                                     m_pContext->Float16x2Ty(),
                                     args,
```
amdrexu: The same as above. Place { comps[0], comps[1] } to EmitCall without defining args. This is more clear since the arg list is very simple.

patch/llpcPatchInOutImportExport.cpp
```diff
@@ -2795,7 +2772,9 @@ Value* PatchInOutImportExport::PatchFsBuiltInInputImport(
     auto& builtInUsage = m_pContext->GetShaderResourceUsage(ShaderStageFragment)->builtInUsage.fs;
     auto& inOutUsage = m_pContext->GetShaderResourceUsage(ShaderStageFragment)->inOutUsage;
 
-    std::vector<Value*> args;
+    Attribute::AttrKind attribs[] = {
+        Attribute::ReadNone
```
amdrexu: The same. Use { Attribute::ReadNone } in EmitCall

patch/llpcPatchInOutImportExport.cpp
```diff
@@ -2844,12 +2822,9 @@ Value* PatchInOutImportExport::PatchFsBuiltInInputImport(
                 GetFunctionArgument(m_pEntryPoint, entryArgIdxs.fragCoord.w),
             };
 
-            std::vector<Attribute::AttrKind> attribs;
-            attribs.push_back(Attribute::ReadNone);
-
-            args.clear();
-            args.push_back(fragCoord[3]);
-
+            Value* args[] = {
```
amdrexu: Use { fragCoord[3] } in EmitCall

patch/llpcPatchInOutImportExport.cpp
jayfoad: Are you sure? `attribs` is used more than once, so if you ever need to change the list of attributes, you will have to make the same change in more than one place.

patch/llpcFragColorExport.cpp
amdrexu: I suggest this because placing the attribs in EmitCall is more clear. We do not have to check what the value of attribs is. And in most cases, we only have one attributes in the list. It is up to your decision. I also consider we can use "NoAttrib" directly for those simple ALU functions.

jayfoad: I will leave it like it is, but I don't mind if someone else wants to change it.

PR #343, opened 2019-11-13
patch/llpcSystemValues: Refactor internal functions in terms of IRBuilder

nhaehnle:
The main motivation behind this is to avoid awkward code that passes
Instructions around instead of Values and then attempts to insert
subsequent code just after those instructions.

In part the original code is inflexible in that it will fail if used on
instructions at the end of an incomplete basic block. More importantly,
using the IRBuilder is a more idiomatic use of LLVM and happens to make
the code much more readable by virtue of making it more concise.


JaxLinAMD:
retest this please

PR #344, opened 2019-11-13
patch/configBuilder: Refactor the PAL ABI register metadata collection

nhaehnle:
Allow register metadata to be built up successively via
ConfigBuilderBase::AppendConfig by using a proper container in the base
class.

This has various benefits:

1. The code becomes more idiomatic C++: no naked operator new/delete,
   use of containers instead of manually maintained pointer/count pairs,
   use of constructors instead of Init methods.

2. Code duplication is reduced.

3. The code becomes more (de)composable which allows further code
   improvements in the future (not part of this series of changes).
   For example, a future change could break up the various
   BuildPipelineXyz methods, which further reduces code duplication and
   overall code volume.

4. We avoid a heap allocation in the common case thanks to the use of
   llvm::SmallVector.

### DIFF
--- a/patch/llpcConfigBuilderBase.cpp
+++ b/patch/llpcConfigBuilderBase.cpp
@@ -75,7 +75,6 @@ ConfigBuilderBase::ConfigBuilderBase(
 // =====================================================================================================================
 ConfigBuilderBase::~ConfigBuilderBase()
 {
-    delete[] m_pConfig;
 }
 
 // =====================================================================================================================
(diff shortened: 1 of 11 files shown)


nhaehnle:
That's an interesting idea, but like you say it makes the code tricky and a bit dangerous (what if we do the allocation this way, but then later a dynamic insertion of registers causes the vector to re-allocate). So I prefer to keep it as-is.

patch/llpcConfigBuilderBase.h
```diff
@@ -68,12 +73,25 @@ class ConfigBuilderBase
...
     void SetEsGsLdsSize(uint32_t value);
 
+    void AppendConfig(llvm::ArrayRef<Util::Abi::PalMetadataNoteEntry> config);
+    void AppendConfig(uint32_t key, uint32_t value);
+
+    template<typename T>
+    void AppendConfig(const T& config)
+    {
+        static_assert(T::containsPalAbiMetadataOnly,
+                      "may only be used with structs that are fully metadata notes");
```
amdrexu: may -> May
I think we use Uppercase-initial for assert messages like what we have done for normal comments.

patch/llpcConfigBuilderBase.cpp
```diff
@@ -346,6 +345,47 @@ void ConfigBuilderBase::SetPipelineHash()
...
+    entry.key = key;
+    entry.value = value;
+    m_config.push_back(entry);
+}
+
+// =====================================================================================================================
+/// Append an array of entries to the PAL register metadata. Invalid keys are filtered out.
+///
+/// @param [in] config The array of register metadata entries.
+void ConfigBuilderBase::AppendConfig(llvm::ArrayRef<Util::Abi::PalMetadataNoteEntry> config)
```
amdrexu: Could remove "llvm::" prefix

patch/llpcConfigBuilderBase.cpp
```diff
@@ -346,6 +345,47 @@ void ConfigBuilderBase::SetPipelineHash()
...
+{
+    uint32_t count = 0;
+
+    for (const auto &entry : config)
```
amdrexu: It's OK to place "&" and "*" just before the identifiers. Currently, I hope we can still follow old style "const auto&" until we find a timeslot to re-format existing codes.

linqun:
> That's an interesting idea, but like you say it makes the code tricky and a bit dangerous (what if we do the allocation this way, but then later a dynamic insertion of registers causes the vector to re-allocate). So I prefer to keep it as-is.

I am OK to keep current version, but I don't think it will causes reallocate, because the size of m_config is 128, it is enough to hold everything.

JaxLinAMD:
retest this please

nhaehnle:
Okay, thank you.

JaxLinAMD:
@nhaehnle, this pull requst causes "non-llpc" build failed, could you take a look?

nhaehnle:
I think it was the added dependency on llpcConfigBuilderBase.h. I've removed that now.

JaxLinAMD:
retest this please

amdrexu:
Please rebase your commit.

PR #346, opened 2019-11-14
Enable gcc error on unused function

trenouf:
This will help to stop errors (like that fixed in PR345) creeping in
that break clang builds.

Change-Id: If218f365b029021ba8f179c50014084d9110554e


trenouf:
Hopefully this will fail to show that the error check works...

trenouf:
This should now pass, as I have merged Xue's fix.

PR #351, opened 2019-11-16
llpc: export resource node and fragment shader output during buildShaderModule

amingriyue:
Parse resource node usage and fragment shader output from llvm pass. and export it to next phase.
v2:
  1.Collect resource node in SPIRAS_Constant
  2.Add new parameter to control resource node collection behavior
  3.Get arraySize from LLVM type
  4.Using simple struct std::vector to return resource node data

v3:
  1.fix Jenkins test failures.
  2.switch to use std::set for resource node instead of std::map
  2.add new index member in FsOutVariable

v4: Some nitpick fix
v5: Coding style improvement
v6,v7: Further improve coding style

### DIFF
--- a/context/llpcCompiler.cpp
+++ b/context/llpcCompiler.cpp
@@ -68,6 +68,7 @@
 #include "llpcPatch.h"
 #include "llpcPipelineDumper.h"
 #include "llpcSpirvLower.h"
+#include "llpcSpirvLowerResourceCollect.h"
 #include "llpcTimerProfiler.h"
 #include "llpcVertexFetch.h"
 #include <mutex>
@@ -495,6 +496,8 @@ Result Compiler::BuildShaderModule(
     std::vector<ShaderEntryName> entryNames;
     SmallVector<ShaderModuleEntryData, 4> moduleEntryDatas;
     SmallVector<ShaderModuleEntry, 4> moduleEntries;
+    SmallVector<FsOutInfo, 4> fsOutInfos;
+    std::map<uint32_t, std::vector<ResourceNodeData>> entryResourceNodeDatas; // Map entry ID and resourceNodeData
 
     ShaderEntryState cacheEntryState = ShaderEntryState::New;
     CacheEntryHandle hEntry = nullptr;
@@ -646,7 +649,10 @@ Result Compiler::BuildShaderModule(
                     shaderInfo.pEntryTarget = entryNames[i].pName;
                     lowerPassMgr.add(CreateSpirvLowerTranslator(static_cast<ShaderStage>(entryNames[i].stage),
                                                                 &shaderInfo));
-                    lowerPassMgr.add(CreateSpirvLowerResourceCollect());
+                    bool collectDetailUsage = (entryNames[i].stage == ShaderStageFragment) ? true : false;
+                    auto pResCollectPass = static_cast<SpirvLowerResourceCollect*>(
+                                           CreateSpirvLowerResourceCollect(collectDetailUsage));
+                    lowerPassMgr.add(pResCollectPass);
                     if (EnableOuts())
                     {
                         lowerPassMgr.add(createPrintModulePass(outs(), "\n"
@@ -683,6 +689,26 @@ Result Compiler::BuildShaderModule(
 
                     moduleEntry.resUsageSize = moduleBinary.size() - moduleEntry.entryOffset - moduleEntry.entrySize;
                     moduleEntry.passIndex = passIndex;
+                    if (pResCollectPass->DetailUsageValid())
+                    {
+                        auto& resNodeDatas = pResCollectPass->GetResourceNodeDatas();
+                        moduleEntryData.resNodeDataCount = resNodeDatas.size();
+                        for (auto resNodeData : resNodeDatas)
+                        {
+                            ResourceNodeData data = {};
+                            data.type = resNodeData.second;
+                            data.set = resNodeData.first.value.set;
+                            data.binding = resNodeData.first.value.binding;
+                            data.arraySize = resNodeData.first.value.arraySize;
+                            entryResourceNodeDatas[i].push_back(data);
+                        }
+
+                        auto& fsOutInfos = pResCollectPass->GetFsOutInfos();
+                        for (auto& fsOutInfo : fsOutInfos)
+                        {
+                            fsOutInfos.push_back(fsOutInfo);
+                        }
+                    }
                     moduleEntries.push_back(moduleEntry);
                     moduleEntryDatas.push_back(moduleEntryData);
                     delete pModule;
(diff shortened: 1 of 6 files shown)


util/llpcShaderModuleHelper.cpp
```diff
@@ -28,20 +28,327 @@
...
 {
+struct ResourceNodeDataSet
+{
+    std::vector<ResourceNodeData> nodes;   // Vector of resource mapping nodes
+    std::map<uint32_t, uint32_t> bindingMap;  // Map from binding to index in nodes vector
+};
+
+// =====================================================================================================================
+// Collect resource nodes from sprirv binary
+void ShaderModuleHelper::CollectResourceNodes(
```
linqun: please don't collect resource in this way.
We only need fragment shader output info when shader module opt is enabled. we can collect them in some LLVM pass, instead of re-parse the whole SPIRV binary.

amingriyue: OK, I will move them into SpirvLowerResourceCollect pass.

lower/llpcSpirvLowerResourceCollect.h
```diff
@@ -68,6 +75,8 @@ class SpirvLowerResourceCollect:
     // -----------------------------------------------------------------------------------------------------------------
 
     ResourceUsage*  m_pResUsage;    // Resource usage of the shader stage
+
+    std::map<uint32_t, ResourceNodeDataSet> m_ResNodeSets; // Resource node data
```
linqun: I prefer using simple struct like std::set<ResourceNodeData>

JaxLinAMD:
retest this please

amdrexu:
Once everyone agrees with this change, please ping me. I will have a check. Sorry for the delay.

JaxLinAMD:
retest this please

amingriyue:
Hi  Nicolai,
The exported info will be used in build pipeline. Although the information is already in memtadata in IR, as you see, we still need to do some calculation for them, and then we get the exact info. I think we cannot avoid the calculation. When we put the calculation in BuildShaderModule and export, we will be able to use existing data struct/interface in buildPipeline.
Above is my understanding, maybe Qun has more explicit explain for you.

trenouf:
I too do not understand the point of this. Information on FS outputs is already collected in BuilderImplInOut::MarkFsOutputType, when recorded builder calls are replayed in BuilderReplayer. Why do we need the extra step? Whatever calculations are needed on it should be done in the middle-end (BuilderImpl* and patch), not the front-end.

Also, I am in the middle of changes to remove the block of outside-IR data for ResourceUsage, by separating out the shader modes and recording them in IR. I don't think we should be adding another block of outside-IR data there.

JaxLinAMD:
retest this please

linqun:
This information will be used in partial pipeline build (auto-layout pipeline in previous discussion), fs output type will be used as a hit to generate fake color buffer format.

- because this information will be used in client driver, instead of within LLPC, so we need return it in ShaderModuleDataEx.
- to avoid breaking the LLPC refactory, we add nothing in LLPC context or resource usage. instead, we collect related information from resource collection pass directly.
- please let me know if it is still conflict with you. thanks!

JacobHeAMD:
retest this please

context/llpcCompiler.cpp
```diff
@@ -646,7 +649,10 @@ Result Compiler::BuildShaderModule(
                     shaderInfo.pEntryTarget = entryNames[i].pName;
                     lowerPassMgr.add(CreateSpirvLowerTranslator(static_cast<ShaderStage>(entryNames[i].stage),
                                                                 &shaderInfo));
-                    lowerPassMgr.add(CreateSpirvLowerResourceCollect());
+                    bool collectDetailUsage = (entryNames[i].stage == ShaderStageFragment) ? true : false;
+                    auto *pResCollectPass = static_cast<SpirvLowerResourceCollect*>(
```
amdrexu: auto * -> auto

lower/llpcSpirvLowerResourceCollect.h
```diff
@@ -42,14 +42,41 @@ namespace Llpc
...
+// Compact ResourceNodeData into an uint64 key
+union ResourceNodeDataKey
+{
+    struct
+    {
+        uint64_t set        :   16;  // Resource set
+        uint64_t binding    :   16;  // Resource binding
+        uint64_t arraySize  :   16;  // Resource array size
+        uint64_t reserved   :   16;
+    } nodeId;
```
amdrexu: nodeId -> value
I think it is better to call the structure "value" for this "key".

lower/llpcSpirvLowerResourceCollect.cpp
```diff
@@ -52,19 +52,102 @@ char SpirvLowerResourceCollect::ID = 0;
...
+    m_detailUsageValid(false)
 {
     initializeSpirvLowerResourceCollectPass(*PassRegistry::getPassRegistry());
 }
 
+// =====================================================================================================================
+// Collect resource node data
+void SpirvLowerResourceCollect::CollectResourceNodeData(
+    const GlobalVariable* pGlobal,       // Llvm global variable
+    const Type* pGlobalTy)               // Llvm global variable type
```
amdrexu: Remove pGlobalTy. We can add codes at the beginning of this function

```
{
    auto pGlobalTy = pGlobal->getType()->getContainedType(0);

    ....
}
```

lower/llpcSpirvLowerResourceCollect.cpp
```diff
@@ -52,19 +52,102 @@ char SpirvLowerResourceCollect::ID = 0;
...
+    m_collectDetailUsage(collectDetailUsage),
+    m_detailUsageValid(false)
 {
     initializeSpirvLowerResourceCollectPass(*PassRegistry::getPassRegistry());
 }
 
+// =====================================================================================================================
+// Collect resource node data
+void SpirvLowerResourceCollect::CollectResourceNodeData(
+    const GlobalVariable* pGlobal,       // Llvm global variable
```
amdrexu: // Llvm global variable
->
// [in] Global variable to collect resource node data

lower/llpcSpirvLowerResourceCollect.h
```diff
@@ -64,10 +91,18 @@ class SpirvLowerResourceCollect:
     const llvm::Type* GetFlattenArrayElementType(const llvm::Type* pTy) const;
 
     void CollectExecutionModeUsage();
+    void CollectResourceNodeData(const GlobalVariable* pGlobal, const Type* pGlobalTy);
```
amdrexu: Remove `const Type* pGlobalTy`. We can deduce it from `pGlobal`

amingriyue:
retest this please

amingriyue:
retest this please

amingriyue:
retest this please

amingriyue:
retest this please

amingriyue:
Ping to merge, thanks.

PR #352, opened 2019-11-16
Add issues found by GrpahicsFuzz to shaderdb

kuhar:
This PR adds all past issues reported by @paulthomson as shaderdb lit tests.

The motivation is to track which issues are already fixed and ensure LLPC does not regress in the future. The unfixed bugs are also tracked -- failing tests are marked with XFAIL such that we know when they become fixed.

### DIFF
--- a/test/shaderdb/fuzzer/GraphicsFuzz_TestCallIfWhileSwitch.spvas
+++ b/test/shaderdb/fuzzer/GraphicsFuzz_TestCallIfWhileSwitch.spvas
@@ -1,0 +1,104 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -verify-ir -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC.*}} SPIRV-to-LLVM translation results
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
+;
+; Based on https://github.com/GPUOpen-Drivers/llpc/issues/203.
+
+; SPIR-V
+; Version: 1.0
+; Generator: Khronos Glslang Reference Front End; 7
+; Bound: 111
+; Schema: 0
+               OpCapability Shader
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Fragment %4 "main" %59
+               OpExecutionMode %4 OriginUpperLeft
+               OpSource ESSL 310
+               OpName %4 "main"
+               OpName %18 "data"
+               OpName %40 "buf0"
+               OpMemberName %40 0 "injectionSwitch"
+               OpName %42 ""
+               OpName %59 "_GLF_color"
+               OpDecorate %18 RelaxedPrecision
+               OpMemberDecorate %40 0 Offset 0
+               OpDecorate %40 Block
+               OpDecorate %42 DescriptorSet 0
+               OpDecorate %42 Binding 0
+               OpDecorate %59 Location 0
+               OpDecorate %71 RelaxedPrecision
+               OpDecorate %85 RelaxedPrecision
+               OpDecorate %95 RelaxedPrecision
+               OpDecorate %110 RelaxedPrecision
+               OpDecorate %99 RelaxedPrecision
+          %2 = OpTypeVoid
+          %3 = OpTypeFunction %2
+          %6 = OpTypeInt 32 1
+          %7 = OpTypePointer Function %6
+         %13 = OpConstant %6 1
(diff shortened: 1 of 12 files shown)


kuhar:
Current execution log:
```
$ ninja check-amdllpc
[0/1] Running the AMDLLPC regression tests
Testing Time: 11.37s
  Expected Passes    : 939
  Expected Failures  : 3
```

paulthomson:
Nice! I have a script that downloads the latest version of dEQP (and pending GraphicsFuzz CLs), extracts just the GraphicsFuzz AmberScript tests and then extracts the shaders in assembly (.asm) and binary (.spv) form. It now also outputs .spvas files like in this PR (but without any XFAIL commands). I attach the output. This includes all GraphicsFuzz tests (not just the ones that failed on LLPC), but they are all fairly small and each one exposed *some* issue in *some* shader compiler at some point, and we often find they cause failures across different drivers. Let me know if it needs some tweaking to be useful.

[graphicsfuzz.zip](https://github.com/GPUOpen-Drivers/llpc/files/3855625/graphicsfuzz.zip)

Of course, you could go further and extract all shaders from a dEQP run. Or you could just run dEQP itself on the driver on a regular basis (which is no doubt being done to some extent) or as part of PR checks, but this typically requires a lot more effort/power.

test/shaderdb/fuzzer_found/GraphicsFuzz_FragmentShader_CallIfWhileSwitch.spvas
```diff
@@ -0,0 +1,104 @@
+; BEGIN_SHADERTEST
```
amdrexu: Thanks for adding those assembly tests. Please change the directory and test name
```
shaderdb/fuzzer_found/GraphicsFuzz_FragmentShader_CallIfWhileSwitch.spvas
->
shaderdb/fuzzer/GraphicsFuzz_TestCallIfWhileSwitch.spvas
```
(1) Create a "fuzzer" fold is OK. Although the issues are found, we actually don't need to mention this as folder name ;)
(2) The case name style is often this: OpXXX_TestYYY, ExtXXX_TestYYY. So for all cases, don't mention the shader stages unless we do need this info because of stag-dependent implementation. Please add "Test" as well to following existing case name style.

JaxLinAMD:
retest this please

JaxLinAMD:
@kuhar shaderdb test failed:
unexpected passing tests: shaderdb/fuzzer_found/GraphicsFuzz_FragmentShader_MixFloorAdd.spvas
                                            shaderdb/fuzzer_found/GraphicsFuzz_FragmentShader_SimilarNestedIfs.spvas

please take a look.

test/shaderdb/fuzzer_found/GraphicsFuzz_FragmentShader_CallIfWhileSwitch.spvas
kuhar: Thanks for the suggestion. I renamed all the files.

JaxLinAMD:
retest this please

JaxLinAMD:
Please see precheck in result.
http://ocltc.amd.com:8111/viewLog.html?buildId=25839264&tab=buildResultsDiv&buildTypeId=Vulkan_XglLlpc_Tests_Amdllpc_VulkanTestsAmdllpcLinux

Thanks,
Jax

kuhar:
@JaxLinAMD

> This site can’t be reached
> ocltc.amd.com’s server IP address could not be found.
> `ERR_NAME_NOT_RESOLVED`
Anyhow, I'll just remove these two test for now if they require assertions enabled.

JaxLinAMD:
@kuhar besides these two tests, also shaderdb/fuzzer/GraphicsFuzz_TestDiscardContinueReturn.spvas
is "unexpected pass". thanks.

kuhar:
@JaxLinAMD 
I removed all previously xfailed tests. Should work on all build configurations now.

kuhar:
I think that the remaining XFAIL'd tests can be in the future added as an optional CMAKE target and only run when assertions are enabled. This could be useful for running debug build in CI. But let's keep it separate from this PR.

JaxLinAMD:
retest this please

nhaehnle:
> One potential issue that I can think of is that XFAILED tests can still pass with assertions disabled. In that case, we would have to come up with a better way of making sure that we handle these failures.

For future reference, llvm-lit allows test to be run conditionally depending on some check. In this particular case, the canonical way would be to add a line:
```
# REQUIRES: asserts
```
... to indicate that the test should only be run for builds with assertions enabled. There are some gnarly questions about how this integrates into our build system where we might have LLPC and LLVM assertions toggled independently, but at least that's the general direction for this kind of thing.

PR #354, opened 2019-11-18
Fix an issue of VK_EXT_vertex_attribute_disivor (divisor == 0)

amdrexu:
In Vulkan spec, this case is valid. A value of 0 can be used for the
divisor if the feature flag vertexAttributeInstanceRateZeroDivisor is
enabled. In this case, the same vertex attribute will be applied to all
instances.


nhaehnle:
We should use builder-style here eventually, but this change seems fine to me as-is.

amdrexu:
> We should use builder-style here eventually, but this change seems fine to me as-is.

Yes, using builder across LLPC will be a TODO task. Although the changes will be large, I think it is worth making such modification.

PR #360, opened 2019-11-19
Set default value 0xf to channelWriteMask

yukli-amd:
This change is to fix compatibility with the old dumped pipelines,
the channelWriteMask may not set in those pipelines, then the offline
compilation tools will not work properly.

### DIFF
--- a/tool/vfx/vfxSection.h
+++ b/tool/vfx/vfxSection.h
@@ -1081,6 +1081,7 @@ public:
         Section(m_addrTable, MemberCount, SectionTypeUnset, "colorBuffer")
     {
         memset(&m_state, 0, sizeof(m_state));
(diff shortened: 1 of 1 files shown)


tool/vfx/vfxSection.h
```diff
@@ -1081,6 +1081,7 @@ class SectionColorBuffer : public Section
         Section(m_addrTable, MemberCount, SectionTypeUnset, "colorBuffer")
     {
         memset(&m_state, 0, sizeof(m_state));
+        m_state.channelWriteMask = 0xf;
```
amdrexu: 0xf -> 0xF, we use uppercase for hex numbers.

yukli-amd: Sure, will update this when submit this change.

PR #361, opened 2019-11-20
Use IRBuilder to refactor llpcPatchCopyShader

amdrexu:
- Use IRBuilder to rewrite codes constructing IR. This will simplify the
  statements and make codes more readable.
- Inline args to EmitCall if necessary. This will make args close to
  EmitCall and is easy to check what args are.
- Remove "llvm" prefix since this is a source file and "using namespace
  llvm" is specified.

### DIFF
--- a/patch/llpcPatchCopyShader.cpp
+++ b/patch/llpcPatchCopyShader.cpp
@@ -30,6 +30,7 @@
  */
 #define DEBUG_TYPE "llpc-patch-copy-shader"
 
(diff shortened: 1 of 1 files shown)


amdrexu:
I decide to simplify copy shader codes to make it clear, especially for ExportOutput(). This is just a clean-up step.

amdrexu:
Yes, I plan to do this. This change just starts to use IRBuilder.

PR #365, opened 2019-11-22
Refactor llpcPatchCopyShader by using IRBuilder as insert point and rework initialization of LDS pointer and GS-VS ring buffer descriptor

amdrexu:
- Always use IRBuilder as insert point rathert than the single
  instruction.
- The initialization of LDS pointer and GS-VS ring buffer descriptor are
  mutually exclusive for on-chip GS and off-chip GS. No need of doing this
  for both.


amdrexu:
retest this please

amdrexu:
I will land this to resume my rework of copy shader.

PR #367, opened 2019-11-25
Add NGG support to GS

amdrexu:
Phase 3: Clear CTS issues for non culling mode.

- Misuse emitCounter to calculate primitive data, causing wrong vertex
  indices. EmitCounter will be adjusted dynamically when we lower GS_EMIT
  message. OutVertCounter is always incremented and represents the
  counting of GS output vertices.

- Return a value of WWM result of subgroupInclusiveAdd(). The highest
  thread has the value of GS output vertex count.

- Clarify the use of argument waveId in NGG. Originally, this value
  represents GS wave ID used in GS_EMIT/GS_CUT messages. Now, those GS
  messages are lowered in NGG. The value is actually  the ID of a wave
  in one NGG subgroup.

- Promote outVertCountInWave to SGPR. The value should be uniform in
  a wave,  but phi resolving leads it to be a VGPR. When GS threads
  are active, those threads take an uniform value, representing
  total number of GS output vertices in this wave. Otherwise, non GS
  threads take a value of zero. After we handle GS threads, we need
  a SGPR value from the first thread.

- The field OUTPRIM_TYPE of VGT_GS_OUT_PRIM_TYPE is incorrectly
  set. When GS is present, NGG should use GS output primitive type
  specified in shader.

Clear CTS test group: dEQP-VK.geometry.input.*

### DIFF
--- a/patch/gfx9/chip/llpcGfx9ConfigBuilder.cpp
+++ b/patch/gfx9/chip/llpcGfx9ConfigBuilder.cpp
@@ -2285,7 +2285,31 @@ Result ConfigBuilder::BuildPrimShaderRegConfig(
     SET_REG_FIELD(&pConfig->m_primShaderRegs, VGT_GS_PER_VS, GS_PER_VS, GsThreadsPerVsThread);
 
     VGT_GS_OUTPRIM_TYPE gsOutputPrimitiveType = POINTLIST;
-    if (hasTs)
+    if (hasGs)
+    {
+        // GS present
+        if (gsInOutUsage.outputMapLocCount == 0)
+        {
+            gsOutputPrimitiveType = POINTLIST;
+        }
+        else if (gsBuiltInUsage.outputPrimitive == OutputPoints)
+        {
+            gsOutputPrimitiveType = POINTLIST;
+        }
+        else if (gsBuiltInUsage.outputPrimitive == OutputLineStrip)
+        {
+            gsOutputPrimitiveType = LINESTRIP;
+        }
+        else if (gsBuiltInUsage.outputPrimitive == OutputTriangleStrip)
+        {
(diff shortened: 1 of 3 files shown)


amdrexu:
retest this please

patch/gfx9/chip/llpcGfx9ConfigBuilder.cpp
```diff
@@ -2308,31 +2311,56 @@ Result ConfigBuilder::BuildPrimShaderRegConfig(
...
-        {
-            gsOutputPrimitiveType = LINESTRIP;
-        }
-        else if ((topology == VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST) ||
-                 (topology == VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP) ||
-                 (topology == VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN) ||
-                 (topology == VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST_WITH_ADJACENCY) ||
-                 (topology == VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP_WITH_ADJACENCY))
+        // GS not present
+        if (hasTs)
```
nhaehnle: Indentation could be reduced here by one level by making it a chain: `if (hasGs) ... else if (hasTs) ... else ...`.

amdrexu: Thank you. Will correct this. For subgroupInclusiveAdd(), I also find this problem months ago. It is regretable we cannot use methods in LLPC builder in patch passes due to current design.

amdrexu:
retest this please

amdrexu:
I will land this.

PR #369, opened 2019-11-25
Use char[] instead of char* for global string constants

nhaehnle:
With char[], the compiler is not tempted to create the pointer as a real
object that would require a relocation by the runtime dynamic linker.

Most string constants already use char[], this change cleans up some places
that I found with a grep.

### DIFF
--- a/builder/llpcBuilderRecorder.h
+++ b/builder/llpcBuilderRecorder.h
@@ -41,10 +41,10 @@ namespace Llpc
 using namespace llvm;
 
 // Prefix of all recorded calls.
-static const char* const BuilderCallPrefix = "llpc.call.";
+static const char BuilderCallPrefix[] = "llpc.call.";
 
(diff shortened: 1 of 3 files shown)


kuhar:
FYI: there's also llvm::StringLiteral that may fit these uses: https://llvm.org/doxygen/classllvm_1_1StringLiteral.html

JaxLinAMD:
retest this please

JaxLinAMD:
retest this please

nhaehnle:
> FYI: there's also llvm::StringLiteral that may fit these uses: https://llvm.org/doxygen/classllvm_1_1StringLiteral.html

That's a good idea, actually, because it avoids implicit strlen()s at runtime.

Would be good if somebody else could pick that up, though -- this was just a quick drive-by commit of something that I happened to notice while reviewing something else.

PR #371, opened 2019-11-25
Tuning option for loop unroll threshold

tcorringham:
Add a shader tuning option "unrollThreshold" to allow the default loop unroll
threshold used by LLVM to be initialised to the specified value.
If a non-zero unrollThreshold is specified then it is passed to LLVM by use
of an attribute "amdgpu-unroll-threshold" that is attached to each function.
The unroll threshold used for each loop is then initialised to the specified
value before the target specific heuristics are applied, which may increase it.

### DIFF
--- a/tool/vfx/vfxSection.h
+++ b/tool/vfx/vfxSection.h
@@ -1190,13 +1190,15 @@ public:
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION >= 35
         INIT_STATE_MEMBER_NAME_TO_ADDR(SectionShaderOption, disableLicm, MemberTypeBool, false);
 #endif
+        INIT_STATE_MEMBER_NAME_TO_ADDR(SectionShaderOption, unrollThreshold, MemberTypeInt, false);
+
         VFX_ASSERT(pTableItem - &m_addrTable[0] <= MemberCount);
     }
 
     void GetSubState(SubState& state) { state = m_state; };
 
 private:
-    static const uint32_t  MemberCount = 15;
+    static const uint32_t  MemberCount = 16;
     static StrToMemberAddr m_addrTable[MemberCount];
 
     SubState               m_state;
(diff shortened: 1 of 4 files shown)


tool/vfx/vfxSection.h
```diff
@@ -1190,6 +1190,10 @@ class SectionShaderOption : public Section
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION >= 35
         INIT_STATE_MEMBER_NAME_TO_ADDR(SectionShaderOption, disableLicm, MemberTypeBool, false);
 #endif
+#if LLPC_CLIENT_INTERFACE_MAJOR_VERSION >= 39
+        INIT_STATE_MEMBER_NAME_TO_ADDR(SectionShaderOption, unrollThreshold, MemberTypeInt, false);
+#endif
+
```
amdrexu: I think you need to adjust the value of MemberCount
MemberCount = 15 -> MemberCount = 16

util/llpcPipelineDumper.cpp
```diff
@@ -640,6 +640,9 @@ void PipelineDumper::DumpPipelineShaderInfo(
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION >= 35
     dumpFile << "options.disableLicm = " << pShaderInfo->options.disableLicm << "\n";
 #endif
+#if LLPC_CLIENT_INTERFACE_MAJOR_VERSION >= 39
+    dumpFile << "options.unrollThreshold = " << pShaderInfo->options.unrollThreshold << "\n";
```
amdrexu: Please check if you update the hash in UpdateHashForPipelineShaderInfo()

tcorringham: Good spot - I'd missed that.

amdrexu:
See if we could avoid bumping the version. Do a internal pre-checkin testing. Codes are fine.

PR #374, opened 2019-11-26
llpc: Document policy of required zero-initialization of structures

nhaehnle:
The intent of this policy is to allow us to bump the interface major
version less frequently, and to avoid having to guard lots of code with
preprocessor macros that depend on the interface version.

Such guarding tends to be fragile because it explodes the matrix of
possible compilation options.

### DIFF
--- a/include/llpc.h
+++ b/include/llpc.h
@@ -66,6 +66,17 @@
 //* |     23.0 | Add flag robustBufferAccess in PipelineOptions to check out of bounds of private array.               |
 //* |     22.0 | Internal revision.                                                                                    |
 //* |     21.0 | Add stage in Pipeline shader info and struct PipelineBuildInfo to simplify pipeline dump interface.   |
+//*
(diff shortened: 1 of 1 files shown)


include/llpc.h
```diff
@@ -66,6 +66,17 @@
 //* |     23.0 | Add flag robustBufferAccess in PipelineOptions to check out of bounds of private array.               |
 //* |     22.0 | Internal revision.                                                                                    |
 //* |     21.0 | Add stage in Pipeline shader info and struct PipelineBuildInfo to simplify pipeline dump interface.   |
+//*
+//* Important note: All structures defined in this file that are passed as input into LLPC must be zero-initialized
```
amdrexu: I prefer to use uppercase to highligh it just like what we have done in source codes: NOTE, TODO

Important note -> IMPORTANT NOTE

nhaehnle: Yeah, that makes sense, I'm going to change it.

JaxLinAMD:
retest this please

PR #375, opened 2019-11-26
RFC: docs: Add a design note for better post-pipeline-optimization cache checks

nhaehnle:


docs/DdnInterShaderDataCacheTracking.md
```diff
@@ -0,0 +1,87 @@
...
+     the information may otherwise be obsolete.
+
+  2. Knowledge about what kind of full pipeline optimizations are applied, including some of their details, is
+     centralized in `PatchCheckShaderCache`. This limits the design's extensibility.
+
+
+Implementation overview
+-----------------------
+
+Every function in LLPC gets an attached `!llpc.hash` metadata node.
```
linqun: Until now, we only build hash for fragment shader and non-fragment shader. it is not very easy to build hash for every stage.
We plan to support partial pipeline compile, in this optimization, we need two kinds of hash
- current shader hash
- current shader hash and ignore the resource node offset in root level.
i.e. we need support multiple hash in !llpc.hash

docs/DdnInterShaderDataCacheTracking.md
```diff
@@ -0,0 +1,87 @@
...
+The plan is to define an LLVMDialect abstract base class in LLVM core. Dialects can be registered with an LLVMContext,
+and all of the context's associated printing, parsing, reading and writing activity will then consult the dialect to
+determine whether some metadata object is to be interpreted via a custom C++ class.
+
+The notion of a "dialect" then provides a beachhead for further extensibility. Most importantly for LLPC, this
+includes the ability to register additional intrinsic functions and build infrastructure for automatically generating
+C++ facade classes for interacting with calls to those intrinsic functions as if they were native instructions.
+This has at least the following benefits:
+
+  1. We will be able to use checks such as `isa<InputImportInst>` instead of performing a string comparison
```
linqun: we have add a ID in metadata for all LLPC builder functions, we should check this ID instead of function name.
please feel free to change it if you have some free time.

docs/DdnInterShaderDataCacheTracking.md
nhaehnle: Can you explain this more? What do you mean by ignoring the resource node offset in root level?

docs/DdnInterShaderDataCacheTracking.md
nhaehnle: That's not a bad idea as long as we assume that core LLVM infrastructure cannot be changed. A few things, though:

- This doesn't just apply to builder functions, but there may be other intermediate functions to which this applies.
- It still relies on digging through metadata, which has its own overhead.

So I'm not against what you're saying, by all means get it done if you can assign somebody to it, but I'm trying to help us find a better longer term solution -- which is less certain because it requires buy-in from the LLVM community, but it does make sense to also pursue such things.

docs/DdnInterShaderDataCacheTracking.md
linqun: in partial pipeline compile, we want to ignore resource node offset mismatch in root level between cached pipeline and input pipeline, because we can adjust user data register mapping on final ELF binary for offset mismatch, instead of trig full pipeline compile. you can find the detail in our previous internal discussion "Vulkan DDN for auto-layout pipeline".

docs/DdnInterShaderDataCacheTracking.md
linqun: Got it. thanks!

docs/DdnInterShaderDataCacheTracking.md
nhaehnle: Ah I see. I still wonder about having two different hashes, though.

Generally, the principle that shader caching is built on is "equal input --> equal output".

Now we're talking about a situation where maybe "different input --> equal output". That sounds like a fragile and risky design. If it is really true, then it should be possible to make the input smaller instead.

So if previously, the situation was "(shader 0, resource node offsets A) --> output" and "(shader 0, resource node offsets B) --> same output", then I would argue that the resource node offsets shouldn't be part of the input to begin with.

This makes conceptual sense if you want to fixup the resource node offsets via ELF relocations in the resulting binary: there is simply no need to feed the resource node offsets into the compiler in that case, since a relocation will be generated anyway.

linqun: there is simply no need to feed the resource node offsets into the compiler in that case, since a relocation will be generated anyway.
- I am afraid we can't use relocation for resource node offsets, until now, resource node offset mapping is a part of PAL metadata in .note section. I don't think PAL support reloc in this section.

for partial pipeline compile, the situation is
(shader 0, resource node offsets A) --> output A
(shader 0, no offset ) --> output B
we still follow equal input --> equal output, but in second case, we need modify the output B a little as the final output.

nhaehnle: Wouldn't it then make sense to do the "little modification" of the output in _all_ cases?

Then in all cases it could be "(shader 0, no offset) --> output B", and in all cases the output B would be fixed up at the end.

amdrexu:
retest this please

PR #376, opened 2019-11-27
Refactor copy shader by simplifying GS-VS ring data loading

amdrexu:
- Extend LoadValueFromGsVsRing() to handle array type and vector type.
  Previously, it only handles scalar type and is not convenient for our
  uses. Also, on-chip LDS read can handle such types in LLVM backend
  without scalarization. The LDS data is continuous, different from
  off-chip buffer store.

- Move NGG load calls to LoadValueFromGsVsRing(). There is no need of
  leaving it outside this function. Such details should be encapsulated.

- Simplify ExportOutput(). NGG handling and non NGG handling are
  merged. Also, built-in output exports are simplified as well.


amdrexu:
retest this please

PR #433, opened 2020-01-17
Add ddn of PartialPipelineComiple

amingriyue:
This DDN proposes a method to build pipeline layout from SPIRV binary and make it compatible with APP’s pipeline layout without recompile.


nhaehnle:
Is the plan to re-open this with a new version soon?

PR #444, opened 2020-01-30
Refactor buffer loads

Flakebi:
Previously, buffer.load instructions were emitted for small loads but
raw buffer loads should be used instead.
This patch takes the chance and refactors the loading code.

Effect:
- Increase compilation speed by skipping unnecessary vector creations
- Use DWORDx3 loads when possible

### DIFF
--- a/patch/llpcPatchBufferOp.cpp
+++ b/patch/llpcPatchBufferOp.cpp
@@ -1533,7 +1533,7 @@ Value* PatchBufferOp::ReplaceLoad(
 
     Type* const pLoadType = pLoadInst->getType();
 
-    const uint32_t bytesToLoad = static_cast<uint32_t>(dataLayout.getTypeSizeInBits(pLoadType) / 8);
+    const uint32_t bytesToLoad = static_cast<uint32_t>(dataLayout.getTypeStoreSize(pLoadType));
 
     uint32_t alignment = pLoadInst->getAlignment();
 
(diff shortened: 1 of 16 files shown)


amdvlk-admin:

patch/llpcPatchBufferOp.cpp
```diff
@@ -1531,15 +1531,15 @@ Value* PatchBufferOp::ReplaceLoad(
 
     const DataLayout& dataLayout = m_pBuilder->GetInsertBlock()->getModule()->getDataLayout();
 
-    Type* const pLoadType = pLoadInst->getType();
+    Type* const loadType = pLoadInst->getType();
```
amdrexu: loadType -> pLoadType
Current style still add "p" prefix for pointer variables.

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
     }
 
     SmallVector<Value*, 8> partLoads;
+    Type* smallestType;
```
amdrexu: smallestType -> pSmallestType = nullptr
Initialize all local variables.

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
...
-                                   m_pBuilder->CreateAdd(pBaseIndex, m_pBuilder->getInt32(offset));
+        const uint32_t offset = bytesToLoad - remainingBytes;
+        Value* pOffset = (offset == 0) ?
+                               pBaseIndex :
+                               m_pBuilder->CreateAdd(pBaseIndex, m_pBuilder->getInt32(offset));
 
-            Type* pIntLoadType = nullptr;
-            Type* pFloatLoadType = nullptr;
+        Type* pIntLoadType = nullptr;
+        uint32_t loadSize;
```
amdrexu: loadSize -> loadSize = 0

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
...
+            loadSize = 2;
+        }
+        else
+        {
+            pIntLoadType = Type::getInt8Ty(*m_pContext);
+            loadSize = 1;
+        }
 
-            Value* pPartLoad = nullptr;
+        Value* partLoad = nullptr;
```
amdrexu: partLoad -> pPartLoad
This is a coding style change. Please drop it. Keep current style. Tim will do this with dedicated commits in the future.

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
...
-
-    // And create an undef vector whose total size is the number of bytes we loaded.
-    Value* pNewLoad = UndefValue::get(VectorType::get(pSmallestType, bytesToLoad / smallestByteSize));
+    smallestByteSize = static_cast<IntegerType*>(smallestType)->getBitWidth() / 8;
 
-    uint32_t index = 0;
-
-    for (Value* pPartLoad : partLoads)
+    // This will be either the loaded value or a vector of the loaded values
+    Value* newLoad;
```
amdrexu: newLoad -> pNewLoad = nullptr

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
...
 
-        // Bitcast it to a vector of the smallest load type.
-        VectorType* const pCastType = VectorType::get(pSmallestType, byteSize / smallestByteSize);
-        pPartLoad = m_pBuilder->CreateBitCast(pPartLoad, pCastType);
-        CopyMetadata(pPartLoad, pLoadInst);
+        uint32_t index = 0;
 
-        // Run through our the elements of our bitcasted type and insert them into the main load.
-        for (uint32_t i = 0, compCount = static_cast<uint32_t>(pCastType->getNumElements()); i < compCount; i++)
+        for (Value* partLoad : partLoads)
```
amdrexu: partLoad -> pPartLoad

patch/llpcPatchBufferOp.cpp
```diff
@@ -1531,15 +1531,15 @@ Value* PatchBufferOp::ReplaceLoad(
 
     const DataLayout& dataLayout = m_pBuilder->GetInsertBlock()->getModule()->getDataLayout();
 
-    Type* const pLoadType = pLoadInst->getType();
+    Type* const loadType = pLoadInst->getType();
 
-    const uint32_t bytesToLoad = static_cast<uint32_t>(dataLayout.getTypeSizeInBits(pLoadType) / 8);
+    const uint32_t bytesToLoad = static_cast<uint32_t>(dataLayout.getTypeSizeInBits(loadType) / 8);
```
kuhar: It should be safer to use datalayout.getTypeStoreSize

patch/llpcPatchBufferOp.cpp
```diff
@@ -1595,192 +1595,159 @@ Value* PatchBufferOp::ReplaceLoad(
...
-        // Run through our the elements of our bitcasted type and insert them into the main load.
-        for (uint32_t i = 0, compCount = static_cast<uint32_t>(pCastType->getNumElements()); i < compCount; i++)
+        for (Value* partLoad : partLoads)
         {
-            Value* const pLoadElem = m_pBuilder->CreateExtractElement(pPartLoad, i);
-            CopyMetadata(pLoadElem, pLoadInst);
-            pNewLoad = m_pBuilder->CreateInsertElement(pNewLoad, pLoadElem, index++);
-            CopyMetadata(pNewLoad, pLoadInst);
+            // Get the byte size of our load part.
+            const uint32_t byteSize = static_cast<uint32_t>(dataLayout.getTypeSizeInBits(partLoad->getType()) / 8);
```
kuhar: same as above

patch/llpcPatchBufferOp.cpp
nhaehnle: I don't think this is a good idea. If the variable is not assigned to, it is a bug; initializing the variable to 0 only obfuscates the bug.

For example, if we keep the definition as-is, then a good compiler will warn if a future change forgets to initialize loadSize on all code paths. Conversely, if loadSize is initialized as 0, then such a bug will not be detected by a compiler warning.

patch/llpcPatchBufferOp.cpp
nhaehnle: I don't think the pPrefix review comments are helpful. We agreed that we would go away from this style.

patch/llpcPatchBufferOp.cpp
amdrexu: You can not always count on compiler warning because it is configurable and couldn't be disabled manually. We have encountered many problems caused by uninitialized local variables and it often occured in release driver. We spend lots of developer resources reproducing such issues. Sometimes, it is difficult to pinpoint the root cause because of the intermittent occurence. Sometimes, it even depends on OS, making the reproduction of such bugs more challenging. I understand your purpose. So we should add asserts in any place to verify our assumption.

patch/llpcPatchBufferOp.cpp
amdrexu: Yes, we are OK without "p" prefix. Currently, as discussed, recently added codes follow the style of surrounding codes. This is just to keep coding consistency, not making new commers feel confusing which style to follow in a mixed environment.
I think Tim will start style transition soon. It will not be a problem in the near future.

Flakebi:
Thanks for the reviews, I addressed your comments.

amdvlk-admin:
### Test summary for commit 577eca1e7535201a90f6f8a82f65f412418a3456

<li>XGL: dev c0791473c087f5b5d75a21c6f2ace72ccd232a65</li>
</ul></details>

<li>Passed: 939/939 (100.0%)</li>
<li>Failed: 0/939 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

patch/llpcPatchBufferOp.cpp
kuhar: @Flakebi this seems unchanged but was marked as resolved

kuhar:
> Thanks for the reviews, I addressed your comments.
Some comments seem to have been erroneously marked as resolved. @Flakebi could you take a second look?

PR #454, opened 2020-02-10
Unify buffer loads and stores

Flakebi:
The ReplaceLoad and ReplaceStore functions in llpcPatchBufferOp
shared many parts of their code.
This patch unifies them into one function.

Additionally, the previous refactoring of loads now benefits the store
replacement: Stores of v3i32s are now suported and store.byte and
store.ushort are not emitted anymore.


amdvlk-admin:

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1521,31 +1520,53 @@ bool PatchBufferOp::RemoveUsersForInvariantStarts(
...
 
 // =====================================================================================================================
-// Replace a fat pointer load with the intrinsics required to do the load.
-Value* PatchBufferOp::ReplaceLoad(
-    LoadInst* const pLoadInst) // [in] The load instruction to replace.
+// Replace a fat pointer load or store with the required intrinsics.
+Value* PatchBufferOp::ReplaceLoadStore(
+    bool isLoad,
+    LoadInst* const pLoadInst, // [in] The load instruction to replace.
+    StoreInst* const pStoreInst) // [in] The store instruction to replace.
```
kuhar: Have you considered taking a single argument as `llvm::Instruction` and then checking whether it's a load or store in the function body?

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1564,512 +1585,279 @@ Value* PatchBufferOp::ReplaceLoad(
...
-        pNewLoad->setSyncScopeID(pLoadInst->getSyncScopeID());
-        CopyMetadata(pNewLoad, pLoadInst);
+        pPointer = m_pBuilder->CreateBitCast(pPointer, pType->getPointerTo(ADDR_SPACE_GLOBAL));
 
-        if (isInvariant)
+        if (isLoad)
         {
-            pNewLoad->setMetadata(LLVMContext::MD_invariant_load, MDNode::get(*m_pContext, None));
+            LoadInst* const pNewLoad = m_pBuilder->CreateLoad(pPointer);
+            pNewLoad->setVolatile(pLoadInst->isVolatile());
```
kuhar: `volatile`, ordering, etc. can checking can be hoisted to top of the function; it's already done for alignment and ordering.

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1564,512 +1585,279 @@ Value* PatchBufferOp::ReplaceLoad(
...
     {
-        const uint32_t offset = bytesToLoad - remainingBytes;
+        const uint32_t offset = bytesToHandle - remainingBytes;
         Value* pOffset = (offset == 0) ?
                                pBaseIndex :
                                m_pBuilder->CreateAdd(pBaseIndex, m_pBuilder->getInt32(offset));
 
-        Type* pIntLoadType = nullptr;
-        uint32_t loadSize = 0;
+        Type* pIntType = nullptr;
```
kuhar: nit: maybe `pIntMemAccessType`? or `pIntAccessType`

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1564,512 +1585,279 @@ Value* PatchBufferOp::ReplaceLoad(
...
+        uint32_t size = 0;
```
kuhar: nit: maybe `memAccessSize` or `accessSize`?

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1564,512 +1585,279 @@ Value* PatchBufferOp::ReplaceLoad(
...
         }
         else
         {
-            pIntLoadType = Type::getInt8Ty(*m_pContext);
-            loadSize = 1;
+            pIntType = Type::getInt8Ty(*m_pContext);
+            size = 1;
         }
-        LLPC_ASSERT(loadSize != 0);
+        LLPC_ASSERT(size != 0);
```
kuhar: This should also check that `pIntType` is not null, or the base case when size=1 and type=i8 could be moved to the top.

llpc/patch/llpcPatchBufferOp.cpp
Flakebi: I considered this, it would definitely look better as the function signature. However, it allows inserting arbitrary instructions, not only loads and stores. With the current one, the compiler checks if the instruction is actually a `LoadInst` or `StoreInst`.

That’s why I used this signature, I’m not sure which of the two possibilities is better.

kuhar: Makes sense to me. But then you should check which one of the two is not null and don't really need the bool, right? Otherwise callers may mistakenly pass both or use the boolean inconsistently with the pointer arguments.

kuhar:
> Would it be possible to split this into two function, one that deals with the metadata and memory access properties like alignment, types, metadata, etc, and a second function that decides how to lower the memory instruction into smaller accesses? Do you think it would simplify the code?

kuhar:
> Would it be possible to split this into two function, one that deals with the metadata and memory access properties like alignment, types, metadata, etc, and a second function that decides how to lower the memory instruction into smaller accesses?

I'm somewhat concerned that the part that decides how to lay out memory accesses based on types, sizes, and alignments is non-trivial and am not sure if the existing shaderdb provide enough coverage. If this was a more stand-alone function, it would be possible to test it in isolation if there's such a need.

llpc/patch/llpcPatchBufferOp.cpp
Flakebi: Good catch. I changed it to use the variables from above.
`isVolatile` is only used in this part, so I didn’t want to hoist it.

Flakebi:
Thanks for the good review @kuhar!
I hope I was able to address your comments.

> Would it be possible to split this into two function, one that deals with the metadata and memory access properties like alignment, types, metadata, etc, and a second function that decides how to lower the memory instruction into smaller accesses?

If I understand this right, you mean splitting the following lines into a new function?
https://github.com/GPUOpen-Drivers/llpc/blob/b3c6dbd21cc33480af7c3cbc4a167e84ab0b7760/llpc/patch/llpcPatchBufferOp.cpp#L1635-L1725

This is ~60 lines (not counting load/store specific lines) out of the 300 lines of this function.
To pull this out into a new function, I would return a `vector<Type>`.
Then, in the Replace function, iterate over the vector and create loads/stores.
If we want to test this, it could work.
Apart from that benifit it adds overhead, code and performance wise, so I think it is not worth it in this form if we do not add tests for this specific function.

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1521,31 +1520,58 @@ bool PatchBufferOp::RemoveUsersForInvariantStarts(
 }
 
 // =====================================================================================================================
-// Replace a fat pointer load with the intrinsics required to do the load.
-Value* PatchBufferOp::ReplaceLoad(
-    LoadInst* const pLoadInst) // [in] The load instruction to replace.
+// Replace a fat pointer load or store with the required intrinsics.
+Value* PatchBufferOp::ReplaceLoadStore(
```
kuhar: Can the return value not be an `Instruction`? If not, consider changing this to return `Instruction`.

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1564,512 +1590,274 @@ Value* PatchBufferOp::ReplaceLoad(
...
         {
-            pNewLoad->setMetadata(LLVMContext::MD_invariant_load, MDNode::get(*m_pContext, None));
+            LoadInst* const pNewLoad = m_pBuilder->CreateLoad(pPointer);
+            pNewLoad->setVolatile(pLoadInst->isVolatile());
+            pNewLoad->setAlignment(MaybeAlign(alignment));
+            pNewLoad->setOrdering(ordering);
+            pNewLoad->setSyncScopeID(syncScopeID);
+            CopyMetadata(pNewLoad, pLoadInst);
+
+            if (isInvariant)
```
kuhar: I think the current coding style mandates using braces for all if statements. But @amdrexu is an expert here ;)

kuhar:
> > Would it be possible to split this into two function, one that deals with the metadata and memory access properties like alignment, types, metadata, etc, and a second function that decides how to lower the memory instruction into smaller accesses?
> 
> If I understand this right, you mean splitting the following lines into a new function?
> https://github.com/GPUOpen-Drivers/llpc/blob/b3c6dbd21cc33480af7c3cbc4a167e84ab0b7760/llpc/patch/llpcPatchBufferOp.cpp#L1635-L1725

Right, this is was I was referring to.
 
> This is ~60 lines (not counting load/store specific lines) out of the 300 lines of this function.
> To pull this out into a new function, I would return a `vector<Type>`.
> Then, in the Replace function, iterate over the vector and create loads/stores.
> If we want to test this, it could work.
> Apart from that benifit it adds overhead, code and performance wise, so I think it is not worth it in this form if we do not add tests for this specific function.

I don't think it's that much overhead in terms of code lines, but at the same time agree that there may not be enough benefit to it if this logic is well-exercised already. Thanks for clarifying!

The code looks fine to me.

llpc/patch/llpcPatchBufferOp.cpp
Flakebi: Seems like `BitCast` and `InsertElement` only generate `Value`s, not `Instruction`s, so this has to return a `Value`.

llpc/patch/llpcPatchBufferOp.cpp
amdrexu: Yes, current style always adds {} for single-line statement.

Flakebi:
@amdrexu I added the braces

amdrexu:
In the future, once you fix review issues, please click "Resolve Conversation" buttons. Thus, we can check what is still left unresolved and is under discussion. Thanks :)

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1521,31 +1520,63 @@ bool PatchBufferOp::RemoveUsersForInvariantStarts(
...
+    }
+    else
+    {
+        pInst = pStoreInst;
+        pType = pStoreInst->getValueOperand()->getType();
+        pPointerOperand = pStoreInst->getPointerOperand();
+        ordering = pStoreInst->getOrdering();
+        alignment = pStoreInst->getAlignment();
+        syncScopeID = pStoreInst->getSyncScopeID();
+    }
```
nhaehnle: Please instead just pass pInst as an Instruction* to the function, and use dyn_cast here. Note that dyn_cast (and cast, and isa) is very fast as it does not use RTTI.

llpc/patch/llpcPatchBufferOp.cpp
```diff
@@ -1521,31 +1520,62 @@ bool PatchBufferOp::RemoveUsersForInvariantStarts(
...
 // =====================================================================================================================
-// Replace a fat pointer load with the intrinsics required to do the load.
-Value* PatchBufferOp::ReplaceLoad(
-    LoadInst* const pLoadInst) // [in] The load instruction to replace.
+// Replace a fat pointer load or store with the required intrinsics.
+Value* PatchBufferOp::ReplaceLoadStore(
+    Instruction* const pInst) // [in] The instruction to replace.
 {
-    m_pBuilder->SetInsertPoint(pLoadInst);
+    LoadInst* const pLoadInst = dyn_cast<LoadInst>(pInst);
```
kuhar: pInst may be null. You could take it as a reference and not worry at all, or assert at the very beginning, or use dyn_cast_or_null

amdrexu:
@nhaehnle You can check if the update commit meets your review comments. If yes, you need to re-approve it to unblock the merging process. Otherwise, it will always stay in blocked status.

JaxLinAMD:
retest this please

amdvlk-admin:
### Test summary for commit 4c411fb7d68a0869f2fce75d5033f4c06e522322

<li>XGL: dev 052ebfa2e77abd14fdb4f5120b35b081f3fc4bdc</li>
</ul></details>

<li>Passed: 942/942 (100.0%)</li>
<li>Failed: 0/942 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

PR #438, opened 2020-01-21
Initial code for reloctable shader elf

s-perron:
Part of https://github.com/GPUOpen-Drivers/llpc/issues/431.

This code will allows VsPs shader to have the two shader compiled
separetly and then be linked together.  It disables cross shader
optimizations like the removal of unused outputs from the vertex
shaders.

PR #457, opened 2020-02-10
Make Descriptor Offsets Relocatable

csyonghe:
This PR is based on PR #440.

The commit is based on the original changes made by s-perron
that emits a relocatable entry for each descriptor load, instead
of baking in a constant offset during initial compilation. The
compiler patches up the compiled binary with actual descriptor
offset at a later time when that info is available.

### DIFF
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -734,15 +734,15 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
     IShaderCache *userShaderCache = nullptr;
     if (context->isGraphics()) {
       auto pipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(context->getPipelineBuildInfo());
-      cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, stage);
+      cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-      pUserShaderCache = pPipelineInfo->pShaderCache;
+      userShaderCache = pipelineInfo->pShaderCache;
 #endif
     } else {
       auto pipelineInfo = reinterpret_cast<const ComputePipelineBuildInfo *>(context->getPipelineBuildInfo());
-      cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true);
+      cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true, true);
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-      pUserShaderCache = pPipelineInfo->pShaderCache;
+      userShaderCache = pipelineInfo->pShaderCache;
 #endif
     }
 
@@ -1083,8 +1083,8 @@ unsigned GraphicsShaderCacheChecker::check(const Module *module, unsigned stageM
 
   IShaderCache *appCache = nullptr;
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-  auto pPipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(m_pContext->GetPipelineBuildInfo());
-  pAppCache = pPipelineInfo->pShaderCache;
+  auto pPipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(m_context->getPipelineBuildInfo());
+  appCache = pPipelineInfo->pShaderCache;
 #endif
   if (stageMask & shaderStageToMask(ShaderStageFragment)) {
     m_fragmentCacheEntryState = m_compiler->lookUpShaderCaches(appCache, &fragmentHash, &m_fragmentElf,
@@ -1115,7 +1115,7 @@ unsigned GraphicsShaderCacheChecker::check(const Module *module, unsigned stageM
 void GraphicsShaderCacheChecker::updateRootUserDateOffset(ElfPackage *pipelineElf) {
   ElfWriter<Elf64> writer(m_context->getGfxIpVersion());
   // Load ELF binary
-  auto result = writer.ReadFromBuffer(pipelineElf->data(), pipelineElf->size());
+  auto result = writer.readFromBuffer(pipelineElf->data(), pipelineElf->size());
   assert(result == Result::Success);
   (void(result)); // unused
   writer.updateElfBinary(m_context, pipelineElf);
@@ -1171,7 +1171,7 @@ void GraphicsShaderCacheChecker::updateAndMerge(Result result, ElfPackage *outpu
 
     // Merge and store the result in pPipelineElf
     ElfWriter<Elf64> writer(m_context->getGfxIpVersion());
-    auto result = writer.ReadFromBuffer(nonFragmentElf.pCode, nonFragmentElf.codeSize);
+    auto result = writer.readFromBuffer(nonFragmentElf.pCode, nonFragmentElf.codeSize);
     assert(result == Result::Success);
     (void(result)); // unused
     writer.mergeElfBinary(m_context, &fragmentElf, outputPipelineElf);
@@ -1243,14 +1243,15 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
   const PipelineShaderInfo *shaderInfo[ShaderStageGfxCount] = {
       &pipelineInfo->vs, &pipelineInfo->tcs, &pipelineInfo->tes, &pipelineInfo->gs, &pipelineInfo->fs,
   };
+  bool buildingRelocatableElf = canUseRelocatableGraphicsShaderElf(shaderInfo);
 
   for (unsigned i = 0; i < ShaderStageGfxCount && result == Result::Success; ++i)
     result = validatePipelineShaderInfo(shaderInfo[i]);
 
   MetroHash::Hash cacheHash = {};
   MetroHash::Hash pipelineHash = {};
-  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true);
-  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, false);
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, buildingRelocatableElf);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, false, false);
 
   if (result == Result::Success && EnableOuts()) {
     LLPC_OUTS("===============================================================================\n");
@@ -1277,10 +1278,9 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
   }
 
   ShaderEntryState cacheEntryState = ShaderEntryState::New;
-  bool buildingRelocatableElf = canUseRelocatableGraphicsShaderElf(shaderInfo);
   IShaderCache *appCache = nullptr;
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-  pAppCache = pPipelineInfo->pShaderCache;
+  appCache = pipelineInfo->pShaderCache;
 #endif
   ShaderCache *shaderCache = nullptr;
   CacheEntryHandle hEntry = nullptr;
@@ -1371,8 +1371,8 @@ Result Compiler::BuildComputePipeline(const ComputePipelineBuildInfo *pipelineIn
 
   MetroHash::Hash cacheHash = {};
   MetroHash::Hash pipelineHash = {};
-  cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true);
-  pipelineHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, false);
+  cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true, buildingRelocatableElf);
+  pipelineHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, false, buildingRelocatableElf);
 
   if (result == Result::Success && EnableOuts()) {
     const ShaderModuleData *moduleData = reinterpret_cast<const ShaderModuleData *>(pipelineInfo->cs.pModuleData);
@@ -1397,7 +1397,7 @@ Result Compiler::BuildComputePipeline(const ComputePipelineBuildInfo *pipelineIn
   ShaderEntryState cacheEntryState = ShaderEntryState::New;
   IShaderCache *appCache = nullptr;
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-  pAppCache = pPipelineInfo->pShaderCache;
+  appCache = pipelineInfo->pShaderCache;
 #endif
   ShaderCache *shaderCache = nullptr;
   CacheEntryHandle hEntry = nullptr;
@@ -1537,7 +1537,7 @@ Result Compiler::CreateShaderCache(const ShaderCacheCreateInfo *pCreateInfo, // 
   ShaderCache *pShaderCache = new ShaderCache();
 
   if (pShaderCache != nullptr) {
-    result = pShaderCache->Init(pCreateInfo, &auxCreateInfo);
+    result = pShaderCache->init(pCreateInfo, &auxCreateInfo);
     if (result != Result::Success) {
       pShaderCache->Destroy();
       pShaderCache = nullptr;
@@ -1716,7 +1716,7 @@ void Compiler::buildShaderCacheHash(Context *context, unsigned stageMask, ArrayR
     MetroHash64 hasher;
 
     // Update common shader info
-    PipelineDumper::updateHashForPipelineShaderInfo(stage, shaderInfo, true, &hasher);
+    PipelineDumper::updateHashForPipelineShaderInfo(stage, shaderInfo, true, &hasher, false);
     hasher.Update(pipelineInfo->iaState.deviceIndex);
 
     // Update input/output usage (provided by middle-end caller of this callback).
@@ -1774,14 +1774,14 @@ void Compiler::linkRelocatableShaderElf(ElfPackage *shaderElfs, ElfPackage *pipe
     ElfReader<Elf64> fsReader(m_gfxIp);
     if (!shaderElfs[ShaderStageVertex].empty()) {
       size_t codeSize = shaderElfs[ShaderStageVertex].size_in_bytes();
-      result = vsReader.ReadFromBuffer(shaderElfs[ShaderStageVertex].data(), &codeSize);
+      result = vsReader.readFromBuffer(shaderElfs[ShaderStageVertex].data(), &codeSize);
       if (result != Result::Success)
         return;
     }
 
     if (!shaderElfs[ShaderStageFragment].empty()) {
       size_t codeSize = shaderElfs[ShaderStageFragment].size_in_bytes();
-      result = fsReader.ReadFromBuffer(shaderElfs[ShaderStageFragment].data(), &codeSize);
+      result = fsReader.readFromBuffer(shaderElfs[ShaderStageFragment].data(), &codeSize);
       if (result != Result::Success)
         return;
     }
@@ -1790,7 +1790,7 @@ void Compiler::linkRelocatableShaderElf(ElfPackage *shaderElfs, ElfPackage *pipe
   } else {
     ElfReader<Elf64> csReader(m_gfxIp);
     size_t codeSize = shaderElfs[ShaderStageCompute].size_in_bytes();
-    result = csReader.ReadFromBuffer(shaderElfs[ShaderStageCompute].data(), &codeSize);
+    result = csReader.readFromBuffer(shaderElfs[ShaderStageCompute].data(), &codeSize);
     if (result != Result::Success)
       return;
     result = writer.linkComputeRelocatableElf(csReader, context);
--- a/llpc/test/shaderdb/PipelineCs_RelocConst.pipe
+++ b/llpc/test/shaderdb/PipelineCs_RelocConst.pipe
@@ -1,0 +1,59 @@
+; This test case checks that descriptor offset relocation works for buffer descriptors in a compute pipeline.
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -o %t.elf %gfxip %s && llvm-objdump --arch=amdgcn --mcpu=gfx900 -d %t.elf | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: 0000000000000000 <_amdgpu_cs_main>:
+; SHADERTEST: s_mov_b32 s[[RELOREG:[0-9]+]], 16 {{.*}}
+; SHADERTEST: s_load_dwordx4 s[{{.*}}:{{.*}}], s[{{.*}}:{{.*}}], s[[RELOREG]] //{{.*}}
+; END_SHADERTEST
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -v %gfxip %s | FileCheck -check-prefix=SHADERTEST1 %s
+; SHADERTEST1-LABEL: {{^// LLPC}} pipeline patching results
+; SHADERTEST1-DAG: ![[METADATANODE:[0-9]+]] = !{!"doff_0_0_b"}
+; SHADERTEST1-DAG: %[[RELOCONST:[0-9]+]] = call i32 @llvm.amdgcn.reloc.constant(metadata ![[METADATANODE]])
+; SHADERTEST1: %[[RELOCONSTEXT:[0-9]+]] = zext i32 %[[RELOCONST]] to i64
+; SHADERTEST1: %[[BUFFFERDESCADDR:[0-9]+]] = add i64 %{{[0-9]+}}, %[[RELOCONSTEXT]]
+; SHADERTEST1: %[[BUFFERDESCPTR:[0-9]+]] = inttoptr i64 %[[BUFFFERDESCADDR]] {{.*}}
+; SHADERTEST1: %{{[0-9]+}} = load <4 x i32>, <4 x i32> addrspace(4)* %[[BUFFERDESCPTR]], align 16, !invariant.load !{{.*}}
+; SHADERTEST1: AMDLLPC SUCCESS
+; END_SHADERTEST
+
+[CsGlsl]
+#version 450
+#extension GL_ARB_separate_shader_objects : enable
+
+layout(binding = 0) uniform UniformBufferObject {
+    vec4 i;
+} ubo;
+
+layout(set = 1, binding = 0, std430) buffer OUT
+{
+    vec4 o;
+};
+
+layout(local_size_x = 2, local_size_y = 3) in;
+void main() {
+    o = ubo.i;
+}
+
+
+[CsInfo]
+entryPoint = main
+userDataNode[0].type = DescriptorTableVaPtr
+userDataNode[0].offsetInDwords = 0
+userDataNode[0].sizeInDwords = 1
+userDataNode[0].set = 0
+userDataNode[0].next[0].type = DescriptorBuffer
+userDataNode[0].next[0].offsetInDwords = 4
+userDataNode[0].next[0].sizeInDwords = 8
+userDataNode[0].next[0].set = 0
+userDataNode[0].next[0].binding = 0
+userDataNode[1].type = DescriptorTableVaPtr
+userDataNode[1].offsetInDwords = 1
+userDataNode[1].sizeInDwords = 1
+userDataNode[1].set = 1
+userDataNode[1].next[0].type = DescriptorBuffer
+userDataNode[1].next[0].offsetInDwords = 4
+userDataNode[1].next[0].sizeInDwords = 8
+userDataNode[1].next[0].set = 1
+userDataNode[1].next[0].binding = 0
--- a/llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
+++ b/llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
@@ -1,0 +1,122 @@
+
+// This test case checks that descriptor offset relocation works for buffer descriptors in a vs/fs pipeline.
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -o %t.elf %gfxip %s && llvm-objdump --triple=amdgcn --mcpu=gfx900 -d %t.elf | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: 0000000000000000 <_amdgpu_vs_main>:
+; SHADERTEST: s_mov_b32 s[[RELOREG:[0-9]+]], 12 //{{.*}}
+; SHADERTEST: s_load_dwordx4 s[{{.*}}:{{.*}}], s[{{.*}}:{{.*}}], s[[RELOREG]] //{{.*}}
+; SHADERTEST: {{[0-9A-Za-z]+}} <_amdgpu_ps_main>:
+; SHADERTEST: s_mov_b32 s[[RELOREG1:[0-9]+]], 12 //{{.*}}
+; SHADERTEST: s_load_dwordx4 s[{{.*}}:{{.*}}], s[{{.*}}:{{.*}}], s[[RELOREG1]] //{{.*}}
+; END_SHADERTEST
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -v %gfxip %s | FileCheck -check-prefix=SHADERTEST1 %s
+; SHADERTEST1-LABEL: {{^// LLPC}} pipeline patching results
+; SHADERTEST1-DAG: ![[METADATANODE:[0-9]+]] = !{!"doff_0_0_b"}
+; SHADERTEST1-DAG: %[[RELOCONST:[0-9]+]] = call i32 @llvm.amdgcn.reloc.constant(metadata ![[METADATANODE]])
+; SHADERTEST1: %[[RELOCONSTEXT:[0-9]+]] = zext i32 %[[RELOCONST]] to i64
+; SHADERTEST1: %[[BUFFFERDESCADDR:[0-9]+]] = add i64 %{{[0-9]+}}, %[[RELOCONSTEXT]]
+; SHADERTEST1: %[[BUFFERDESCPTR:[0-9]+]] = inttoptr i64 %[[BUFFFERDESCADDR]] {{.*}}
+; SHADERTEST1: %{{[0-9]+}} = load <4 x i32>, <4 x i32> addrspace(4)* %[[BUFFERDESCPTR]], align 16, !invariant.load !{{.*}}
+; SHADERTEST1: AMDLLPC SUCCESS
+; END_SHADERTEST
+
+[VsGlsl]
+#version 450
+#extension GL_ARB_separate_shader_objects : enable
+
+layout(binding = 0) uniform UniformBufferObject {
+    mat4 model;
+    mat4 view;
+    vec4 proj;
+} ubo;
+
+layout(location = 0) in vec2 inPosition;
+layout(location = 1) in vec3 inColor;
+
+layout(location = 0) out vec3 fragColor;
+
+void main() {
+    gl_Position = ubo.proj;
+    fragColor = inColor;
+}
+
+
+[VsInfo]
+entryPoint = main
+userDataNode[0].type = IndirectUserDataVaPtr
+userDataNode[0].offsetInDwords = 0
+userDataNode[0].sizeInDwords = 1
+userDataNode[0].indirectUserDataCount = 0
+userDataNode[1].type = DescriptorTableVaPtr
+userDataNode[1].offsetInDwords = 1
+userDataNode[1].sizeInDwords = 4
+userDataNode[1].set = 0
+userDataNode[1].next[0].type = DescriptorBuffer
+userDataNode[1].next[0].offsetInDwords = 3
+userDataNode[1].next[0].sizeInDwords = 8
+userDataNode[1].next[0].set = 0
+userDataNode[1].next[0].binding = 0
+
+trapPresent = 0
+debugMode = 0
+enablePerformanceData = 0
+vgprLimit = 0
+sgprLimit = 0
+maxThreadGroupsPerComputeUnit = 0
+
+[FsGlsl]
+#version 450
+#extension GL_ARB_separate_shader_objects : enable
+
+layout(binding = 0) uniform UniformBufferObject {
+    mat4 model;
+    mat4 view;
+    vec4 proj;
+} ubo;
+
+layout(location = 0) in vec3 fragColor;
+layout(location = 0) out vec4 outputColor;
+void main() {
+    outputColor = vec4(fragColor, 1.0) + ubo.proj;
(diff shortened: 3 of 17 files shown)


amdvlk-admin:

csyonghe:
This branch won't pass CI because it requires https://github.com/GPUOpen-Drivers/llvm-project/pull/1 for llvm-project.

llpc/context/llpcCompiler.cpp
```diff
@@ -865,9 +865,26 @@ Result Compiler::BuildPipelineWithRelocatableElf(
...
+#endif
+        }
+        else
+        {
+            auto pPipelineInfo = reinterpret_cast<const ComputePipelineBuildInfo*>(pContext->GetPipelineBuildInfo());
+            cacheHash = PipelineDumper::GenerateHashForComputePipeline(pPipelineInfo, true);
+#if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
+            pUserShaderCache = pPipelineInfo->pShaderCache;
+#endif
+        }
```
nhaehnle: Can you write this code without the preprocessor?

llpc/context/llpcCompiler.cpp
```diff
@@ -975,6 +992,42 @@ bool Compiler::CanUseRelocatableGraphicsShaderElf(
...
+        {
+            useRelocatableShaderElf = false;
+        }
+        else
+        {
+            ++relocatableElfCounter;
+        }
+    }
+    return useRelocatableShaderElf;
+}
```
nhaehnle: This isn't really a complaint to you specifically, but this is the kind of function that really ought to be much shorter:
```
bool Compiler::CanUseRelocatableComputeShaderElf(
    const PipelineShaderInfo* pShaderInfo   // Shader info for the pipeline to be built
    ) const
{
    if (!llvm::cl::UseRelocatableShaderElf)
        return false;

    if (pShaderInfo) {
        const ShaderModuleData* pModuleData = reinterpret_cast<const ShaderModuleData*>(pShaderInfo->pModuleData);
        if (pModuleData && pModuleData->binType != BinaryType::Spirv)
            return false;
    }

    if (cl::RelocatableShaderElfLimit != -1) {
        static uint32_t relocatableElfCounter = 0;
        if (relocatableElfCounter >= cl::RelocatableShaderElfLimit)
            return false;
        ++relocatableElfCounter;
    }
    return true;
}
```
I can live with the code style being bad for the time being, but please remove the `useRelocatableShaderElf` and use early-returns instead, as that makes the code much clearer to read.

llpc/context/llpcCompiler.cpp
```diff
@@ -1717,11 +1781,24 @@ Result Compiler::BuildComputePipeline(
...
     CacheEntryHandle hEntry = nullptr;
-    cacheEntryState = LookUpShaderCache(&cacheHash, &elfBin, &hEntry);
+    if (!buildingRelocatableElf)
+    {
+        cacheEntryState = LookUpShaderCache(&cacheHash, &elfBin, &hEntry);
+    }
+    else
+    {
+        cacheEntryState = ShaderEntryState::Compiling;
+    }
```
nhaehnle: Could you please first make a change that eliminates the preprocessor #if?

llpc/context/llpcCompiler.cpp
```diff
@@ -1675,6 +1737,8 @@ Result Compiler::BuildComputePipeline(
 {
     BinaryData elfBin = {};
 
+    bool buildingRelocatableElf = CanUseRelocatableComputeShaderElf(&pPipelineInfo->cs);
+
```
nhaehnle: Prefer to check this directly in the if-statement below, rather than adding a new local variable. (After you removed the duplication due to preprocessor #if).

llpc/context/llpcCompiler.cpp
```diff
@@ -1738,18 +1815,22 @@ Result Compiler::BuildComputePipeline(
...
+        if (!buildingRelocatableElf)
+        {
 #if LLPC_CLIENT_INTERFACE_MAJOR_VERSION < 38
-        UpdateShaderCaches((result == Result::Success), &elfBin, pShaderCache, hEntry, ShaderCacheCount);
+            UpdateShaderCaches((result == Result::Success), &elfBin, pShaderCache, hEntry, ShaderCacheCount);
 #else
-        UpdateShaderCache((result == Result::Success), &elfBin, hEntry);
+            UpdateShaderCache((result == Result::Success), &elfBin, hEntry);
 #endif
+        }
```
nhaehnle: Again, please eliminate the preprocessor #if-based code duplication first.

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -512,8 +528,23 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
+            {
+                pDescOffset = ConstantInt::get(Type::getInt32Ty(*m_pContext), descOffset);
+            }
+            else
+            {
+                IRBuilder<> builder(*m_pContext);
+                builder.SetInsertPoint(pInsertPoint);
+                SmallVector<Value*, 16> args;
+                args.push_back(ConstantInt::get(Type::getInt32Ty(*m_pContext), descSet));
+                args.push_back(ConstantInt::get(Type::getInt32Ty(*m_pContext), binding));
```
nhaehnle: builder.getInt32; also, prefer feeding the args directly to CreateIntrinsic instead of a SmallVector (using {arg0, arg1} syntax).

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -438,8 +440,22 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
```
nhaehnle: builder.getInt32, plus feeding the args directly to CreateIntrinsic, as in {arg0, arg1}.

llpc/context/llpcCompiler.cpp
csyonghe: This is a change that belong to #440 and #438. The preprocessor is pre-existing code related to the different handling of shader caches controlled by LLPC_CLIENT_INTERFACE_MAJOR_VERSION. Could you please advise how you want them eliminated?

nhaehnle: I ended up doing this myself. See #466

llpc/util/llpcElfWriter.cpp
```diff
@@ -1004,6 +1080,96 @@ void ElfWriter<Elf>::Reinitialize()
...
+        uint32_t offset = 0;
+
+        auto pPipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo*>(pContext->GetPipelineBuildInfo());
+
+        offset = GetResourceOffset(descSet, binding, pPipelineInfo->vs.pUserDataNodes, pPipelineInfo->vs
+        .userDataNodeCount);
+        if (offset == InvalidValue)
+        {
+            offset = GetResourceOffset(descSet, binding, pPipelineInfo->fs.pUserDataNodes, pPipelineInfo->fs
+                    .userDataNodeCount);
```
amingriyue: If I don't take mistake, this offset is a relevant offset in second level of resource nodding map. Do you know how we deal with root offset?

llpc/context/llpcCompiler.cpp
```diff
@@ -2172,6 +2179,8 @@ void Compiler::LinkRelocatableShaderElf(
                 return;
             }
         }
+        WriteElfFile("vs.elf", pShaderElfs[ShaderStageVertex]);
+        WriteElfFile("fs.elf", pShaderElfs[ShaderStageFragment]);
```
s-perron: We need to remember to make this conditional and more robust.  That is, what about compute pipelines?  No need to fix that for now.

llpc/context/llpcCompiler.cpp
```diff
@@ -2155,13 +2162,16 @@ void Compiler::LinkRelocatableShaderElf(
...
```
s-perron: This should be conditional and moved somewhere to be more generic.  Maybe a loop at the start of this function, or just before we call it, that will what will write our the shader for every stage that is not empty.  We can get suggestions on how to name thing, so we do not overwrite the files if we compile multiple pipeline.

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -509,8 +527,35 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
+                    ssRelocName << "_s";
+                }
+                else if (nodeType1 == ResourceMappingNodeType::DescriptorResource)
+                {
+                    ssRelocName << "_r";
+                }
+                else
+                {
+                    ssRelocName << "_b";
+                }
```
s-perron: We should start a markdown document that we can add to the LLPC repo that documents how the relocations work.  We can get feedback from AMD on how they want it documented.

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -509,8 +527,35 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
+                pDescOffset = builder.CreateIntrinsic(Intrinsic::amdgcn_reloc_constant, {}, {
+                    llvm::MetadataAsValue::get(*m_pContext, llvm::MDNode::get(*m_pContext,
+                    llvm::MDString::get(*m_pContext, ssRelocName.str().c_str())))
```
s-perron: I don't know what the style guide says.  This is hard to read because it has two lines forming one statement, but both are at the same level of indentation.

llpc/test/shaderdb/PipelineCs_RelocConst.pipe
```diff
@@ -0,0 +1,58 @@
+
```
s-perron: Some comments describing what you are trying to test will be useful.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
```diff
@@ -0,0 +1,103 @@
...
+; SHADERTEST-LABEL: 0000000000000000 _amdgpu_vs_main:
+; SHADERTEST: s_mov_b32 s[[RELOREG:[0-9]+]], 12 //{{.*}}
+; SHADERTEST: s_load_dwordx4 s[{{.*}}:{{.*}}], s[{{.*}}:{{.*}}], s[[RELOREG]] //{{.*}}
+; SHADERTEST: {{[0-9A-Za-z]+}} _amdgpu_ps_main:
+; END_SHADERTEST
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -v %gfxip %s | FileCheck -check-prefix=SHADERTEST1 %s
+; SHADERTEST1-LABEL: {{^// LLPC}} pipeline patching results
+; SHADERTEST1: %[[RELOCONST:[0-9]+]] = call i32 @llvm.amdgcn.reloc.constant(metadata !{{.*}})
```
s-perron: We should check that the name generated for the symbol is correct.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1193,128 @@ void ElfWriter<Elf>::Reinitialize()
...
+        if (resource->type > ResourceMappingNodeType::DescriptorBuffer)
+        {
+            continue;
+        }
+        if (resource->srdRange.set != descSet || resource->srdRange.binding != binding)
+        {
+            continue;
+        }
+
+        dbgs() << "Found user data node.\n";
```
s-perron: We can remove the debugging code.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
csyonghe: The name of the relocation symbol is not expected to be in the final linked elf so this check is not possible. However the test does attempt to check if the correct offset is written in the linked result.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/test/shaderdb/PipelineCs_RelocConst.pipe
csyonghe: Done.

llpc/patch/llpcPatchDescriptorLoad.cpp
csyonghe: Reformatted to the same line.

llpc/patch/llpcPatchDescriptorLoad.cpp
csyonghe: Ack.

llpc/context/llpcCompiler.cpp
csyonghe: Removed all these debugging code. We can add them back easily later if we need it.

csyonghe:
@nhaehnle @trenouf  This PR is now ready for review. These changes only implement the first step of userDataNode relocation: the offsets of descriptors in a descriptor table. It does not address other issues brought up by @trenouf  in #474. Those will come in future PRs.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
s-perron: At this point you are not checking the final linked elf.  It is looking at the llvm-ir after patching.  The llvm-ir should have the symbol name in the metadata that is a parameter to the builtin.  We want this part of the test to say that the patching passes did the right thing.

csyonghe: Done.

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -435,8 +438,19 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
+            if (!m_pPipelineState->GetBuilderContext()->BuildingRelocatableElf())
+            {
+                pDescOffset = ConstantInt::get(Type::getInt32Ty(*m_pContext), descOffset);
+            }
+            else
+            {
+                IRBuilder<> builder(*m_pContext);
+                builder.SetInsertPoint(pInsertPoint);
+                std::stringstream ssRelocName;
+                ssRelocName << "doff_" << descSet << "_" << binding << "_b";
```
nhaehnle: Prefer to use llvm::Twine.

llpc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -509,8 +523,32 @@ Value* PatchDescriptorLoad::LoadDescriptor(
...
+                    ssRelocName << "_s";
+                }
+                else if (nodeType1 == ResourceMappingNodeType::DescriptorResource)
+                {
+                    ssRelocName << "_r";
+                }
+                else
+                {
+                    ssRelocName << "_b";
+                }
```
nhaehnle: llvm::Twine

llpc/util/llpcElfReader.cpp
```diff
@@ -198,8 +198,8 @@ uint32_t ElfReader<Elf>::GetSymbolCount() const
 // Gets info of the symbol in the symbol table section according to the specified index.
 template<class Elf>
 void ElfReader<Elf>::GetSymbol(
-    uint32_t   idx,       // Symbol index
-    ElfSymbol* pSymbol)   // [out] Info of the symbol
+    uint32_t   idx,             // Symbol index
+    ElfSymbol* pSymbol) const   // [out] Info of the symbol
```
nhaehnle: Ugh, the style change to avoid those aligned comments can't come soon enough :/

llpc/util/llpcElfReader.cpp
```diff
@@ -321,7 +321,7 @@ void ElfReader<Elf>::GetSymbolsBySectionIndex(
                 secSymbols.push_back(symbol);
             }
         }
-
+
```
nhaehnle: Please configure your editor to automatically remove whitespace at the end of a line.

llpc/util/llpcElfWriter.cpp
```diff
@@ -836,6 +889,28 @@ void ElfWriter<Elf>::UpdateMetaNote(
...
+        const char*  pName,       // [in] Name of the section to look for
+        const void** pData,       // [out] Pointer to section data
+        size_t*      pDataLength  // [out] Size of the section data
+) const
+{
+    Result result = Result::ErrorInvalidValue;
+
+    auto pEntry = m_map.find(pName);
+
+    if (pEntry != m_map.end())
```
nhaehnle: Prefer to remove the line of whitespace here, due to the tight logical connection between the find and the if-block.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1193,142 @@ void ElfWriter<Elf>::Reinitialize()
     m_map[NoteName] = m_noteSecIdx;
 }
 
+uint32_t GetResourceOffset(uint32_t descSet, uint32_t binding, ResourceMappingNodeType type, const ResourceMappingNode* userResources, uint32_t
+    resourceCount);
```
nhaehnle: Should be static. Is there a strong reason for making this a forward declaration as opposed to just having the function definition here?

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1193,142 @@ void ElfWriter<Elf>::Reinitialize()
...
+
+struct RelocationEntry
+{
+    uint32_t offset;
+    const char* name;
+};
```
nhaehnle: Should be in an anonymous namespace. Also, this should contain the relocation type.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1167,21 +1378,40 @@ Result ElfWriter<Elf>::LinkGraphicsRelocatableElf(
 
     // Now get the symbols that belong in the symbol table.
     uint32_t offset = 0;
+    std::vector<RelocationEntry> relocations;
+
+    // Insert a dummy symbol. objdump somehow doesn't recognize the first entry in the symbol table section.
```
nhaehnle: Update the comment: This is a requirement of the ELF spec.

llpc/util/llpcInternal.cpp
```diff
@@ -84,6 +85,18 @@ CallInst* EmitCall(
     return builder.CreateNamedCall(funcName, pRetTy, args, attribs);
 }
 
+// =====================================================================================================================
+// Emits a amdgcn.reloc.constant intrinsics that represents a relocatable value with the given symbol name.
+llvm::CallInst* EmitRelocationConstant(
+    llvm::LLVMContext* context,                    // The LLVM context
+    llvm::Instruction* pInsertPos,                 // [in] Where to insert this call
```
nhaehnle: Please change the interface to take an llvm::IRBuilderBase instead. You also don't need to pass the LLVMContext explicitly.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
```diff
@@ -0,0 +1,105 @@
+
+// This test case checks that descriptor offset relocation works for buffer descriptors in a vs/fs pipeline.
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -o test_pipeline_out_relocconst.elf %gfxip %s && llvm-objdump -triple=amdgcn -mcpu=gfx900 -d test_pipeline_out_relocconst.elf | FileCheck -check-prefix=SHADERTEST %s && rm test_pipeline_out_relocconst.elf
+; SHADERTEST-LABEL: 0000000000000000 _amdgpu_vs_main:
+; SHADERTEST: s_mov_b32 s[[RELOREG:[0-9]+]], 12 //{{.*}}
+; SHADERTEST: s_load_dwordx4 s[{{.*}}:{{.*}}], s[{{.*}}:{{.*}}], s[[RELOREG]] //{{.*}}
+; SHADERTEST: {{[0-9A-Za-z]+}} _amdgpu_ps_main:
+; END_SHADERTEST
```
nhaehnle: Could you please augment the test to have relocations in both the VS and the PS, to make sure the relocation offsets are lined up correctly?

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1193,142 @@ void ElfWriter<Elf>::Reinitialize()
...
+{
+    for(uint32_t i = 0; i < relocations.size(); ++i)
+    {
+        auto reloc = relocations[i];
+        if (strncmp(reloc.name, "doff_", 5) == 0)
+        {
+            FixUpDescriptorOffsetRelocation(pWriter, pContext, reloc, isGraphicsPipeline);
+        }
+    }
+}
```
nhaehnle: This doesn't feel quite right for the design of a relocation kernel. Could you please refactor this so that:
- there is a function GetRelocationSymbolValue which returns the symbol value (e.g., descriptor offset for "doff_*") and a failure indication if the symbol is not found
- WriteRelocationOffset is effectively inlined here

And some defensive programming to check against the relocation type is necessary. Also, please take a look at the definition of relocation types in AMDGPUUsage.rst in LLVM -- the fixup of the relocation is currently incorrect because it doesn't consider the addend.

llpc/util/llpcElfReader.cpp
csyonghe: Done.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/util/llpcInternal.cpp
csyonghe: Done.

llpc/patch/llpcPatchDescriptorLoad.cpp
csyonghe: Done.

llpc/patch/llpcPatchDescriptorLoad.cpp
csyonghe: Done.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
csyonghe: Done.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/util/llpcElfWriter.cpp
csyonghe: Done.

llpc/test/shaderdb/PipelineVsFs_RelocConst.pipe
csyonghe: Done.

amdrexu:
Once the review issues are resolved, could you please click the button "Resolve conversation". Also, please merge the two commits, just update original commits to address review issues.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+        *value = GetResourceOffset(descSet, binding, type, pPipelineInfo->cs.pUserDataNodes, pPipelineInfo->cs.userDataNodeCount);
+    }
+
+    return *value != InvalidValue;
+}
+
+// =====================================================================================================================
+// Get the value of a relocation symbol. Returns true if success, and false if the symbol is unknown.
+bool GetRelocationSymbolValue(
+    Context *pContext,       // Pipeline compilation context
```
amdrexu: Add tag: // [in] ...

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    RelocationEntry reloc,   // The relocation entry
+    bool isGraphicsPipeline, // Whether we are processing a graphics pipeline
+    uint32_t* value)         // [out] the value of the relocation symbol
```
amdrexu: value -> pValue

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    size_t dataLength;
+    pWriter->GetSectionData(TextName, const_cast<const void**>(reinterpret_cast<void **>(&pData)), &dataLength);
+    assert(pData != nullptr && dataLength >= relocEntry.offset);
+    memcpy(pData + offset, &value, 4);
+}
+
+// =====================================================================================================================
+// Fixup a descriptor offset relocation (doff_x_y_t symbol) in the ELF with an actual offset value.
+bool GetDescriptorOffsetRelocationValue(
+    Context *pContext,               // Pipeline compilation context
```
amdrexu: Add tag: // [in] ...

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    RelocationEntry relocEntry,      // The relocation entry to fixup
+    bool isGraphicsPipeline,         // Whether we are processing a compute or graphics pipeline
+    uint32_t* value)                 // [out] The value of the relocation
```
amdrexu: value -> pValue

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+
+// =====================================================================================================================
+// Write the relocation value at the specificed location in the elf file.
+template<class Elf>
+void WriteRelocationConstant(
+    ElfWriter<Elf> *pWriter,    // [InOut] Writer to target ELF
+    uint32_t offset,            // The offset at which the value will be writtten
+    uint32_t value)             // The value of the relocation to write to the target ELF
+{
+    char* pData;
```
amdrexu: pData -> pData = nullptr

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    size_t dataLength;
```
amdrexu: dataLength -> dataLength = 0

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
     m_map[NoteName] = m_noteSecIdx;
 }
 
+// =====================================================================================================================
+// Retrieves the actual descriptor offset at the specified binding from UserDataNode.
+static uint32_t GetResourceOffset(
+    uint32_t descSet,                           // DescriptorSet index of the resource
+    uint32_t binding,                           // Binding slot of the resource
+    ResourceMappingNodeType type,               // Type of resource requested by the shader
```
amdrexu: type -> nodeType

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    const ResourceMappingNode* userResources,   // The UserDataNode provided at runtime
```
amdrexu: userResources -> pNodes // [in] ...

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    uint32_t resourceCount)                     // Number of resource nodes in UserDataNode.
```
amdrexu: resourceCount -> nodeCount

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+{
+    for(uint32_t  r = 0; r < resourceCount; ++r)
```
amdrexu: r -> i

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+    {
+        const ResourceMappingNode* resource = userResources+r;
```
amdrexu: const ResourceMappingNode* resource = userResources+r
->
const ResourceMappingNode* pNode = pNodes + i
or
const ResourceMappingNode* pNode = pNodes[i]

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
...
+
+        if (resource->type == ResourceMappingNodeType::DescriptorTableVaPtr) {
+            uint32_t offset = GetResourceOffset(descSet, binding, type, resource->tablePtr.pNext, resource->tablePtr.nodeCount);
```
amdrexu: This line exceeds 120-character

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,151 @@ void ElfWriter<Elf>::Reinitialize()
     m_map[NoteName] = m_noteSecIdx;
 }
 
+// =====================================================================================================================
+// Retrieves the actual descriptor offset at the specified binding from UserDataNode.
+static uint32_t GetResourceOffset(
```
amdrexu: GetResourceOffset -> GetDescriptorResourceOffset

csyonghe:
When I run make check-amdllpc locally, all tests seem to pass, but the testing process always freeze at 99% (seemed like a deadlock). Does anyone know why?

csyonghe:
OK, it seems like it is because I didn't update my llvm repo to latest commit and amdllpc is not getting built correctly.

csyonghe:
@nhaehnle  @amdrexu  I fixed the code according to your comments. Please let me know if there are any further issues. 

If https://reviews.llvm.org/D76440 looks good, can you also approve it?

amdrexu:
@csyonghe LGTM. Because Nicolai marked "request changes", once he re-approved it, this PR will be unblocked.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,155 @@ void ElfWriter<Elf>::Reinitialize()
...
+    return InvalidValue;
+}
+
+// =====================================================================================================================
+// Write the relocation value at the specificed location in the elf file.
+template<class Elf>
+void WriteRelocationConstant(
+    ElfWriter<Elf>* pWriter,    // [InOut] Writer to target ELF
+    uint32_t offset,            // The offset at which the value will be writtten
+    uint32_t value)             // The value of the relocation to write to the target ELF
```
nhaehnle: This function isn't used anymore?

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1206,155 @@ void ElfWriter<Elf>::Reinitialize()
...
+
+    for(uint32_t i = 0; i < relocations.size(); ++i)
+    {
+        auto reloc = relocations[i];
+        uint32_t relocationValue = 0;
+        if (GetRelocationSymbolValue(pContext, reloc, isGraphicsPipeline, &relocationValue))
+        {
+            assert(pData != nullptr && dataLength >= reloc.offset);
+            memcpy(pData + reloc.offset, &relocationValue, 4);
+        }
```
nhaehnle: Could you please make this assert when the symbol value is not found and when the relocation type is not as expected?

Also, this doesn't yet take the addend into account, does it?

csyonghe: Addends are being handled in `ElfReader<>::GetRelocation()`. Since currently all relocations are read from ".rel" section, there is no addends included. If in the future we are going to have "rela" section support, the addends can be handled there.

nhaehnle: This is not correct: When a .rel section is used, the addend is taken from the target section.

So for example, if the symbol value is 160 and the target dword in the target section is not 0 but instead e.g. 32, then the resulting value in the target section should be 192 for most relocation types.

csyonghe: Where is this "target dword" defined?

csyonghe: Here's the `Reloc` structure defined in `llpcElfReader.h`:
```
   // ELF relocation entry
    struct Reloc
    {
        uint64_t r_offset;     // Location (file byte offset, or program virtual address)
        union
        {
            uint64_t r_info;   // Symbol table index and type of relocation to apply
            struct
            {
                uint32_t  r_type;   // Type of relocation
                uint32_t  r_symbol; // Index of the symbol in the symbol table
            };
        };
    };
```

I cannot find anything related to "target" here, or in `SectionHeader` or in `Symbol`.

csyonghe: Do you refer to the dword value itself at the `offset` location? i.e. the true offset is
`actualOffset = *(uint32_t*)(pData+rel.offset) + rel.offset` ?

llpc/util/llpcElfWriter.cpp
```diff
@@ -1118,6 +1205,152 @@ void ElfWriter<Elf>::Reinitialize()
...
+
+// =====================================================================================================================
+// Get the value of a relocation symbol. Returns true if success, and false if the symbol is unknown.
+bool GetRelocationSymbolValue(
+    Context* pContext,       // [in] Pipeline compilation context
+    RelocationEntry reloc,   // The relocation entry
+    bool isGraphicsPipeline, // Whether we are processing a graphics pipeline
+    uint32_t* pValue)         // [out] the value of the relocation symbol
+{
+    assert(reloc.type == R_AMDGPU_ABS32 && "can only handle R_AMDGPU_ABS32 typed relocations.");
```
nhaehnle, line 1309: Please move this assertion into FixUpRelocations instead: GetRelocationSymbolValue is the place that knows how to get a symbol value, FixUpRelocations is the place that knows what to do with it.

JaxLinAMD:
there are new conflicts to resolve.

lgc/util/llpcInternal.cpp
```diff
@@ -81,6 +82,18 @@ CallInst* EmitCall(
...
 
+// =====================================================================================================================
+// Emits a amdgcn.reloc.constant intrinsics that represents a relocatable value with the given symbol name.
+llvm::CallInst* EmitRelocationConstant(
+    llvm::IRBuilderBase* pBuilder,             // [in,out] An IRBuilder for instruction insertion
+    const llvm::Twine& symbolName)             // Name of the relocation symbol associated with this relocation
+{
+    auto pMDNode = llvm::MDNode::get(pBuilder->getContext(), llvm::MDString::get(pBuilder->getContext(),
+                                     symbolName.str()));
+    return pBuilder->CreateIntrinsic(Intrinsic::amdgcn_reloc_constant, {},
```
trenouf, line 93: Does it work to have an external global variable (perhaps in the const32 address space so it has a 32-bit address) and use the address of that as the constant? Then we wouldn't need the special amdgcn_reloc_constant intrinsic.

csyonghe:
Can anyone start the Jenkins test?

trenouf:
I have just started one. Not sure how it will fare; there have been some problems with the Jenkins test over the last few days.

JaxLinAMD:
retest this please

JaxLinAMD:
retest this please

JaxLinAMD:
retest this please

JaxLinAMD:
@csyonghe  all agents are hanged because of this pull request, please make sure hang is not observed on your local environment when do next force-push.

csyonghe:
> @csyonghe all agents are hanged because of this pull request, please make sure hang is not observed on your local environment when do next force-push.

How do I setup my local environment to run the tests?

JaxLinAMD:
@csyonghe , do you mean how to fetch cts source code and compile cts debug version driver?

csyonghe:
I am not sure how to produce the exact same environment as the testing server, including buildling the  driver and run it. I can try build the Stadia version and run it on stadia server, but that might differ from the server. I am just wondering if there are some instructions I can follow to setup the testing environment.

JaxLinAMD:
I think you can try on stadia environment if it's more convenient for you. if you want try on local enviroment, you need to setup amdgpu-pro driver and cts binary.

JaxLinAMD:
do you need instructions on building open-source driver or building cts binary?

csyonghe:
I can build the driver. I also have the instructions on building the cts binary. I am just not sure how to deploy and test the driver on a local environment.

JaxLinAMD:
you need to install a opensource driver from repos in README, and replace /usr/lib/x86_64-linux-gnu/amdvlk64.so on ubuntu with drivers you built.


sudo wget -qO - http://repo.radeon.com/amdvlk/apt/debian/amdvlk.gpg.key | sudo apt-key add -
sudo sh -c 'echo deb [arch=amd64] http://repo.radeon.com/amdvlk/apt/debian/ bionic main > /etc/apt/sources.list.d/amdvlk.list'
sudo apt-get remove amdvlk /* If old version is installed on the machine, remove it first */
sudo apt update
sudo apt-get install amdvlk

csyonghe:
I see. Thanks!

amdrexu:
@csyonghe I still see some coding issues with new coding style. Please have a check.
Such as:
`pMDNode -> metaNode` (we don't use p prefix any more and names are camelcase, avoid acronym)
`A* -> A *` (this could be done by clang-format, positions of * and & follow LLVM style now)

llpc/util/llpcElfWriter.cpp
nhaehnle: This dropped off my radar, but yes, that's what I meant.

lgc/patch/llpcPatchDescriptorLoad.cpp
```diff
@@ -297,9 +320,12 @@ Value *PatchDescriptorLoad::getDescPtr(ResourceNodeType resType, unsigned descSe
       offsetInDwords += DescriptorSizeResource / 4;
     offset = builder.getInt32(offsetInDwords);
   }
-  descPtr = builder.CreateBitCast(descPtr, builder.getInt32Ty()->getPointerTo(ADDR_SPACE_CONST));
-  descPtr = builder.CreateGEP(builder.getInt32Ty(), descPtr, offset);
-
+  // The LLVM's internal handling of GEP instruction results in a lot of junk code and prevented selection
```
trenouf, line 323: Re Jax's report that this change causes test agents to hang: I'm a bit suspicious of this edit here, if only because nothing else looks like it should affect normal pipeline compilation. You could try making it conditional on !isa<ConstantInt>(descPtr), falling back to the old GEP code when it is a ConstantInt, which it will always be for pipeline compilation.

csyonghe: You are right. This is actually wrong when relocation is not used. In normal mode, offset is in dwords not bytes, so it cannot be just an add. I have corrected this issue and trying to test this before making a push. Thanks so much for pointing this out.

lgc/util/llpcInternal.cpp
trenouf: To answer my own question:

No that does not work. It tries to get an address to the external global variable in a PC-relative way, which is not what we want.

lgc/patch/llpcPatchDescriptorLoad.cpp
trenouf: Ah right, that would explain the crash. :-)

I don't like the offset being in dwords for one path and bytes for the other path. I'll have a look at changing the normal path to bytes in a separate change.

csyonghe:
The Jenkins test still haven't returned any result. Can we restart it?

JacobHeAMD:
There is some problem with Jenkins tests. I've already asked infrastructure team to have a look.

JacobHeAMD:
retest this please

csyonghe:
Can you post the details on which case is failing?

trenouf:
It looks to me like it passed CTS on two agents and hung waiting for the third, so I would take that as a pass.

JaxLinAMD:
let me re-trigger the test again.

JaxLinAMD:
retest this please

JaxLinAMD:
merge process is started, please do not force push.

Issue #431, opened 2020-01-15
Cache creator tool


s-perron:
# Problem description

Compiling shaders when a game is running can be time consuming, and lowers the quality of the game.  This can be because of long wait times when a level starts or, much worse, hitching during game play.  One of the solutions to this is to create a cache of precompiled pipelines or shaders.  Then much less work needs to be done while the game is running.

This leads to a second problem.  How do we create the cache?  For some games, building the cache as people play the game does not work.  Developers are not able to see how their game will perform as they are developing it.  This would be a time consuming process that does not scale well, especially when the developers have a large number of shaders that are only run under very specific situations.

In this document, we propose a tool which we will call the “LLPC cache creator”.  This will be a tool that will create a shader cache that can be used by LLPC without running the application.  The input to the tool will be the Spir-V for the shaders and a few other options like the architecture.  It will write a cache to disk that could be passed to LLPC.

# Developer workflow

Developers write the shaders for their game.  Those will be given to the cache creator to produce the cache for LLPC with appropriate inputs.  Once the cache is available this is a file that could be loaded and passed to `vkCreatePipelineCache`.

The developer will not have to run their game to prime the cache, but this is accomplished offline.

# Requirements

*   The shader cache that is built does not require running the game.
*   The input is
    *   Spir-v binaries
    *   The gpu architecture to target
    *   Possible values for spec constants
    *   Possible formats for images, samplers, frame buffers, etc.
*   The output is a shader cache that can be used by LLPC
*   Initially, the shader cache will have a 100% hit rate for
    *   Compute pipelines meeting the constraints of the inputs
    *   Graphics pipelines composed of just a vertex shader and a fragment shader meeting the constraints of the inputs.
*   On average, the runtime of the generated code when using the cache must not be more than 5% worse than the code generated by doing a full compile.
*   On average, the cache hit time must be no more than 20% of the cache hit time of the current implementation.
*   The size of the shader cache cannot be larger than the size of the cache generated by the current implementation.
*   The tool must run in a generic Linux, even if an AMD GPU is not available.

# Relocatable shader elf

The cache creator tool has to build a cache that cover cases that may never come up.  This increases the size of the cache.  We need ways of implementing the cache that will reduce the size in order to meet the requirement that the cache will not increase in size.  This will be accomplished by compiling CS, VS, and PS shaders into individual **relocatable shader elf files** that can be reused in multiple pipelines.  These elf files will be stored in the cache in place of the current full pipeline elf files.

[...]

PR #471, opened 2020-02-18
Implement YCbCr Conversion Algorithm

jiaxie-amd:
- Add the calculation method of image descripter for each plane
 - Add the YCbCr conversion algorithm

### DIFF
--- a/llpc/builder/llpcBuilderImpl.h
+++ b/llpc/builder/llpcBuilderImpl.h
@@ -436,6 +436,169 @@ private:
     LLPC_DISALLOW_DEFAULT_CTOR(BuilderImplImage)
     LLPC_DISALLOW_COPY_AND_ASSIGN(BuilderImplImage)
 
+    // Represents the type of sampler filter.
+    enum class SamplerFilter : uint32_t
+    {
+        Nearest = 0,
+        Linear,
+    };
+
+    // Represents the YCbCr conversion model.
+    enum class SamplerYCbCrModelConversion : uint32_t
+    {
+        RgbIdentity = 0,
+        YCbCrIdentity,
+        YCbCr709,
+        YCbCr601,
+        YCbCr2020,
+    };
+
+    // Represents whether color channels are encoded using the full range of numerical
+    // values or whether values are reserved for headroom and foot room.
+    enum class SamplerYCbCrRange : uint32_t
+    {
+        ItuFull = 0,
+        ItuNarrow,
+    };
+
+    // Represents the location of downsampled chroma channel samples relative to the luma samples.
+    enum class ChromaLocation : uint32_t
+    {
+        CositedEven = 0,
+        Midpoint,
+    };
+
+    // Represents the component values placed in each component of the output vector.
+    class ComponentSwizzle
+    {
+    public:
+        enum Value: uint32_t
+        {
+            Identity = 0,
+            Zero,
+            One,
+            R,
+            G,
+            B,
+            A,
+        };
+
+        ComponentSwizzle() = default;
+        ComponentSwizzle(const Value& swizzle) : value(swizzle) {};
+        ComponentSwizzle(const uint32_t swizzle) { value = static_cast<Value>(swizzle); }
+
+        operator Value() const { return value; }
+        ComponentSwizzle& operator=(uint32_t op) { value = static_cast<Value>(op); return *this; }
+        constexpr uint32_t operator-(ComponentSwizzle::Value op) const { return value - op; }
+        constexpr bool operator!=(ComponentSwizzle::Value op) const { return value != op; }
+        constexpr bool operator==(ComponentSwizzle::Value op) const { return value == op; }
+
+    private:
+        Value value;
+    };
+
+    // General YCbCr sample info
+    struct YCbCrSampleInfo
+    {
+        Type*              pResultTy;
+        uint32_t           dim;
+        uint32_t           flags;
+        Value*             pImageDesc;
+        Value*             pSamplerDesc;
+        ArrayRef<Value*>   address;
+        const Twine&       instName;
+        bool               isSample;
+    };
+
+    // YCbCr sample info for downsampled chroma channels in the X dimension
+    struct XChromaSampleInfo
+    {
+        YCbCrSampleInfo*   pYCbCrInfo;
+        Value*             pImageDesc1;
+        Value*             pI;
+        Value*             pJ;
+        Value*             pChromaWidth;
+        Value*             pChromaHeight;
+        ChromaLocation     xChromaOffset;
+    };
+
+    // YCbCr sample info for downsampled chroma channels in both X and Y dimension
+    struct XYChromaSampleInfo : XChromaSampleInfo
+    {
+        Value*             pImageDesc2;
+        uint32_t           planeNum;
+        ChromaLocation     yChromaOffset;
+    };
+
+    // YCbCr wrapped sample info
+    struct YCbCrWrappedSampleInfo : XYChromaSampleInfo
+    {
+        Value*             pImageDesc3;
+        bool               subsampledX;
+        bool               subsampledY;
+    };
+
+    // Replace [beginBit, beginBit + adjustBits) bits with pData in specific pWord
+    Value* ReplaceBitsInWord(Value*   pWord,
+                             uint32_t beginBit,
+                             uint32_t adjustBits,
+                             Value*   pData);
+
+    // Implement transfer from ST coordinates to UV coordiantes operation
+    Value* TransferSTtoUVCoords(Value* pST,
+                                Value* pSize);
+
+    // Implement the adjustment of UV coordiantes when the sample location associated with
+    // downsampled chroma channels in the X/XY dimension occurs
+    Value* YCbCrCalculateImplicitChromaUV(ChromaLocation offset,
+                                          Value*         pUV);
+
+    // Transfer IJ coordinates from UV coordinates
+    Value* TransferUVtoIJCoords(SamplerFilter filter,
+                                Value*        pUV);
+
+    // Calculate UV offset to top-left pixel
+    Value* CalculateUVoffset(Value* pUV);
+
+    // Implement bilinear blending
+    Value* BilinearBlend(Value* pAlpha,
+                         Value* pBeta,
+                         Value* pTL,
+                         Value* pTR,
+                         Value* pBL,
+                         Value* pBR);
+
+    // Implement wrapped YCbCr sample
+    Value* YCbCrWrappedSample(YCbCrWrappedSampleInfo& wrapInfo);
+
+    // Implement reconstructed YCbCr sample operation for downsampled chroma channels in the X dimension
+    Value* YCbCrReconstructLinearXChromaSample(XChromaSampleInfo& xChromaInfo);
+
+    // Implement reconstructed YCbCr sample operation for downsampled chroma channels in both X and Y dimension
+    Value* YCbCrReconstructLinearXYChromaSample(XYChromaSampleInfo& xyChromaInfo);
+
+    // Implement interanl image sample for YCbCr conversion
+    Value* YCbCrCreateImageSampleInternal(SmallVectorImpl<Value*>& coords,
+                                          YCbCrSampleInfo*         pYCbCrInfo);
+
+    // Generate sampler descriptor for YCbCr conversion
+    Value* YCbCrGenerateSamplerDesc(Value*        pSamplerDesc,
+                                    SamplerFilter filter,
+                                    bool          forceExplicitReconstruction);
+
+    // Implement range expanding operation on checking whether the encoding uses full numerical range on chroma channel
+    Value* YCbCrRangeExpand(SamplerYCbCrRange  range,
+                            const uint32_t*    pBits,
+                            Value*             pSample);
+
+    // Implement the color transfer operation for conversion from YCbCr to RGB color model
+    Value* YCbCrConvertColor(Type*                       pResultTy,
+                             SamplerYCbCrModelConversion colorModel,
+                             SamplerYCbCrRange           range,
+                             uint32_t*                   pBits,
+                             Value*                      pImageOp);
+
+
     // Implement pre-GFX9 integer gather workaround to patch descriptor or coordinate before the gather
     Value* PreprocessIntegerImageGather(uint32_t dim, Value*& pImageDesc, Value*& pCoord);
 
@@ -456,6 +619,17 @@ private:
                                    ArrayRef<Value*> address,
                                    const Twine&     instName,
                                    bool             isSample);
+
+    // Common code to create an image YCbCr sample.
+    Value* CreateImageYCbCrSample(Type*            pResultTy,
+                                  uint32_t         dim,
+                                  uint32_t         flags,
+                                  Value*           pCoord,
+                                  Value*           pImageDesc,
+                                  Value*           pSamplerDesc,
+                                  ArrayRef<Value*> address,
+                                  const Twine&     instName,
+                                  bool             isSample);
 
     // Common code for CreateImageAtomic and CreateImageAtomicCompareSwap
     Value* CreateImageAtomicCommon(uint32_t          atomicOp,
--- a/llpc/builder/llpcBuilderImplImage.cpp
+++ b/llpc/builder/llpcBuilderImplImage.cpp
@@ -1221,6 +1221,1185 @@ Value* BuilderImplImage::CreateImageStore(
 }
 
 // =====================================================================================================================
+// Implement wrapped YCbCr sample
+Value* BuilderImplImage::YCbCrWrappedSample(
+    YCbCrWrappedSampleInfo& wrapInfo) // [In] Wrapped YCbCr sample infomation
+{
+    SmallVector<Value*, 4> coordsChroma;
+    YCbCrSampleInfo* pSampleInfo = wrapInfo.pYCbCrInfo;
+    Value* pChromaWidth  = wrapInfo.pChromaWidth;
+    Value* pChromaHeight = wrapInfo.pChromaHeight;
+
+    if (wrapInfo.subsampledX)
+    {
+        pChromaWidth = CreateFMul(wrapInfo.pChromaWidth, ConstantFP::get(getFloatTy(), 0.5f));
+    }
+
+    if (wrapInfo.subsampledY)
+    {
+        pChromaHeight = CreateFMul(wrapInfo.pChromaHeight, ConstantFP::get(getFloatTy(), 0.5f));
+    }
+
+    coordsChroma.push_back(CreateFDiv(wrapInfo.pI, pChromaWidth));
+    coordsChroma.push_back(CreateFDiv(wrapInfo.pJ, pChromaHeight));
+
+    pSampleInfo->pImageDesc = wrapInfo.pImageDesc1;
+
+    Value* pResult = nullptr;
+
+    if (wrapInfo.planeNum == 1)
+    {
+        pSampleInfo->pImageDesc = wrapInfo.subsampledX ? wrapInfo.pImageDesc2 : wrapInfo.pImageDesc1;
+
+        Instruction* pImageOp = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChroma, pSampleInfo));
+        pResult = CreateShuffleVector(pImageOp, pImageOp, { 0, 2 });
+    }
+    else if (wrapInfo.planeNum == 2)
+    {
+        pSampleInfo->pImageDesc = wrapInfo.pImageDesc2;
+        Instruction* pImageOp = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChroma, pSampleInfo));
+        pResult = CreateShuffleVector(pImageOp, pImageOp, { 0, 2 });
+    }
+    else if (wrapInfo.planeNum == 3)
+    {
+        pSampleInfo->pImageDesc = wrapInfo.pImageDesc2;
+        Instruction* pImageOp1 = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChroma, pSampleInfo));
+
+        pSampleInfo->pImageDesc = wrapInfo.pImageDesc3;
+        Instruction* pImageOp2 = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChroma, pSampleInfo));
+        pResult = CreateShuffleVector(pImageOp2, pImageOp1, { 0, 6 });
+    }
+    else
+    {
+        LLPC_NEVER_CALLED();
+    }
+
+    return pResult;
+}
+
+// =====================================================================================================================
+// Create YCbCr Reconstruct linear X chroma sample
+Value* BuilderImplImage::YCbCrReconstructLinearXChromaSample(
+    XChromaSampleInfo& xChromaInfo) // [In] Infomation for downsampled chroma channels in X dimension
+{
+    YCbCrSampleInfo* pSampleInfo = xChromaInfo.pYCbCrInfo;
+    Value* pIsEvenI = CreateICmpEQ(CreateSMod(CreateFPToSI(xChromaInfo.pI, getInt32Ty()), getInt32(2)), getInt32(0));
+
+    Value* pSubI = CreateUnaryIntrinsic(Intrinsic::floor,
+                                        CreateFDiv(xChromaInfo.pI, ConstantFP::get(getFloatTy(), 2.0)));
+    if (xChromaInfo.xChromaOffset != ChromaLocation::CositedEven)
+    {
+        pSubI = CreateSelect(pIsEvenI,
+                            CreateFSub(pSubI, ConstantFP::get(getFloatTy(), 1.0)),
+                            pSubI);
+    }
+
+    Value* pAlpha = nullptr;
+    if (xChromaInfo.xChromaOffset == ChromaLocation::CositedEven)
+    {
+        pAlpha = CreateSelect(pIsEvenI, ConstantFP::get(getFloatTy(), 0.0), ConstantFP::get(getFloatTy(), 0.5));
+    }
+    else
+    {
+        pAlpha = CreateSelect(pIsEvenI, ConstantFP::get(getFloatTy(), 0.25), ConstantFP::get(getFloatTy(), 0.75));
+    }
+
+    Value* pT = CreateFDiv(xChromaInfo.pJ, xChromaInfo.pChromaHeight);
+
+    SmallVector<Value*, 4> coordsChromaA;
+    pSampleInfo->pImageDesc = xChromaInfo.pImageDesc1;
+    coordsChromaA.push_back(CreateFDiv(pSubI, xChromaInfo.pChromaWidth));
+    coordsChromaA.push_back(pT);
+    Instruction* pImageOpA = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChromaA, pSampleInfo));
+
+    SmallVector<Value*, 4> coordsChromaB;
+    coordsChromaB.push_back(CreateFDiv(CreateFAdd(pSubI, ConstantFP::get(getFloatTy(), 1.0)), xChromaInfo.pChromaWidth));
+    coordsChromaB.push_back(pT);
+    Instruction* pImageOpB = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChromaB, pSampleInfo));
+
+    Value* pResult = CreateFMix(pImageOpB, pImageOpA, pAlpha);
+
+    return CreateShuffleVector(pResult, pResult, { 0, 2 });
+}
+
+// =====================================================================================================================
+// Create YCbCr Reconstruct linear XY chroma sample
+Value* BuilderImplImage::YCbCrReconstructLinearXYChromaSample(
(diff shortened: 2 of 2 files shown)


amdvlk-admin:
### Test summary for commit cb22396dd019d4b674beb177557721fa45de5c45

<li>XGL: dev 052ebfa2e77abd14fdb4f5120b35b081f3fc4bdc</li>
</ul></details>

<li>Passed: 942/942 (100.0%)</li>
<li>Failed: 0/942 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

<li>Passed: 65431/65431 (100.0%)</li>
<li>Failed: 0/65431 (0.0%)</li>
</ul></details>


</ul></details>

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
     LLPC_DISALLOW_DEFAULT_CTOR(BuilderImplImage)
     LLPC_DISALLOW_COPY_AND_ASSIGN(BuilderImplImage)
 
+    // Represents the type of sampler filter.
+    enum class SamplerFilter : uint32_t
```
kuhar: Why do we need 32 bits for 2 possible values? Will this enum or the enums below be serialized?

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+            One,
+            R,
+            G,
+            B,
+            A,
+        };
+
+        ComponentSwizzle() = default;
+        ComponentSwizzle(const Value& swizzle) : value(swizzle) {};
+        ComponentSwizzle(const uint32_t swizzle) { value = static_cast<Value>(swizzle); }
```
kuhar: `swizzle` may be an arbitrary int. Can we check that it is at least within the range of `Value`?

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+
+        operator Value() const { return value; }
+        ComponentSwizzle& operator=(uint32_t op) { value = static_cast<Value>(op); return *this; }
```
kuhar: Same.

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+        constexpr uint32_t operator-(ComponentSwizzle::Value op) const { return value - op; }
```
kuhar: What is the semantic meaning of subtraction here? For example, can I subtract `Value::R` from `Value::Identity` and what does the result mean?

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+    // General YCbCr sample info
+    struct YCbCrSampleInfo
+    {
+        Type*              pResultTy;
+        uint32_t           dim;
+        uint32_t           flags;
+        Value*             pImageDesc;
+        Value*             pSamplerDesc;
+        ArrayRef<Value*>   address;
+        const Twine&       instName;
```
kuhar: Is it possible to use an `llvm::Twine` as a member here? I am mostly worried about dangling pointers here with `YCbrCrSampleInfo` that outlive temporaries.
See http://llvm.org/docs/ProgrammersManual.html#the-twine-class and https://llvm.org/doxygen/classllvm_1_1Twine.html#details.

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+                                   pYCbCrInfo->pImageDesc,
+                                   pYCbCrInfo->pSamplerDesc,
+                                   pYCbCrInfo->address,
+                                   pYCbCrInfo->instName,
+                                   pYCbCrInfo->isSample);
+}
+
+// =====================================================================================================================
+// Replace [beginBit, beginBit + adjustBits) bits with data in specific word
+Value* BuilderImplImage::ReplaceBitsInWord(
```
kuhar: Isn't this simply bit-addressible memcpy?

kuhar: IE it doesn't really replace anything but builds a new value

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    Value*      pWord,      // [in] Target wowrd
+    uint32_t    beginBit,   // The first bit to be replaced
+    uint32_t    adjustBits, // The number of bits should be replaced
```
kuhar: Consider rewording this to better reflect that it's just a copy operations. For example, `s/adjustBits/numBitsToCopy` or `numBits` or `bitLen`, etc.

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    Value*      pData)      // [in] The data used to replace specific bits
+{
+    uint32_t mask = ((1 << adjustBits) - 1) << beginBit;
```
kuhar: `const uint32_t`

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+        if (auto pDataVecTy = dyn_cast<VectorType>(pData->getType()))
+        {
+            pWord = CreateVectorSplat(pDataVecTy->getNumElements(), pWord);
+            pMask = CreateVectorSplat(pDataVecTy->getNumElements(), pMask);
+            pInvMask = CreateVectorSplat(pDataVecTy->getNumElements(), pInvMask);
+            pBeginBit = CreateVectorSplat(pDataVecTy->getNumElements(), pBeginBit);
+        }
+    }
+
+    Value* pNewWord = CreateAnd(pWord, pInvMask); // (pWord & ~mask)
```
kuhar: This comment is excellent. Can we have something like for the previous function as well?

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+                                   pYCbCrInfo->address,
+                                   pYCbCrInfo->instName,
+                                   pYCbCrInfo->isSample);
+}
+
+// =====================================================================================================================
+// Replace [beginBit, beginBit + adjustBits) bits with data in specific word
+Value* BuilderImplImage::ReplaceBitsInWord(
+    Value*      pWord,      // [in] Target wowrd
+    uint32_t    beginBit,   // The first bit to be replaced
```
kuhar: Can you clarify whether this is lowest or highest bits?

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+            pWord = CreateVectorSplat(pDataVecTy->getNumElements(), pWord);
+            pMask = CreateVectorSplat(pDataVecTy->getNumElements(), pMask);
+            pInvMask = CreateVectorSplat(pDataVecTy->getNumElements(), pInvMask);
+            pBeginBit = CreateVectorSplat(pDataVecTy->getNumElements(), pBeginBit);
+        }
+    }
+
+    Value* pNewWord = CreateAnd(pWord, pInvMask); // (pWord & ~mask)
+    pData = CreateAnd(CreateShl(pData, pBeginBit), pMask); // ((pData << beginBit) & mask)
+    pNewWord = CreateOr(pNewWord, pData);
```
kuhar: `return CreateOr(...);`

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    Value*                      pSamplerDesc,   // [in] Sampler descriptor
+    ArrayRef<Value*>            address,        // Address and other arguments
+    const Twine&                instName,       // [in] Name to give instruction(s)
+    bool                        isSample)       // Is sample rather than gather
+{
+    Value* pResult = nullptr;
+
+    // Extract YCbCr meta data
+    SamplerYCbCrConversionMetaData yCbCrMetaData = {};
+    yCbCrMetaData.word0.u32All = dyn_cast<ConstantInt>(CreateExtractElement(pSamplerDesc, getInt64(4)))->getZExtValue();
```
kuhar: Can dyn_cast return `nullptr` here?  What happens then?

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+        pImageDesc2 = UndefValue::get(pImageDesc->getType());
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord0Cr, getInt64(0));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord1,    getInt64(1));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord2Cr, getInt64(2));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord3Cr, getInt64(3));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord4Cr, getInt64(4));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord5,    getInt64(5));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord6,    getInt64(6));
+        pImageDesc2 = CreateInsertElement(pImageDesc2, pWord7,    getInt64(7));
+    }
```
nhaehnle: Terrible code duplication.

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    const SamplerYCbCrRange yCbCrRange           = static_cast<SamplerYCbCrRange>(yCbCrMetaData.word0.yCbCrRange);
+    const SamplerYCbCrModelConversion yCbCrModel = static_cast<SamplerYCbCrModelConversion>(yCbCrMetaData.word0.yCbCrModel);
+
+    // Init sample descriptor for luma channel
+    Value* pSamplerDescLuma   = pSamplerDesc;
+
+    // Init sample descriptor for chroma channels
+    Value* pSamplerDescChroma = YCbCrGenerateSamplerDesc(pSamplerDesc, chromaFilter, forceExplicitReconstruct);
+
+    // Extract SQ_IMG_RSRC_WORD
```
nhaehnle: This code and most of what follows is 100% unmaintainable and does not take into account the differences in image descriptor definitions between different hardware generations.

Please clean this up by:
* Establish a helper method that decomposes an image descriptor into its relevant fields.
* Establish a helper method that create an image descriptor (T#) from scratch, given relevant input data.
* Rewrite this method to first decompose the image descriptor using the helper, then recreating new ones as required.

When writing the helper methods, be sure to take into account the changes in descriptor format across hardware generations.

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    return pResult;
+}
+
+// =====================================================================================================================
+// Create YCbCr image sample internal
+Value* BuilderImplImage::YCbCrCreateImageSampleInternal(
+    SmallVectorImpl<Value*>& coords,     // ST coords
+    YCbCrSampleInfo*         pYCbCrInfo) // [In] YCbCr smaple information
+{
+    Value* pCoords = CreateInsertElement(Constant::getNullValue(VectorType::get(coords[0]->getType(), 2)),
```
nhaehnle: Use an UndefValue instead of a null value.

llpc/builder/llpcBuilderImpl.h
nhaehnle: I would propose a `getChannel` method instead, which asserts that we're one of R G B A, and returns 0 1 2 3 accordingly.

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+        Midpoint,
+    };
+
+    // Represents the component values placed in each component of the output vector.
+    class ComponentSwizzle
+    {
+    public:
+        enum Value: uint32_t
+        {
+            Identity = 0,
```
nhaehnle: Please remove the Identity swizzle enum from here. The corresponding hardware bits don't have an Identity either.

To be specific: please set `Zero = 0, One = 1, R = 4, G = 5, B = 6, A = 7`

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+    xyChromaInfo.xChromaOffset = xChromaOffset;
+    xyChromaInfo.yChromaOffset = yChromaOffset;
+
+    // Init wrapped smaple chroma info
+    YCbCrWrappedSampleInfo wrappedSampleInfo = {};
+    wrappedSampleInfo.pYCbCrInfo = &sampleInfo;
+    wrappedSampleInfo.pChromaWidth = pWidth;
+    wrappedSampleInfo.pChromaHeight = pHeight;
+    wrappedSampleInfo.pI = pU;
+    wrappedSampleInfo.pJ = pV;
```
nhaehnle: This seems like a mistake...

llpc/builder/llpcBuilderImpl.h
jiaxie-amd: Ok, I'll use bool instead for 2 possible value cases.

llpc/builder/llpcBuilderImpl.h
```diff
@@ -438,6 +438,169 @@ class BuilderImplImage : virtual public BuilderImplBase
...
+            Zero,
+            One,
+            R,
+            G,
+            B,
+            A,
+        };
+
+        ComponentSwizzle() = default;
+        ComponentSwizzle(const Value& swizzle) : value(swizzle) {};
```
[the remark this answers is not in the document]
jiaxie-amd: Revised, thanks.

llpc/builder/llpcBuilderImpl.h
jiaxie-amd: Yes, this overload operator will be removed and getChannel() is used instead, which would be meaningful.

llpc/builder/llpcBuilderImpl.h
kuhar: Enum is fine.
I was trying to point out that you force all the enums below to be 32-bits wide without explaining why you need it. By default, I'd expect you to leave out the type part and let a compiler pick something. Forcing a particular type would be understandable if you had to serialize these enums or perform frequent int-to-enum/enum-to-int conversions, but in that case it may make more sense to use `std::underlying_type<SomeEnum>::type` to get the right integer type instead.

kuhar:
@jiaxie-amd after thinking about this a bit more, I think it's very easy to get caught up with minor coding issues here and miss the bigger picture.

Would it be possible for you to split this PR into a couple of smaller one, each with its own test? I think that would make it much easier to review.
For example, the bit manipulation part and generating llvm expressions could be factored out into a file with YCbCr conversion utilities and it's own small set of unit/LIT tests.

jiaxie-amd:
> @jiaxie-amd after thinking about this a bit more, I think it's very easy to get caught up with minor coding issues here and miss the bigger picture.
> 
> Would it be possible for you to split this PR into a couple of smaller ones, each with its own tests? I think that would make it much easier to review.
> For example, the bit manipulation part and generating llvm expressions could be factored out into a file with YCbCr conversion utilities and it's own small set of unit/LIT tests.

@kuhar 
Thanks for helping review this patch.
Regarding to your concerns, Yes, the patch has many utility functions to serve CreateImageYCbCrSample, mainly focus on two big issue:
1. How to use YCbCrMetaData(passed from pipeline createinfo) to adjust image srd (runtime data in gpu ram) for sampling each YCbCr image plane's data correctly.
2. How to use YCbCrMetaData to transform the sampled data from YCbCr color space to RGBA correctly.

So, as you can see, for handling YCbCrMetaData, I wrote some utility functions to handle Word bits which is gotten from Vulkan. And most of the rest functions are used for generating transform alogrithm.

As you've mentioned above, I'm not quite sure whether the newly added bits manipulating functions are sufficient(good) enough or not (so, thanks for your suggestions). But i'm sure they're correct, for that this patch has been verified by vk-cts with thousands of test cases, no test fails,
 please refer to: dEQP-VK.ycbcr.conversion.* if you're interested.

But as @nhaehnle pointed out, I'm more concerned about how to easily maintain these code, for that the above code only works for specifc Graphics card(Gfx9), due to the bits in img srd is different among Gfxips. Which means this yCbCr smapling path needs update when new Gfxip added in.  I have the patch ready for Gfx10, but not uploaded, but what about latter? So, I want to solve some trival issues as you guys raised and then focus on giving a good solution for easily maintain these code with latest gfxip. 

It would be glad if this patch could be merged quickly after revising some main problems. And then i'll take a longer time to refine the code (such as auto generate some gfxip related code and split ycbcr related module apart) to make it easy maintainable and understandable.

jiaxie-amd:
> Thank you for this. Unfortunately, I have lots of comments too. In order for things not to get lost in the noise, the _really_ important bits are:
> 
> * Re-think how you signal whether an image sample needs to be a YCbCr sample. Relying on the sampler being an LLVM constant is too fragile.
> * CreateImageYCbCrSample needs to re-think the image descriptor (T#) manipulation from the ground up.
> * Why ComponentSwizzle::Identity? This should be fixed and modified to be in line with the corresponding hardware fields.

@nhaehnle  Regarding to your  concerns:
1. I'm not sure if you've reviewed http://dxreview2.amd.com/r/21362/, that's a long discussion about the framwork on how to get yCbCrMetaData. With that changelist, the constant value is
really a singal which would be routing to ycbcr sampler path, and ycbcr is also the only case to
trigger it.
2. I've submitted a change http://dxreview2.amd.com/r/23432/ to eliminate identity enums, please help review, this would solve your second concern.

Thanks!

kuhar:
> As you've mentioned above, I'm not quite sure whether the newly added bits manipulating functions are sufficient(good) enough or not (so, thanks for your suggestions). But i'm sure they're correct, for that this patch has been verified by vk-cts with thousands of test cases, no test fails,
please refer to: dEQP-VK.ycbcr.conversion.* if you're interested.

My main concern is that is takes a lot of effort to run CTS for external contributors, as there's not public buildbot/CI that we can use. So if we want to commit changes to the related code, it's difficult to know what tests can possibly be affected by them. The code in the PR doesn't mention `dEQP-VK.ycbcr.conversion.*` tests, and it's not obvious if it's exercised and tested at all. Even if you google this string, it's not quick to even find these tests unless you know what you are looking for.
Sorry for being a bit nitpicky, but note that VK-CTS is only a test suite, so it doesn't exhaustively check the implementation, so these changes were validated but not verified. Bit-fiddling code is, in general, a usual source of difficult to triage bugs and difficult to get 100% correct.

Having LIT/unit tests is great for documentation and lowering the bar for contributing future improvements to the code. I'm deeply concerned about long-term maintainability of this patch without any in-tree tests.

kuhar:
> As you've mentioned above, I'm not quite sure whether the newly added bits manipulating functions are sufficient(good) enough or not (so, thanks for your suggestions). But i'm sure they're correct, for that this patch has been verified by vk-cts with thousands of test cases, no test fails,
please refer to: dEQP-VK.ycbcr.conversion.* if you're interested.

> It would be glad if this patch could be merged quickly after revising some main problems. And then i'll take a longer time to refine the code (such as auto generate some gfxip related code and split ycbcr related module apart) to make it easy maintainable and understandable.

Thanks for the context and outlining your constraints.
Could you add a comment mentioning that `dEQP-VK.ycbcr.conversion.*` exercises this code?

jiaxie-amd:
> Thanks for the context and outlining your constraints.
> Could you add a comment mentioning that `dEQP-VK.ycbcr.conversion.*` exercises this code?

Yes, i'll patch the cts result.  Per your suggests, I'm working on adding some LIT tests to cover these newly added functions, thanks!

amdvlk-admin:
### Test summary for commit eb8700a6689cd42653a222949d01c8b92cf9f1bd

<li>SPVGEN: dev f49024d7b85368dd156a9eef596007b66fe67fc2</li>
<li>XGL: dev 052ebfa2e77abd14fdb4f5120b35b081f3fc4bdc</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev a163b38723cbc05f3014d4eaa1936c82bbfbf3ea</li>
</ul></details>

<li>Passed: 945/945 (100.0%)</li>
<li>Failed: 0/945 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

<li>Passed: 65431/65431 (100.0%)</li>
<li>Failed: 0/65431 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle:
>     1. I'm not sure if you've reviewed http://dxreview2.amd.com/r/21362/, that's a long discussion about the framwork on how to get yCbCrMetaData. With that changelist, the constant value is
>        really a singal which would be routing to ycbcr sampler path, and ycbcr is also the only case to
>        trigger it.

No. I followed some email threads and expressed concerns there, but presumably those were ignored. It's short-sighted to say that ycbcr is the only case to trigger it.


>     2. I've submitted a change http://dxreview2.amd.com/r/23432/ to eliminate identity enums, please help review, this would solve your second concern.

Thank you.

I'm still concerned about the way the code is dissecting and reassembling image descriptors and consider this a blocking issue. The code does not correctly take different hardware generations into account, and probably the only feasible way of doing that is to encapsulate the dissection of a single image descriptor, and the reassembly of a single image descriptor, into their own functions.

nhaehnle:
> But as @nhaehnle pointed out, I'm more concerned about how to easily maintain these code, for that the above code only works for specifc Graphics card(Gfx9), due to the bits in img srd is different among Gfxips. Which means this yCbCr smapling path needs update when new Gfxip added in. I have the patch ready for Gfx10, but not uploaded, but what about latter? So, I want to solve some trival issues as you guys raised and then focus on giving a good solution for easily maintain these code with latest gfxip.
> 
> It would be glad if this patch could be merged quickly after revising some main problems. And then i'll take a longer time to refine the code (such as auto generate some gfxip related code and split ycbcr related module apart) to make it easy maintainable and understandable.

Sorry, I missed that part in my earlier reply. I'm uncomfortable with having this patch merged as-is without doing the gfxip-related fixes. I think we should hold ourselves to a higher standard than that. We'll probably have to discuss this further.

amdvlk-admin:
### Test summary for commit 10c7db6119fb5ac2600f4a64bb22922b2120f6a1

<li>PAL: dev 45f531beaf2c2b0bc2272e63a2da0022f1b07ccf</li>
<li>SPVGEN: dev 81f6794e048edd31d8c03343b9ed10dd4c0331ca</li>
<li>XGL: dev 8024f27f9457e3235bf4fcde0d2879bbaae7b0f2</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev a163b38723cbc05f3014d4eaa1936c82bbfbf3ea</li>
</ul></details>

<li>Passed: 945/945 (100.0%)</li>
<li>Failed: 0/945 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

<li>Passed: 65431/65431 (100.0%)</li>
<li>Failed: 0/65431 (0.0%)</li>
</ul></details>


</ul></details>

JaxLinAMD:
retest this please

llpc/builder/llpcBuilderImplImage.cpp
```diff
@@ -1185,6 +1185,1185 @@ Value* BuilderImplImage::CreateImageStore(
...
+
+    pSampleInfo->pImageDesc = wrapInfo.pImageDesc1;
+
+    Value* pResult = nullptr;
+
+    if (wrapInfo.planeNum == 1)
+    {
+        pSampleInfo->pImageDesc = wrapInfo.subsampledX ? wrapInfo.pImageDesc2 : wrapInfo.pImageDesc1;
+
+        Instruction* pImageOp = static_cast<Instruction*>(YCbCrCreateImageSampleInternal(coordsChroma, pSampleInfo));
```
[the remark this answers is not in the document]
trenouf: Hi Harris. The reason to use "cast" here (and in all the other cases) is that Value and its subclasses  participate in llvm's roll-your-own RTTI system, so you can use dyn_cast and cast on it. The advantage of cast (over static_cast) is that it asserts if it is not an Instruction.

nhaehnle:
Are you going to make the changes we discussed?

jiaxie-amd:
Hi Nicolai,

Yes, I want to abandon the previous #471 PR, and upload new changes which contains all the things we discussed.
But I’m busy on another urgent thread, so the code would be delayed.

BR,
Harris

trenouf:
A couple of issue that have cropped up for me with this change.

Shader compilation
--

With the move towards shader compilation without pipeline state (#431), we will not be able to use shader compilation and linking with YCbCr conversion. We don't know at shader compilation time whether YCbCr is in use, and we don't want to insert code for it on every image sample in every shader just in case. As far as I know, there is no shader capability in spir-v for this; the extension does not need spir-v changes at all.

The only fixes I can think of at the moment are:

1. For the extension as currently defined, fail at pipeline link time if any converting samplers are present in the resource layout. The front-end will then have to compile the pipeline in whole-pipeline mode from scratch.

2. We could define a whole new extension, same as the old one, except that we add a spir-v capability that the shader has to declare. Then, shader compilation can add conditional conversion code on every image sample in that shader.

Not inlining
--

We are working on allowing a compute shader to have subfunctions that are not inlined. You can have a shader that passes a pointer to a sampler into a subfunction, and that breaks the code in llpcBuilderImplDesc that thinks it can trace back from the descriptor load back to the descriptor pointer creation (where the set and binding are).

I guess the solution for now is:

1. Only attempt that scanning back if there are any converting samplers in the resource layout.

2. If there are any converting samplers, and the scanning back does not find the set,binding for the sampler, then fail as now.

The proper solution in the future would be to reformulate the conversion code to work with a variable converting sampler, and then rely on constant propagation etc to optimize it in the common case that the sampler is in fact constant.

nhaehnle:
> With the move towards shader compilation without pipeline state (#431), we will not be able to use shader compilation and linking with YCbCr conversion. We don't know at shader compilation time whether YCbCr is in use, and we don't want to insert code for it on every image sample in every shader just in case. As far as I know, there is no shader capability in spir-v for this; the extension does not need spir-v changes at all.
> 
> The only fixes I can think of at the moment are:
> 
>     1. For the extension as currently defined, fail at pipeline link time if any converting samplers are present in the resource layout. The front-enjustd will then have to compile the pipeline in whole-pipeline mode from scratch.

Yes. The YCbCr extension is an example of a horribly designed spec, and for anything related to relocations or separate compilation we should just assume that it isn't used.


>     2. We could define a whole new extension, same as the old one, except that we add a spir-v capability that the shader has to declare. Then, shader compilation can add conditional conversion code on every image sample in that shader.

If we go down the route of a new extension, we should be truly explicit about it and require a different image type to be used in SPIR-V for YCbCr.


> ## Not inlining
> 
> We are working on allowing a compute shader to have subfunctions that are not inlined. You can have a shader that passes a pointer to a sampler into a subfunction, and that breaks the code in llpcBuilderImplDesc that thinks it can trace back from the descriptor load back to the descriptor pointer creation (where the set and binding are).
> 
> I guess the solution for now is:
> 
>     1. Only attempt that scanning back if there are any converting samplers in the resource layout.
> 
>     2. If there are any converting samplers, and the scanning back does not find the set,binding for the sampler, then fail as now.

What are the places where we scan back in the first place? Is it done in more places than just for YCbCr?

trenouf:
> What are the places where we scan back in the first place? Is it done in more places than just for YCbCr?

It is also done in llpcPatchDescriptorLoad.cpp for general descriptor loading. I think that is unnecessary as long as the driver guarantees to us that a descriptor of the same spir-v type (image, sampler or combined) is always the same size and stride, which I think it does. The other types handled here (buffer, push const, etc) do not have that guarantee, and should be handled separately: you never get a pointer to one of those being passed around in spir-v. I think I designed the builder api a bit wrong there.

trenouf:
> ## Not inlining
> 
> The proper solution in the future would be to reformulate the conversion code to work with a variable converting sampler, and then rely on constant propagation etc to optimize it in the common case that the sampler is in fact constant.

With the current state of the extension definition, we can only make this work if we decide to make all samplers 8 dwords.

amdrexu:
@jiaxie-amd Is it PR still valid for review? I think you have made other changes to re-implement this extension?

jiaxie-amd:
Oh, this PR is no longer valid, there’s other PR’s re-implemented this extension.

Best Regard,
Harris

amdrexu:
OK. I close it.

PR #485, opened 2020-02-28
Coding standards doc and .clang-format file

trenouf:
This documents the LLVM-like coding standard we propose to switch to for LLPC (not XGL or PAL).

There was a bit of remaining internal debate on a couple of .clang-format points:

* indent 2 or 4? This doc goes for 2, the same as LLVM, but 4 was also suggested. The reason I prefer 2 is that, once you go for the LLVM convention of open brace not being on its own line (which is good to avoid wasting vertical space), there is confusion with 4 when an "if" condition takes more than one line, in that it is difficult to see where the continuation of the "if" condition stops and the body starts, as they have the same indent. Also I prefer consistency with LLVM.

* spaces inside braces? (I.e. {1, 2, 3} or { 1, 2, 3 }) The doc goes for no spaces, same as LLVM, but the format with spaces was suggested.

Comments?

### DIFF
--- a/docs/CodingStandards.md
+++ b/docs/CodingStandards.md
@@ -1,0 +1,170 @@
+# LLPC coding standard
+
+## Introduction
+
+This file documents the coding standard used by the LLPC project, that is:
+* LLPC itself
+* LGC
+* The VKGC utilities and definitions that have been separated out of LLPC.
+
+LLPC coding standard is generally based on
+[LLVM coding standard](https://llvm.org/docs/CodingStandards.html),
+with some specific changes. One notable change is that LLPC follows
+[MLIR](https://github.com/tensorflow/mlir/blob/master/g3doc/DeveloperGuide.md)
+and LLD in using camelBack for variable and field names; this change is
+expected to be made in LLVM per
+[this proposal](https://llvm.org/docs/Proposals/VariableNames.html).
+
+## `using namespace`
+
+Do not use `using namespace` in a header file, even inside a namespace.
+
+Do not use `using namespace std` anywhere.
+
+See the
+[LLVM coding standard](https://llvm.org/docs/CodingStandards.html#do-not-use-using-namespace-std)
+for more details.
+
+## Integer types
+
+Usually you want an integer to be at least 32 bits. In that case use `int` or `unsigned`.
+Only use `int32_t` or `uint32_t` if you need an integer of _exactly_ 32 bits, for example
+in an externally-defined memory layout.
+
+## Identifier naming
+
+Names should be in CamelCase (initial capital) or camelBack (initial lower-case), depending on
+the category of identifier:
+
+* A *type name* (including classes, structs, enums, typedefs, etc) should be a noun phrase
+  in CamelCase (e.g. `Boat` or `TextFileReader`).
+
+* A *variable name* should be a noun phrase in camelBack (unlike current LLVM style).
+
+  Much LLVM code uses extreme abbreviation in variable names (e.g. `M` for a module); the LLVM coding
+  standard advises against this, and in LLPC we can avoid it by using the same name as the type but
+  in camelBack (e.g. `module`) or a somewhat abbreviated form (e.g. `inputMod`).
+
+  Do not use a type prefix on a variable name (e.g. p for pointer) (unlike PAL style).
+
+  A global or static variable name should start with an upper-case letter.
+
+* A *public member variable* (usually a struct member) should be named like a variable.
+
+* A *non-public member variable* should be named like a variable, but
+  with an extra `m_` prefix (unlike LLVM style, but like PAL style, and like LLDB and WebKit).
+
+* A *function name* or *method name* should be an imperative verb phrase in camelBack
+  (like LLVM style, but unlike PAL style).
+
+  There are some exceptions to the camelBack rule for historical reasons:
+
+  - Following subclasses of llvm::Instruction, a static method that creates an instance of the class is
+    called `Create` with an upper-case C.
+
+  - Builder and its subclasses follow llvm::IRBuilder in that any method whose name starts with Create
+    has an upper-case C. IRBuilder also uses an initial upper-case letter for a method that sets
+    or gets its internal state, but not a method that gets a type or constant.
+
+  Further exceptions outlined in the LLVM coding standard are where a class has a method that
+  mimics an STL class action, such as `push_back`. Such a method can keep the STL-like name.
+
+* A *constant* (technically a static variable or class/struct member variable, but `final` so it is used
+  like a constant) should be in CamelCase.
+
+* An *enumerator* (e.g. `enum { Foo, Bar }`) acts like a constant, so should start with an
+  upper-case letter. In addition, if the enum is not an enum class, and is not hidden inside a
+  class that suitably qualifies it, then each enumerator should have
+  a prefix corresponding to (not necessarily exactly the same as) the enum declaration name.
+
+Unlike in LLVM, an initialism (e.g. "NGG") should be treated as a word in identifier-naming
+rules. Thus, if it appears at the start of a variable or function name, it should be entirely
+lower-case (`nggControl`); if it appears at the start of a type or constant name, or anywhere
+other than the start in any name, it should have an initial upper-case letter (`NggControl`).
+
+For API compatibility, vkgcDefs.h, llpc.h and vfx.h have not been modified from PAL-style naming,
+which means they still
+contain struct members with a p prefix for pointer, and CamelCase for methods. Any method
+in .cpp files that implements or overrides a method in a class declared there, such as an
+ICompiler method, also keeps CamelCase rather than camelBack. Similarly the ElfReader/ElfWriter
+methods `ReadFromBuffer`, `GetSectionIndex`, `GetSymbolsBySectionIndex` and `GetSectionData`
+remain in PAL-style CamelCase. In addition, ElfReader's `ElfSymbol::pSymName` retains its
+PAL-style p prefix as it is directly referenced outside of LLPC/VKGC.
+
+## Local variable initialization
+
+A POD type local variable declaration should always be initialized, even if the initialized
+value is always overwritten in later code (for example in both legs of conditional code).
+A local variable of aggregate type can be initialized with `= {}` to zero all fields.
+
+A non-POD type with a sensible default constructor could also be initialized to make it
+obvious that you are running the default constructor, perhaps by appending `{}` to the
+variable name. This is considered unnecessary for well-known types with a sensible default
+constructor, such as the standard library and LLVM container types.
+
+## Braces
+
+Brace placement is covered by clang-format below.
+
+C++ allows the body of an if, else, while, do, for to be a single simple statement,
+without any braces. We follow the LLVM convention of only using an unbraced body if
+it is a single line, including comments. Otherwise the body should be braced. For
+an "if..else", if one of the two bodies is braced, the other one should be too. For
+an "if..else if..else if..else", if any body is braced, the others should be too.
+
+## Redundant comparisons
+
+Comparison of a bool with `false` or a pointer with `nullptr` is redundant and should
+be avoided. Use `!` instead where appropriate.
+
+## Redundant parentheses
+
+Certain cases of theoretically redundant parentheses around a subexpression are needed
+because compilers warn about them otherwise, for the good reason that C/C++ operator
+precedence makes it difficult to understand and easy to get wrong (in particular
+`&` `|` `^` vs comparison operators, and `&&` vs `||`). However redundant parentheses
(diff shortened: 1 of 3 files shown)


docs/CodingStandards.md
```diff
@@ -0,0 +1,162 @@
...
+
+Despite the look of much LLVM code, the LLVM coding standard does actually recommend
+placing a comment at the top of a function/method. :-)
+
+In LLPC, that is mandatory. Parameters should be commented in the main comment at
+the top of the function like this:
+
+```
+// @param hatstand : [in/out] The Hatstand object to modify
+```
```
kuhar: IMO there's little value in comments that basically repeat what the code says in a trivial way. Consider the example you gave; in code, it'll look something like this:

```
// @param hatstand : [in/out] The Hatstand object to modify.
// @param newName:  The new name.
void foo(Hatstand &hatstand, const std::string &newName) {
```
I think it's much more reasonable to use good argument names in the first place, and if the use of the parameter is not obvious, then give a higher-level explanation that connects this parameter with the function contract/logic.

Otherwise you end up writing and *having to maintain* the same thing twice for little benefit, which becomes basically this very quickly:
image

docs/CodingStandards.md
```diff
@@ -0,0 +1,162 @@
...
+- llvm files, in alphabetical order
+- system header files used with angle brackets, in alphabetical order.
+
+Clang-format enforces this order, if you run clang-format on that part of the file.
+
+You should use clang-format to format code modified or added in your change, but do not
+reformat any other code. Reviewers should reject a change that has spurious reformatting
+outside the lines affected by the substantive change, unless the whole point of the change
+is the reformatting. You can use
+[git clang-format](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format)
```
kuhar: Maybe point to llvm-project instead?
https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/git-clang-format

kuhar: Or just run something like: `git diff -U0 | clang-format-diff -p1 -i`

docs/CodingStandards.md
amdrexu: Initially, this was discussed if we should let some parameter comments just be vacant since the names had already shown how to use them. Later, we find this will lead developers less likely to do commenting because everyone could believe "there is no need of making comments for this or these parameters. They are so obvious". Hence, we require each parameter should have its own comments even if the words or sentences are trivial. This is also the reviewers' responsibilities to help code writer to better describe what they want to tell code readers especially for complex or tricky use of a special parameter.

amdrexu:
> As for the spaces, most of the projects I'm working on, including LLVM, use 2, so I'd just go with that. However, with a clang-format file I don't have strong preference either way, because can just rely on it doing a good job without actively thinking about it. Same for spaces inside braces.

Yes, clang-format will help us to do this formatting. What we would like to discuss is from the code readers' perspective, which choices are better for them to read the code lines. Clearly, you prefer 2 space indentation. It's good. What about the initializer style: `{X, Y, Z}` or `{ X, Y, Z }`?

docs/CodingStandards.md
nhaehnle: I agree with Rex here. While it's true that often, the parameter descriptions are ideally redundant, having this rule forces both the original writer of the code and the reviewer to think about whether there isn't something more interesting to comment about. Often, there is.

kuhar:
> Yes, clang-format will help us to do this formatting. What we would like to discuss is from the code readers' perspective, which choices are better for them to read the code lines. Clearly, you prefer 2 space indentation. It's good. What about the initializer style: `{X, Y, Z}` or `{ X, Y, Z }`?

I like `{X, Y, Z}` more. But my point was that I think it's not a big deal either way with clang-format in place :slightly_smiling_face:.

amdrexu:
Another point for 2-space indentation is that it is apparent the code reading is not easy like 4-space indentation. It is worsen when {} is removed for single-line statement. So we'd better to encourage developers adding necessary blank line between code pieces, such as this example:
```
D = D / 3;
C = C * 2;
if (C)
  A = B;
else
  A = C;
E++;
F += A;
```
If possible, could be
```
D = D / 3;

C = C * 2;
if (C)
  A = B;
else
  A = C;

E++;
F += A;
```
I think this is also for better code review to catch possible issues. Blank lines acts as marks between sentences.

kuhar:
In your example @amdrexu, I'd put braces in `if` and `else`. I personally only omit them when when there is no `else`.
This is orthogonal in my opinion, but I also like introducing gaps with empty lines when the code is conceptually performing a next step of some algorithm. I find it helps me with reading code by splitting functions into smaller blocks of related things. I think this ends up being what you are proposing most of the time.

Flakebi:
Why not tabs instead of spaces? That way everyone can decide for himself what looks best. And you can even change based on context ;).

amdrexu:
> Why not tabs instead of spaces? That way everyone can decide for himself what looks best. And you can even change based on context ;).

We encountered such issues when developing OpenGL driver. Eventually, tabs and spaces are mixed everywhere introduced by careless developers. If you don't configurate the width correctly, the codes are in a messy view. So we disable the use of tabs now. This is also specified in PAL coding conventions.

Flakebi:
I think mixing different indentation and code styles can be easily countered by checking this in CI, which we should do regardless of which style we use.

Anyway, what do you think about adding a `.editorconfig`? It allows editors to automatically pick up style guidelines for projects and it’s supported by [most editors out there](https://github.com/editorconfig). Visual Studio and GitHub are using it automatically without a plugin.
It looks like this:
```editorconfig
root = true

[*]
end_of_line = lf
insert_final_newline = true
charset = utf-8
indent_style = tab  # or space
indent_size = 2     # Used as tab-width or amount of spaces
trim_trailing_whitespace=true

# Can configure by file type
[Makefile]
indent_size = 4
```

kuhar:
Sorry, I thought that @Flakebi was joking about tabs instead of spaces. Please no tabs, I prefer any number of spaces for indentation over tabs.
On top of what @amdrexu said, in practice you often end up using different editors than your main one (e.g., github, sending code over email/chat, gdb/lldb, etc.), where different indentation is super annoying.

docs/CodingStandards.md
s-perron: Maybe git-clang-format has changed since I last downloaded it, but I thought it formatted the entire file for every file that is changed, not just the lines that were change in those files.  So git clang format may not do what you want.

s-perron: If we are changing a file that has not been formatted to match the new standard, should be do a PR that just formats that file first, and then make our change?

Will there a mass reformatting taking place?

trenouf: Yes there will be a mass reformatting (together with a hacked clang-tidy and a script that you can use in git filter-branch on your own local changes so you can rebase, or you can ask me to do that to your local changes).

s-perron: Perfect.  I'm in favour of a mass rebase.  Short-term pain for long-term gain.

jayfoad: git-clang-format can reformat just the lines that you've changed: https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/git-clang-format#L38

I usually run `git clang-format HEAD^` after committing.

amdrexu:
Once merged, please don't use clang-format file added by this PR until Tim finishes all coding style transition changes.

amdrexu:
I find defining local variables without initialization may be common in LLVM. Perhaps, the Clang compiler will check this and report possible warning. I understand some cases are in purpose because the author wants each branch could assign a value to this variable and make it finally initialized without exceptions. But a better way is to make assert to check this. So I hope we can force initialization for all local variables, giving them reasonable initial values before making any further operations. Using "={}" for structure and array initialization is even necessary because aggregated types are easy to be extended. PAL style highlighted this in that we have encountered many cases randomly occur because of missing initialization. In shader compiler, this might just lead to some miscalculation or mistaken data flow. But in driver, register could be wrongly set, causing system hang that is very difficult to reproduce and diagnose. You can say the C++ compiler is not configured correctly or it is not so powerful to detect such cases. But as a good habit, developers ought to follow it and rely on themselves, not the compiler.

jayfoad:
> I find defining local variables without initialization may be common in LLVM.

Yes it's common to write for example:
```
Register MyReg;
```
because I know that `Register` is a class type with a useful default initializer.

> PAL style highlighted this in that we have encountered many cases randomly occur because of missing initialization.

You could use tools like `valgrind` to test for this kind of problem.

trenouf:
> Register MyReg;

For that case, can you write `Register MyReg()` ? Would that error (as we want) if Register is a POD type and does not have a useful default constructor?

amdrexu:
> because I know that Register is a class type with a useful default initializer.

Yes, classes and structures with explicit constructors defined could be waived as long as all fields and members are correctly initialized.

Flakebi:
`Register MyReg();` defines a function called `MyReg` that returns a `Register` ;)

Flakebi:
Can we enable [`-Werror=uninitialized`](https://clang.llvm.org/docs/DiagnosticsReference.html#wuninitialized)?
This errors when a variable or even part of a struct is not initialized but used afterwards (or returned from a function) and it allows initialization in different branches, so the following code compiles without problems.
```cpp
struct A {
    int a;
    int b;
};

A returnA(int num) {
    A a;
    a.a = 1;
    if (num == 5) {
        a.b = 1; // Remove this line to get a compiler error
    } else {
        a.b = 0;
    }
    return a;
}
```

amdrexu:
If A has a new field "int c" in the future (e.g. someone needs this field in particular cases), the code might be broken by compiler warning. You have to fix it. If `A a = {}`, the code will be elegantly extended as expected.

trenouf:
> If A has a new field "int c" in the future (e.g. someone needs this field in particular cases), the code might be broken by compiler warning. You have to fix it. If `A a = {}`, the code will be elegantly extended as expected.

Right, but Sebastian's error would catch a case that we missed in review.

trenouf:
V2: * indent 2
        * Update which files/methods have not had identifier renaming for
          API compatibility
        * Local variable initialization
        * if..else bracing
        * lgc #include ordering
        * using namespace in header file

amdvlk-admin:
### Test summary for commit bfab5999472e591d75d7d7f971a5d880fc8ccc61

<li>PAL: dev bb7398b1c7509a30ba4f7d947b5ad5549a4d58db</li>
<li>SPVGEN: dev dc2b146de03ef54f188172d8fa2f9be708cbfaca</li>
<li>XGL: dev ffd39443fec0ce821b7702173eba7184855c7b5e</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 1ebc431ac299fa294a4e730432520d37f326843c</li>
</ul></details>

<li>Failed: 0/955 (0.0%)</li>
</ul></details>

<li>Passed: 103137/103137 (100.0%)</li>
<li>Failed: 0/103137 (0.0%)</li>
</ul></details>

<li>Passed: 103193/103193 (100.0%)</li>
<li>Failed: 0/103193 (0.0%)</li>
</ul></details>


</ul></details>

docs/CodingStandards.md
```diff
@@ -0,0 +1,165 @@
...
+- `#include` blocks are merged together (removing blank lines) before sorting
+- `#include` directives are sorted in this order:
+  - the `.h` file for the current `.cpp` file;
+  - all other ones;
+  - `lgc/*` files;
+  - `llvm/*` files;
+  - standard library.
+
+You should use clang-format to format code modified or added in your change, but do not
+reformat any other code. Reviewers should reject a change that has spurious reformatting
```
kuhar, line 162: Note: a CI bot could be added that checks for that in the future, to reduce the turnaround times.

trenouf:
V3: Ban 'using namespace std' anywhere. Refer to namespace section in
        LLVM doc.

amdrexu:
I am not sure if there are coding instructions about the comments for structure fields and class data members in LLVM. In practice, if we can infer their functions from variable names, comments are not necessary. But sometimes, the function is ambiguous. If we have a "offset" field without any descriptions, we don't have an idea about what the `offset ` actually is. Is it based on DWORDs or bytes? :)

```
// Comments to describe this structure.
struct X
{
    int a; // Are comments always required?
};

// Comments to describe this class.
class Y
{
    int m_a; // Are comments always required?
};
```

kuhar:
@amdrexu I think that the direction LLVM took is to leave things like this up to developers and reviewers, i.e., it's always a judgement call. Personally, I'd prefer to shift this burden to reviewers over having many repetitive comments and keeping them in sync with the code. Just my 2c.

amdrexu:
@kuhar Thanks. It's clear to me now.

trenouf:
This passed testing here and internally as part of #543, so I'm going to merge it anyway.

PR #489, opened 2020-03-03
Move non-LLPC declarations and code out of llpc directory  (step 1).

trenouf:
llpc.h seperation. llpc-sepecific code in llpc.h and others used by the
XGL driver without building LLPC in compiler/include/vkgcDefs.h.

Change-Id: I5216ac38db9a384d179b4c8ccad2ac783ad8814d


trenouf:
This is Xue's change.

trenouf:
I believe the build fails because the spvgen CMakeLists.txt is confused by the re-introduction of an include directory in llpc. Hopefully fixed by spvgen change https://github.com/GPUOpen-Drivers/spvgen/pull/9

trenouf:
Ping.

PR #495, opened 2020-03-09
Update tests using amdgcn.cube* intrinsics in prep for llvm change

dstutt:
Updates to the const folding of amdgcn.cube* intrinsics in an upcoming llvm
version mean that these tests can't use constant arguments - changing to use
non-const to preserve the intent of the test.


jayfoad:
Looks good, thanks!

PR #496, opened 2020-03-09
Remove duplicate definitions of Llpc::{InvalidValue,SizeOfVec4}

nhaehnle:
There is a second set of definitions in llpcUtil.h, which could cause
conflicts going forward.

### DIFF
--- a/llpc/builder/llpcBuilderImplInOut.cpp
+++ b/llpc/builder/llpcBuilderImplInOut.cpp
@@ -32,6 +32,7 @@
 #include "llpcBuilderImpl.h"
 #include "llpcInternal.h"
 #include "llpcPipelineState.h"
+#include "llpcUtil.h"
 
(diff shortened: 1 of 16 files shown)


nhaehnle:
So how do you suggest to resolve the conflict? Move those definitions into another header file?

trenouf:
What conflict? There should be nowhere that includes both llpcInternal.h and llpcUtil.h, as the former is destined for the lgc directory.

trenouf:
Oh, you mean a conflict in your other change because you need something from llpcInternal.h in the elf linking code. The underlying problem is that the elf linking code is in the wrong place -- it should be in the middle-end, as it has middle-end knowledge.

nhaehnle:
Yeah, that's a fair point. I'll have to rework that other PR anyway once I understand what the intention is there, so closing this one.

PR #497, opened 2020-03-09
Clarify the use of descriptor set indices in the PAL metadata

nhaehnle:
The PAL metadata contains a userdata mapping which tells PAL the data to
load into each loaded userdata SGPR. There are two kinds of values:
    
1. Indices into the userdata table maintained by the client driver via Pal::ICmdBuffer::CmdSetUserData
2. Special "system" values as listed in Util::Abi::UserDataMapping
    
Pointers to descriptor sets fall into the first category. The client driver
tells the compiler which location in the userdata table will contain the
pointer to each descriptor set (via the ResourceMappingNode part of the
pipeline build info).
    
When compiling a partial pipeline, the layout of the userdata table may not
be known yet. The compiler can already decide which SGPR will hold the
pointer to the descriptor set (the "key" part of the metadata), but 
does not yet know the final "value" part of the metadata.
    
To solve this problem, introduce a new kind of value, the
VkDescriptorSetIndex, which is emitted as part of the PAL metadata during
the initial compiler. It is resolved, when linking the different parts of a
pipeline together.
    
The emitDescriptorSetIndexInMetadata option instructs the compiler to
generate these VkDescriptorSetIndex mapping.
    
Resolving any such mappings during linking happens unconditionally while
merging the metadata.
    
This commit integrates the "updating" of the "meta note" into the merging
of the metadata, as this is conceptually one operation. Some method names
are updated to better reflect the asymmetry in what they do.
    
The updateDescInElf option is renamed to emitDescriptorSetIndexInMetadata
to better reflect the purpose of the option.

### DIFF
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -1270,14 +1270,6 @@ Result Compiler::BuildPipelineInternal(
         graphicsShaderCacheChecker.UpdateAndMerge(result, pPipelineElf);
     }
 
-    if ((result == Result::Success) &&
-        pFragmentShaderInfo &&
-        pFragmentShaderInfo->options.updateDescInElf &&
-        (pContext->GetShaderStageMask() & ShaderStageToMask(ShaderStageFragment)))
-    {
-        graphicsShaderCacheChecker.UpdateRootUserDateOffset(pPipelineElf);
-    }
-
     pContext->setDiagnosticHandlerCallBack(nullptr);
 
     return result;
@@ -1335,19 +1327,6 @@ uint32_t GraphicsShaderCacheChecker::Check(
 }
 
 // =====================================================================================================================
-// Update root level descriptor offset for graphics pipeline.
-void GraphicsShaderCacheChecker::UpdateRootUserDateOffset(
-    ElfPackage*       pPipelineElf)   // [In, Out] ELF that could be from compile or merged
-{
-    ElfWriter<Elf64> writer(m_pContext->GetGfxIpVersion());
-    // Load ELF binary
-    auto result = writer.ReadFromBuffer(pPipelineElf->data(), pPipelineElf->size());
-    assert(result == Result::Success);
-    (void(result)); // unused
-    writer.UpdateElfBinary(m_pContext, pPipelineElf);
-}
-
-// =====================================================================================================================
 // Update shader caches for graphics pipeline from compile result, and merge ELF outputs if necessary.
 void GraphicsShaderCacheChecker::UpdateAndMerge(
     Result            result,         // Result of compile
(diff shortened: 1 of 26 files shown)


nhaehnle:
Some notes:

* This is on top of https://github.com/GPUOpen-Drivers/llpc/pull/496, which is a minor cleanup
* I would be interested in the rationale for why the use of descriptor set indices is not required for the non-fragment part of a pipeline -- I don't see why this is true.

... and now that I posted this, I think I understand why the separate later update was required. I'm going to fix that up.

amdvlk-admin:
### Test summary for commit 3c62770e254b7eba1091f688c241597d091206ba

<li>PAL: dev 45f531beaf2c2b0bc2272e63a2da0022f1b07ccf</li>
<li>SPVGEN: dev f2d32c98de40d445656b41f14756f207264cc8bb</li>
<li>XGL: dev 8024f27f9457e3235bf4fcde0d2879bbaae7b0f2</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 878d4bd14afe9bc670911954adfb7c16b71e7cae</li>
</ul></details>

<li>Failed: 0/960 (0.0%)</li>
</ul></details>

<li>Passed: 103193/103193 (100.0%)</li>
<li>Failed: 0/103193 (0.0%)</li>
</ul></details>

<li>Passed: 103193/103193 (100.0%)</li>
<li>Failed: 0/103193 (0.0%)</li>
</ul></details>


</ul></details>

llpc/context/llpcCompiler.cpp
```diff
@@ -1270,14 +1270,6 @@ Result Compiler::BuildPipelineInternal(
         graphicsShaderCacheChecker.UpdateAndMerge(result, pPipelineElf);
     }
 
-    if ((result == Result::Success) &&
-        pFragmentShaderInfo &&
```
amingriyue: nit: pFragmentShaderInfo isn't used any more, so you can remove it together.

amdrexu:
I think you need a rebase to remove the previous commit that you have abandoned.

nhaehnle:
> I think you need a rebase to remove the previous commit that you have abandoned.

You're right. I first want to improve my understanding of other aspects of this feature, then I will deal with this pull request.

JaxLinAMD:
retest this please

trenouf:
I think this needs a bit of a rethink, in line with what Yong is looking at. For his requirements (essentially shader compilation), there will be no user data layout available, so PatchEntryPointMutate needs to deal in descriptor set numbers rather than user data offsets. For maintainability and improved test coverage, I think we should always do that, even when doing pipeline compilation. I am looking at this now.

nhaehnle:
Yeah, makes sense. It'd then be good to cleanup how way those are represented in the PAL userdata mapping metadata. Meanwhile, I should really close this PR to avoid confusion...

PR #518, opened 2020-03-20
Use a vertex fetch shader for relocatable shaders

s-perron:
- Get vertex shader inputs as parameters

For relocatable elf, we want to generate a "fetch shader" that will load
all of the vertex inputs into registers.  When that happens, the vertex
inputs should not be loaded from memory anymore, but they should be
passed in as parameters.  This commit changes to codegen for vertex
shader to do that.

- Add the fetch shader stage to LLPC

- Copy VS interface data for use by the fetch shader.

- Generate the entry point for the fetch shader
  - Read the SGPR and VGPR usage from the .note of the VS.
  - Pass that information to middle-end
  - Generate the loads of the vertex input while reading the spir-v.
  - Use the information in during entry-point-mutation to generate update
    the entry point and return instruction for fetch shaders.

- Link the fetch shader and the vertex shader


amdvlk-admin:

s-perron:
@trenouf Here is some of the reason for why I did things the way I did.

1. We need to build the fetch shader without any spir-v, but many passes in the FE fail if there is no function in the llvm-ir.  They expect there to be at least one entry point in order to get information like which shader stage they are compiling for.   
    1. This forced me to build the entry point for the fetch shader early in the FE, and to update some passes to behave differently for fetch shaders.
    1. When generating just a fetch shader, we could replace the passes for the current FE with just a single FE pass that generates the fetch shader, but I seem to remember a desire to have a single pass manager.
1. To build the call to create a call to an LLPC built-in function (see [here](https://github.com/GPUOpen-Drivers/llpc/pull/518/files#diff-63619b54afe227226eed871065663b95R185)).  This requires the llpc builder, which is only available in the FE.  Correct me if I am wrong.
    1. This forced the generation of part of the body into the FE, and I figured I would keep it with the generation of the entry point.
    1. We could change the code that expands the built-in to look at the side data, but this complicate entrypoint mutation because the values it is suppose to return have not been generated yet.

trenouf:
Hi @s-perron 

> @trenouf Here is some of the reason for why I did things the way I did.
> ... 
> 2. To build the call to create a call to an LLPC built-in function (see [here](https://github.com/GPUOpen-Drivers/llpc/pull/518/files#diff-63619b54afe227226eed871065663b95R185)).  This requires the llpc builder, which is only available in the FE.  Correct me if I am wrong.

Hmm, I see what you mean. I guess my idea for doing the whole thing in the middle-end is that it would generate that call in the same way as llpcBuilderImplInOut does, generating an llpc.input.builtin.* call. But I think doing it in the middle-end (LGC) needs a specific "given this pipeline and this vertex shader elf, generate me a fetch shader elf" interface in Pipeline. That's the kind of interface change I'm thinking of.

I haven't looked closely at this PR yet, but does it do something to ensure that (a) the fetchless vertex shader gets vertex id and instance id as inputs, and vertex buffer table pointer as a user data input, and (b) the fetch shader has the same inputs as the vertex shader (minus the vertex fetches that you're passing through)?

s-perron:
> does it do something to ensure that (a) the fetchless vertex shader gets vertex id and instance id as inputs, and vertex buffer table pointer as a user data input, and (b) the fetch shader has the same inputs as the vertex shader (minus the vertex fetches that you're passing through)?

Yes, it does both (a) and (b).

s-perron:
@trenouf Also please look at how I am passing data from the vertex shader compile to the fetch shader compile.  I pull out some information from the `.note` section [here](https://github.com/GPUOpen-Drivers/llpc/pull/518/files#diff-b778ad5a9dd33b45c954ca441626f5c3R930).  I pass that information through to the compiler using the builder context.  

This the builder context the appropriate container?  Should I be passing the how vertex shader elf instead, and parse it in the pass that uses the data?

llpc/context/llpcCompiler.cpp
```diff
@@ -2146,7 +2178,21 @@ void Compiler::LinkRelocatableShaderElf(
...
+            if (result != Result::Success)
+            {
+                return;
+            }
+            result = writer.LinkGraphicsRelocatableElf({&vsReader, &fsReader, &fchReader}, pContext);
+        }
+        else
+        {
+            result = writer.LinkGraphicsRelocatableElf({&vsReader, &fsReader}, pContext);
+        }
```
nhaehnle: Graphics pipelines are now always required to have the fetch shader stage, aren't they? So please convert the check into an assertion and remove the obsolete code paths.

llpc/context/llpcCompiler.cpp
```diff
@@ -2126,6 +2157,7 @@ void Compiler::LinkRelocatableShaderElf(
     {
         ElfReader<Elf64> vsReader(m_gfxIp);
         ElfReader<Elf64> fsReader(m_gfxIp);
+        ElfReader<Elf64> fchReader(m_gfxIp);
         if (!pShaderElfs[ShaderStageVertex].empty())
```
nhaehnle: I didn't realize this before, but vertex shaders are non-optional in graphics pipelines, so... maybe change this condition to an assertion in a separate change?

llpc/context/llpcCompiler.cpp
```diff
@@ -2167,4 +2213,75 @@ void Compiler::LinkRelocatableShaderElf(
...
+            }
+            else if (registerInfo.second.getUInt() == static_cast<u_int64_t>(Util::Abi::UserDataMapping::BaseInstance))
+            {
+                pContext->GetBuilderContext()->SetBaseInstance(infoType-mmSPI_SHADER_USER_DATA_VS_0);
+            }
+        }
+    }
+
+    msgpack::MapDocNode& vertexInputInfoMap = vertexInputInfoNode.getMap(false);
+    for(auto vertexINputInfo : vertexInputInfoMap)
```
nhaehnle: Capitalization of vertexINputInfo

trenouf:
> I'm not entirely sure about the vsInterfaceData, but I have a feeling that it fits better with the Llpc::InterfaceData and Llpc::ResourceUsage, i.e., with the PipelineState.
> 
> Involving the frontend in the generation of the fetch shader feels architecturally wrong for LLPC. Consider that at some point, we likely want LLPC to be able to consume inputs other than SPIR-V. This may mean refactoring some of BuildPipelineInternal to better reflect the separation of the frontend from later passes, but that's okay. Something like:
> 
> * Split the frontend (SPIRV-to-LLVM and SPIR-V lowering) parts of BuildPipelineInternal from the middle (patch) and back-end (codegen) parts.
> * Generate the fetch shader IR directly from Llpc::Compiler and go straight to the middle/back-end.

I would go further than this. I think this should be entirely within the middle-end. So there would be a middle-end (Pipeline) interface that says "link these elfs given the pipeline state", and internally it would generate the fetch shader (if necessary) and run middle-end and back-end passes on it to generate the fetch shader elf.

Steven had concerns about this approach to do with being able to get at the vertex index. So I need to look into this further.

s-perron:
> I'm not entirely sure about the vsInterfaceData, but I have a feeling that it fits better with the Llpc::InterfaceData and Llpc::ResourceUsage, i.e., with the PipelineState.

Some of my decisions are based on not wanting to change the process to build a pipeline in BuildPipelineInternal too much.  I looked into trying to put it into the pipeline state.  As you said, that seems like the most natural place.  However, because I am building each shader separately they each end up with their own pipeline which is created inside build pipeline internal. 

As for doing everything in the middle-end, I'll try to move more stuff into the middle end.  Part of that was not wanting to mess with build pipeline internal because it is very much spaghetti code.  I find there is a lot going on, and it is hard to follow.  Putting an extra set of conditional paths in it might make it more difficult.  However, now better understand where you are going with the separation of the FE and LGC, so I'll make the changes that are needed.

nhaehnle:
> Part of that was not wanting to mess with build pipeline internal because it is very much spaghetti code. I find there is a lot going on, and it is hard to follow.

I'm with you on that. The best case outcome of this is that we can get things slightly less spaghetti-y through this process ;)

llpc/util/llpcElfWriter.cpp
```diff
@@ -1164,7 +1164,7 @@ Result ElfWriter<Elf>::LinkGraphicsRelocatableElf(
 
 
         MergeSection(pTextSection3,
-                     alignTo(pTextSection1->secHead.sh_size, 0x10),
+                     alignTo(pTextSection1->secHead.sh_size, 0x4),
```
nhaehnle: This raises an interesting question. LLVM is able to align basic blocks for loops, and that can in fact have a performance benefit. Technically, up to 64 bytes of alignment can in some cases help. 256 bytes of alignment are only required for shader entry points.

Perhaps LLVM could emit a meaningful alignment for .text sections, and then this code could take that alignment into account when combining sections?

s-perron: Yes, that would be a good idea.  I don't know what the alignment should be.  For now, I lowered it to remove the number of NOPs to make debugging easier.

lgc/patch/PatchFetchShader.cpp
```diff
@@ -0,0 +1,294 @@
...
+  buildFetchShaderBody(entryPoint);
+
+  return true;
+}
+
+// =====================================================================================================================
+// Generate the body of a fetch shader in the current context.
+//
+// @param entryPoint [in]: The entry point to be modified.
+void PatchFetchShader::buildFetchShaderBody(Function *entryPoint) {
```
nhaehnle, line 112: Please change this method to use builder style consistently.

lgc/patch/PatchFetchShader.cpp
```diff
@@ -0,0 +1,294 @@
...
+// =====================================================================================================================
+// Load the vertex input at the given location and return that value.
+//
+// @param fetchType [in]: The type of the vertex input at the given location.
+// @param returnType [in]: The type that the returned value should be.
+// @param location [in]: The location of the vertex input to load.
+// @param insertPos [in]: Newly generated code will be inserted before this instruction.
+// @param vertexFetch [in]: An already initialzed VertexFetch object.
+Value *PatchFetchShader::fetchVertexInput(Type *fetchType, Type *returnType, unsigned location, Instruction *insertPos,
+                                          VertexFetch *vertexFetch) {
```
nhaehnle, line 271: Please change this method to take a builder argument and use that for the bitcast.

Changing vertexFetch->run to take a builder can happen later

lgc/patch/PatchInOutImportExport.cpp
```diff
@@ -113,6 +113,10 @@ bool PatchInOutImportExport::runOnModule(Module &module) {
   // set by TCS).
   auto pipelineShaders = &getAnalysis<PipelineShaders>();
   for (int shaderStage = ShaderStageCountInternal - 1; shaderStage >= 0; --shaderStage) {
+    if (shaderStage == ShaderStageFetch) {
+      continue;
+    }
```
nhaehnle, line 118: You don't need the braces here.

lgc/patch/PatchPeepholeOpt.cpp
```diff
@@ -512,26 +512,29 @@ void PatchPeepholeOpt::visitPHINode(PHINode &phiNode) {
...
-          newPhiNode->addIncoming(newIncomingValue, basicBlock);
-        } else if (Constant *const constant = dyn_cast<Constant>(incoming)) {
-          Constant *const extractElement = ConstantExpr::getExtractElement(constant, elementIndexVal);
-          incomingPairMap.insert({basicBlock, extractElement});
-          newPhiNode->addIncoming(extractElement, basicBlock);
-        } else
-          llvm_unreachable("Should never be called!");
+          incomingPairMap.insert({basicBlock, newIncomingValue});
+        }
+        newPhiNode->addIncoming(newIncomingValue, basicBlock);
```
nhaehnle, line 537: Can you explain this? It looks like an unrelated change.

lgc/interface/lgc/LgcContext.h
```diff
@@ -55,6 +57,82 @@ class PassManager;
...
+  // Get the type of the vertex input at the given location.
+  llvm::Type *getVertexInputType(unsigned location, llvm::LLVMContext *context) const;
+  const LocationToTypeMap &getVertexInputTypeInfo() const { return vertexInputType_; }
+
+  // Clears the types that have been recored for the vertex inputs.
+  void clearVertexInputTypeInfo() { vertexInputType_.clear(); }
+
+  // Get and set the number of the last SGPR that is used as an input to the vertex shader.
+  unsigned getLastVsSgpr() const { return lastSgpr_; }
+  void setLastInterfaceSgpr(unsigned i) { lastSgpr_ = i; }
```
nhaehnle, line 108: Please make getter/setter names consistent (also below).

lgc/patch/SystemValues.cpp
```diff
@@ -425,7 +425,7 @@ Value *ShaderSystemValues::getVertexBufTablePtr() {
   if (!m_vbTablePtr) {
     // Find the node.
     auto vbTableNode = findResourceNodeByType(ResourceNodeType::IndirectUserDataVaPtr);
-    if (vbTableNode) {
+    if (vbTableNode || m_shaderStage == ShaderStageFetch) {
```
nhaehnle, line 428: This conditional here is strange. We don't actually use vbTableNode in the code below.

I suspect we can (and should) change the code to instead just `assert(m_shaderStage == ShaderStageFetch || findResourceNodeByType(...))`.

lgc/patch/VertexFetch.cpp
```diff
@@ -897,7 +900,7 @@ bool VertexFetch::needPatchA2S(const VertexInputDescription *inputDesc) const {
 // Checks whether the second vertex fetch operation is required (particularly for certain 64-bit typed formats).
 //
 // @param inputDesc : Vertex input description
-bool VertexFetch::needSecondVertexFetch(const VertexInputDescription *inputDesc) const {
+bool VertexFetch::needSecondVertexFetch(const VertexInputDescription *inputDesc) {
```
nhaehnle, line 903: This looks like a spurious change -- maybe a leftover from a previous version?

llpc/context/llpcCompiler.cpp
```diff
@@ -29,14 +29,26 @@
  ***********************************************************************************************************************
  */
 #include "llpcCompiler.h"
+#include "../../lgc/imported/chip/gfx9/gfx9_plus_merged_offset.h"
+#include "../../lgc/imported/chip/gfx9/gfx9_plus_merged_registers.h"
```
nhaehnle, line 33: This looks wrong... @trenouf can probably comment.

llpc/context/llpcCompiler.cpp
```diff
@@ -1801,6 +1854,81 @@ void Compiler::linkRelocatableShaderElf(ElfPackage *shaderElfs, ElfPackage *pipe
...
+  }
+  // Add 1 for the offset of the scratch memory.
+  vsInterfaceData->setLastInterfaceSgpr(maxUserData - Pal::Gfx9::mmSPI_SHADER_USER_DATA_VS_0 + 1);
+
+  if (vertexInputInfoNode.getKind() == msgpack::Type::Nil) {
+    return;
+  }
+
+  msgpack::MapDocNode &vertexInputInfoMap = vertexInputInfoNode.getMap(false);
+  for (auto vertexINputInfo : vertexInputInfoMap) {
```
nhaehnle, line 1923: CApitalization

llpc/util/llpcElfWriter.cpp
```diff
@@ -1243,62 +1251,76 @@ void fixUpRelocations(ElfWriter<Elf> *writer, const std::vector<RelocationEntry>
...
       if (strncmp(sym.pSymName, "BB", 2) == 0) {
         continue;
       }
-      // When relocations start being used, we will have to filter out symbols related to relocations.
-      // We should be able to filter the BB* symbols as well.
       ElfSymbol *newSym = getSymbol(sym.pSymName);
       newSym->secIdx = m_textSecIdx;
       newSym->secName = nullptr;
-      newSym->value = sym.value + offset;
+      newSym->value = sym.value + (i == 0 ? 0 : offset);
```
nhaehnle, line 1323: This seems wrong to me. Since i == 0 is the vertex shader, shouldn't it be offset by the (non-zero) vertex shader part base, and i == 1 should be offset by the fragment shader part base?

Maybe this could be cleaned up a little by having mergeSection take an ArrayRef of parts and output a vector of offsets for each part?

Ah, or is this because _amdgpu_vs_main wants to be special-cased? But in that case, it'd feel cleaner to me to explicitly special-case it by just setting its value below.

llpc/util/llpcElfWriter.cpp
```diff
@@ -1370,6 +1407,73 @@ Result ElfWriter<Elf>::linkComputeRelocatableElf(const ElfReader<Elf> &relocatab
   return Result::Success;
 }
 
+// =====================================================================================================================
+// Merge the metadata for the fetch shader and reloctable vertex shader to get the meatadata for the resulting vertex
```
nhaehnle, line 1411: Typo: meatadata

include/vkgcDefs.h
```diff
@@ -151,12 +151,13 @@ enum class BasicType : unsigned {
...
-  ShaderStageFragment,    ///< Fragment shader
-  ShaderStageCompute,     ///< Compute shader
+  ShaderStageVertex = 0,                                ///< Vertex shader
+  ShaderStageTessControl,                               ///< Tessellation control shader
+  ShaderStageTessEval,                                  ///< Tessellation evaluation shader
+  ShaderStageGeometry,                                  ///< Geometry shader
+  ShaderStageFragment,                                  ///< Fragment shader
+  ShaderStageCompute,                                   ///< Compute shader
+  ShaderStageFetch,                                     ///< Fetch shader for vertex inputs
   ShaderStageCount,                                     ///< Count of shader stages
```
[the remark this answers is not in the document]
s-perron: I don't know what to do about this.  I changed 1 line and clang-format changed the rest.  If you want to avoid changes like this, we have to change the clang format file.  I don't think it is a good idea for us to pick and choose which clang format change to keep or revert.

lgc/interface/lgc/LgcContext.h
s-perron: Done

lgc/interface/lgc/LgcContext.h
```diff
@@ -55,6 +57,82 @@ class PassManager;
...
+  unsigned getVertexBufferRegister() const { return vertexBufferRegister_; }
+  void setVertexBuffer(unsigned i) { vertexBufferRegister_ = i; }
+
+private:
+  unsigned lastSgpr_;             // The last SGPR used as an input to the vertext shader.
+  unsigned baseVertexRegister_;   // The SGPR that contains the vertex offset.
+  unsigned baseInstanceRegister_; // The SGRP that contains the instance offset.
+  unsigned vgrpCompCnt_;          // The value of the VGRP_COMP_CNT in the vertex shader.
+  unsigned vertexBufferRegister_; // The SGRP that contains the address of the vertex buffer table.
+  LocationToTypeMap vertexInputType_; // A map from a location to the type of the vertex input at that location.
```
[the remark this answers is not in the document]
s-perron: Right, accidentally went back to the Google style guide.

lgc/patch/ConfigBuilderBase.cpp
```diff
@@ -361,6 +362,19 @@ void ConfigBuilderBase::writePalMetadata() {
...
+  // Add a map containing the type of each vertex input.
+  VsInterfaceData* vsInterfaceData = m_pipelineState->getLgcContext()->getVsInterfaceData();
+  const auto& vsInputTypeInfo = vsInterfaceData->getVertexInputTypeInfo();
+  if (!vsInputTypeInfo.empty()) {
+    msgpack::MapDocNode vertexInputTypes = m_pipelineNode[".vertexInputTypes"].getMap(true);
+    for (const auto &entry : vsInputTypeInfo) {
+      auto key = m_document->getNode(entry.first);
+      StringRef data(reinterpret_cast<const char *>(&entry.second), sizeof(entry.second));
+      auto value = m_document->getNode(data);
+      vertexInputTypes[key] = value;
```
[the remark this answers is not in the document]
s-perron: I've turned `value` into a msgpack array with two elements to represent the struct.

lgc/patch/PatchFetchShader.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "lgc/patch/Patch.h"
+#include "lgc/state/PipelineShaders.h"
+#include "lgc/state/PipelineState.h"
+#include "lgc/state/ResourceUsage.h"
+#include "lgc/state/TargetInfo.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "llpc-patch-copy-shader"
```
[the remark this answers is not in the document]
s-perron: Done

lgc/patch/PatchFetchShader.cpp
```diff
@@ -0,0 +1,294 @@
...
+bool PatchFetchShader::runOnModule(Module &module) // [in,out] LLVM module to be run on
+{
+  LLVM_DEBUG(dbgs() << "Run the pass Patch-Fetch-Shader\n");
+  Patch::init(&module);
+  m_pipelineState = getAnalysis<PipelineStateWrapper>().getPipelineState(&module);
+  m_builder.reset(new IRBuilder<>(module.getContext()));
+
+  if (!m_pipelineState->getLgcContext()->buildingFetchShader()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
s-perron: done

lgc/patch/PatchFetchShader.cpp
s-perron: Sorry, I don't know what you mean.  Can you be more specific?

lgc/patch/PatchFetchShader.cpp
s-perron: I have the builder available as a member variable, so I use that to get the insertion point, and to build the bit cast.  Is that good enough?  I do not pass it as a parameter.

lgc/patch/PatchInOutImportExport.cpp
s-perron: done

lgc/patch/PatchPeepholeOpt.cpp
s-perron: Now that the vertex inputs are coming in as argument, it is possible that the phi nodes have an incoming edge that is an argument. This code use to assume the incoming edges were either an instruction or a constant.  I added the extra case, and had to move some other code around to make it make more sense with the new case.

lgc/patch/VertexFetch.cpp
s-perron: I think that is right.  I'll undo this change.

llpc/context/llpcCompiler.cpp
s-perron: Yes I agree.  The problem is that those files are not accessible in to llpcCompiler.cpp, but while the linking code still lives in that file, we need access to it to read the metadata.  I should be removed when we move the linking to to lgc.

llpc/context/llpcCompiler.cpp
s-perron: done

llpc/util/llpcElfWriter.cpp
s-perron: Yes, I want to special case the vertex shader.  It should start at the beginning of the fetch shader, which will be included in the vertex shader, not at the location it is being copied.  I've fixed it up as you suggested.

llpc/util/llpcElfWriter.cpp
s-perron: done.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -799,6 +800,24 @@ FunctionType *PatchEntryPointMutate::generateEntryPointType(uint64_t *inRegMask)
     entryArgIdxs.vs.instanceId = argTys.size();
     argTys.push_back(Type::getInt32Ty(*m_context)); // Instance ID
 
+    // In relocatable shader, the vertex inputs are passed to the vertex shader
+    // in VGPRS.  Add them to the interface.
+    if (m_pipelineState->getLgcContext()->buildingRelocatableElf()) {
+      // TODO: See if we can store the vertex input descriptions on sorted order based on location.  It will make
+      // finding a particular input faster.
```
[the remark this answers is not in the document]
s-perron: I sort the list below.  That does not seem to cause problem, but it might be better if there is guarantee that they are always sorted.  What are your thoughts on that?  Is it already true and I didn't know?

include/vkgcDefs.h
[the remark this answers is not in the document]
nhaehnle: Okay, that's unfortunate. Just keep it as-is for the purpose of this change, no use fighting clang-format over this.

In the long run, I'd hope we could keep alignment to a minimum, especially in API-related enums where there is a good chance that different feature branches add to the same enum in different ways, and then whitespace changes just cause completely unnecessary merge pain.

lgc/patch/PatchFetchShader.cpp
nhaehnle: There's an IRBuilder<> in use, but I see a bunch of CastInst::CreateZExtOrBitCast, InsertValueInst::Create, ...; just use the corresponding m_builder->Create methods.

lgc/patch/PatchEntryPointMutate.cpp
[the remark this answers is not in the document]
nhaehnle: This seems to have disappeared? I really don't like the fact that Github won't show me the diff anymore :/

lgc/patch/PatchFetchShader.cpp
nhaehnle: Yes, that sounds fine.

lgc/patch/PatchEntryPointMutate.cpp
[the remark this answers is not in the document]
s-perron: Yes, I changed that part of the code to no reference the vertex input descriptions.  So it did become obsolete.  Too bad did that after I made that comment.

lgc/patch/PatchFetchShader.cpp
s-perron: Done

llpc/context/llpcCompiler.cpp
s-perron: I moved the code that needs these header files to llpcElfWriter.cpp, so we no longer include these in this file.  However, that file has the same issue.  It needs access to these header that are not naturally available.

lgc/patch/SystemValues.cpp
s-perron: done

llpc/context/llpcCompiler.cpp
trenouf: Linking and prolog/epilog generation all needs to be in LGC, not the front-end. I am hoping to look at that next.

JaxLinAMD:
@s-perron  there is one change request, also please resolve conflicts.

trenouf:
Hi @s-perron. Sorry for the delay; now I've got the ELF-linker-inside-LGC change essentially done and in early review (#673), I can look at this properly.

As already noted by yourself and Nicolai, the interface aspects and where things are defined doesn't really work in this change. With #673, it can now be rephrased to not need the front-end at all (other than it might want to cache fetch shaders).

It should not be defining any per-compile state in LgcContext, and, if we do need any new between-pass outside-IR state, we should endeavor to put it into something that hangs off PipelineState in a way that is serializable into IR metadata. (ResourceUsage does not do that currently.)

However, in this case, I don't think we need to do even that. The state left by a VS compile for use by a fetch shader compile needs to go into PAL metadata, as you have it. But the internal state of what vertex fetches are done does not need any between-pass state, because it can be derived in the middle-end by looking at lgc.input.import.generic calls in the VS.

Here's the approximate plan I'm thinking of:

Compiling a VS: It scans the lgc.input.import.generic calls and either generates the fetches (currently done later on in PatchInOutImportExport), or generates the input args to the shader and the PAL metadata to tell the fetch shader compilation what happened. I think this can be a new pass, probably just before PatchEntryPointMutate. We would change PatchEntryPointMutate to cope with a shader already having args -- it would put its new args before the existing ones.

Generating a fetch shader: The ELF linker spots it needs a fetch shader, and it generates a new IR module and runs it through certain middle-end passes and the back-end. That code reads PAL metadata in the VS.

We could keep all the vertex fetch stuff together in the existing source file VertexFetch, adding in both the new pass and the code to generate the fetch shader IR.

Given my #673 has completely broken the current state of the change, I'll have a look at doing the above.

PR #531, opened 2020-03-31
Remove some SPIR-V assembly tests

amdrexu:
Those tests are written in SPIR-V assembly. With the update of SPIR-V
tools, the validation rules are strengthened in control flow.

The branch tests are from old dumps of Vulkan CTS. This is the same for
fuzzer tests. Since the tests are being verified by CTS all the time, we
can safely remove those invalid ones. Still, we ought to minimize the
amount of SPIR-V assembly tests because they are fragile.


amdrexu:
We need this fix urgently since it broken internal build after we update SPIR-V tools from latest sources.

PR #536, opened 2020-04-02
Be explicit about signedness in CreateShuffleVector

piotrAMD:
This is in preparation for https://reviews.llvm.org/D72467
which introduces also the signed version of mask indices in
CreateShuffleVector.

### DIFF
--- a/lgc/builder/llpcBuilderImplSubgroup.cpp
+++ b/lgc/builder/llpcBuilderImplSubgroup.cpp
@@ -191,7 +191,9 @@ Value* BuilderImplSubgroup::CreateSubgroupBallot(
     pBallot = CreateBitCast(pBallot, VectorType::get(getInt32Ty(), 2));
 
     ElementCount elementCount = pBallot->getType()->getVectorElementCount();
-    return CreateShuffleVector(pBallot, ConstantVector::getSplat(elementCount, getInt32(0)), { 0, 1, 2, 3 });
+    return CreateShuffleVector(pBallot,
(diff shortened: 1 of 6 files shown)


lgc/builder/llpcBuilderImplSubgroup.cpp
```diff
@@ -191,7 +191,7 @@ Value* BuilderImplSubgroup::CreateSubgroupBallot(
     pBallot = CreateBitCast(pBallot, VectorType::get(getInt32Ty(), 2));
 
     ElementCount elementCount = pBallot->getType()->getVectorElementCount();
-    return CreateShuffleVector(pBallot, ConstantVector::getSplat(elementCount, getInt32(0)), { 0, 1, 2, 3 });
+    return CreateShuffleVector(pBallot, ConstantVector::getSplat(elementCount, getInt32(0)), ArrayRef<uint32_t>{ 0, 1, 2, 3 });
```
amdrexu, line 194: Pay attention to the line limit: 120 character. More characters are added by "ArrayRef<uint32_t>" :) Please check all modifications.

piotrAMD:
> Would it work to say
> 
> { 0U, 1U }
> 
> instead of
> 
> ArrayRef<uint32_t>{ 0, 1 }
> 
> ?

Nope.

trenouf:
> > Would it work to say
> > { 0U, 1U }
> > instead of
> > ArrayRef<uint32_t>{ 0, 1 }
> > ?
> 
> Nope.

Shame.

PR #539, opened 2020-04-03
Use a helper function to avoid API-breaking change

piotrAMD:
https://reviews.llvm.org/D72467 changes the API of ShuffleVectorInst
by replacing:

Constant *getMask()

with:

ArrayRef<int> getShuffleMask().

Temporarily use a function with an unchanged signature:

getShuffleMask(SmallVectorImpl<int> &Result)

to keep the code working before and after the LLVM upgrade.

The intention is to use getShuffleMask() eventually.

### DIFF
--- a/lgc/patch/llpcPatchPeepholeOpt.cpp
+++ b/lgc/patch/llpcPatchPeepholeOpt.cpp
@@ -229,8 +229,25 @@ void PatchPeepholeOpt::visitBitCast(
         insertAfter(*pBitCastRhs, *pBitCastLhs);
 
         // Create our new shuffle instruction.
+
+        // TODO: Simplify by using pShuffleVector->getShuffleMask() (no arguments) when it becomes available.
+        SmallVector<int, 8> maskVals;
+        pShuffleVector->getShuffleMask(maskVals);
+
+        auto pInt32Type = Type::getInt32Ty(pShuffleVector->getContext());
+        SmallVector<Constant*, 8> masks;
+        for (unsigned i = 0; i < maskVals.size(); ++i) {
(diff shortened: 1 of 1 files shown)


lgc/patch/llpcPatchPeepholeOpt.cpp
```diff
@@ -229,8 +229,25 @@ void PatchPeepholeOpt::visitBitCast(
         insertAfter(*pBitCastRhs, *pBitCastLhs);
 
         // Create our new shuffle instruction.
+
+        // TODO: Simplify by using pShuffleVector->getShuffleMask() (no arguments) when it becomes available.
+        SmallVector<int, 8> MaskArray;
```
amdrexu, line 234: MaskArray -> maskVals

lgc/patch/llpcPatchPeepholeOpt.cpp
```diff
@@ -229,8 +229,25 @@ void PatchPeepholeOpt::visitBitCast(
...
+        pShuffleVector->getShuffleMask(MaskArray);
+
+        auto pInt32Type = Type::getInt32Ty(pShuffleVector->getContext());
+        SmallVector<Constant*, 8> Elts;
```
amdrexu, line 238: Elts -> masks

piotrAMD: This is only a temporary change, but I will make the suggested changes anyway.

amdrexu: Thanks for clarification :)

PR #550, opened 2020-04-08
switch to use MetroHash128

amingriyue:
align to pipeline binary cache, which will be easy to be unified later.

### DIFF
--- a/tool/dumper/vkgcPipelineDumper.cpp
+++ b/tool/dumper/vkgcPipelineDumper.cpp
@@ -121,7 +121,7 @@ void VKAPI_CALL IPipelineDumper::DumpSpirvBinary(
     const BinaryData*               pSpirvBin)  // [in] SPIR-V binary
 {
     MetroHash::Hash hash = {};
-    MetroHash64::Hash(reinterpret_cast<const uint8_t*>(pSpirvBin->pCode),
+    MetroHash128::Hash(reinterpret_cast<const uint8_t*>(pSpirvBin->pCode),
                       pSpirvBin->codeSize,
                       hash.bytes);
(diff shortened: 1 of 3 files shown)


amingriyue:
test it please

tool/dumper/vkgcPipelineDumper.cpp
```diff
@@ -955,7 +955,7 @@ MetroHash::Hash PipelineDumper::GenerateHashForComputePipeline(
 // Updates hash code context for vertex input state
 void PipelineDumper::UpdateHashForVertexInputState(
     const VkPipelineVertexInputStateCreateInfo* pVertexInput,  // [in] Vertex input state
-    MetroHash64*                                pHasher)       // [in,out] Haher to generate hash code
+    MetroHash128*                                pHasher)       // [in,out] Haher to generate hash code
```
Flakebi, line 958: The alignment in this file and in ` tool/dumper/vkgcPipelineDumper.h` should be updated :)

kuhar:
Can you describe the overall impact of this change? For example, does it affect pipeline dump names, break hash compatibility with PAL or change the in-memory shader caching?

JacobHeAMD:
retest this please

amingriyue:
> Can you describe the overall impact of this change? For example, does it affect pipeline dump names, break hash compatibility with PAL, make it compatible with the hashes in RGP, or change the in-memory shader caching?
> 
> An obvious alternative would be to move pipeline binary cache to use Metrohash64; why do you prefer to move the rest of the code to 128 instead?

amingriyue:
Thanks, this is also the answer of why I separate this change from my implementation and send to review individually, that I try to find.

I will try to make Hash128 only happen in shader cache.

PR #553, opened 2020-04-09
Cleanup compiler options

Flakebi:
- Use the same options for gcc and clang, the clang options were
  outdated and all options that we set for gcc are also supported by
  clang.
- Remove checking the os, we check the compiler name anyway.

@trenouf The compiler options are duplicated in the `lgc` and `llpc` subdirectories, I think it makes sense to have a top-level `CMakeLists.txt` and put it there as a function, like I did in `llpc/CMakeLists.txt`. I don’t really know what changes are necessary in other projects to introduce that, could you guide me or do that please?

@dstutt As I found out yesterday with Nicolai, clang is actually tested despite the comment :tada:. Would you like to keep the `-Werror` flags for these CI-builds? Then I would add an option to enable them.

### DIFF
--- a/lgc/CMakeLists.txt
+++ b/lgc/CMakeLists.txt
@@ -32,67 +32,71 @@ add_dependencies(LLVMlgc
 )
 
 ### Cached Project Options #############################################################################################
-option(LLPC_BUILD_RENOIR   "LLPC support for Renoir?"   ON)
+option(LLPC_BUILD_RENOIR  "LLPC support for Renoir?"    ON)
+option(LLPC_ENABLE_WERROR "Build LLPC with more errors" OFF)
 
 ### Compiler Options ###################################################################################################
 function(lgc_set_compiler_options PROJECT_NAME)
-    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+        if(LLPC_ENABLE_WERROR)
+            target_compile_options("${PROJECT_NAME}" PRIVATE
+                -Werror
+                -Wno-error=deprecated-declarations
+            )
+        endif()
+
         # SEE: https://gcc.gnu.org/onlinedocs/gcc-6.2.0/gcc/Option-Summary.html#Option-Summary
         # for a list of all options and documentation.
-        #target_compile_options(${PROJECT_NAME} PRIVATE option)
         target_compile_options(${PROJECT_NAME} PRIVATE
-            -Wno-unused-parameter
-            -Wno-type-limits
-            -Wno-switch
-            -Wno-parentheses
-            -Wno-maybe-uninitialized
-            -Wno-delete-non-virtual-dtor
-            -Wno-sign-compare
+            -fno-strict-aliasing
+            -fvisibility-inlines-hidden
+            -Wall
             -Wno-delete-incomplete
+            -Wno-delete-non-virtual-dtor
+            -Wno-invalid-offsetof
+            -Wno-missing-braces
+            -Wno-missing-field-initializers
+            -Wno-parentheses
+            -Wno-sign-compare
+            -Wno-switch
+            -Wno-type-limits
+            -Wno-unused-parameter
             -Wunused-variable
             -Werror=unused-variable
             -Wunused-function
             -Werror=unused-function
-            -Wno-pedantic
-            -Wno-extra
         )
 
         target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:
+            -fno-rtti
+            -fPIC
+            -std=c++14
             -Wno-unused
             -Wno-ignored-qualifiers
             -Wno-missing-field-initializers
             -Wno-invalid-offsetof           # offsetof within non-standard-layout type 'x' is undefined
         >)
 
-        target_compile_options(${PROJECT_NAME} PRIVATE -fno-strict-aliasing)
-
-        target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:
-            -std=c++14
-            -fno-rtti
-            -fPIC
-        >)
-    elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-        target_compile_options(${PROJECT_NAME} PRIVATE
-            -fvisibility-inlines-hidden
-            -fcolor-diagnostics
-            -Wall
-            -Werror
-            -Wno-invalid-offsetof
-            -Wno-missing-braces
-            -Wno-nested-anon-types
-            -Wno-gnu-anonymous-struct
-            -Wno-covered-switch-default
-            -Wno-missing-field-initializers
-            -Wno-extra-semi
-            -Wno-sign-compare
-            -Wno-gnu-anonymous-struct
-        )
-        target_compile_options(${PROJECT_NAME} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:
-            -std=c++14
-            -fno-rtti
-            -fPIC
-        >)
-    elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+            target_compile_options("${PROJECT_NAME}" PRIVATE
+                -Wno-extra
+                -Wno-maybe-uninitialized
+                -Wno-pedantic
+            )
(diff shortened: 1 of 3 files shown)


trenouf:
The options are duplicated between llpc and lgc only because I created lgc/CMakeLists.txt by copying llpc/CMakeLists.txt and editing it. :-) Really, they probably want a different set of options. Also we do not want lgc/CMakeLists.txt relying on the top-level CMakeLists.txt as lgc is configured as an llvm external project, and is intended to be (one day) buildable without the rest of llpc. It would be nice to reduce the options in each place.

JaxLinAMD:
retest this please

JaxLinAMD:
can one of the admins verify this patch?

trenouf:
Does this lose -Wall -Werror from clang?

Flakebi:
When I created the patch, I still thought that clang is not tested (like it always tells you when you build with clang), so I thought the clang options were just outdated and I used the gcc options for both, gcc and clang.

So yes, atm this patch removes `-Wall` and `-Werror` (gcc stays unchanged, it never had them).

trenouf:
I am using clang locally, and I would like to keep them. Although I take your point that -Werror causes problems for someone building on a compiler version that we don't try, so maybe that should be configured by something on the cmake command line.

trenouf:
Over on #570, I create a new lgc/tool/lgc/CMakeLists.txt, and it has a whole new copy of the compiler options setting stuff.

I just tried putting it in a function or macro on lgc/CMakeLists.txt then invoking it from lgc/tool/lgc/CMakeLists.txt (which is add_subdirectoried from lgc/CMakeLists.txt), but it said "unknown cmake command". Is there a way to define a function or macro such that it is visible from other add_subdirectoried cmake files?

dstutt:
> Over on #570, I create a new lgc/tool/lgc/CMakeLists.txt, and it has a whole new copy of the compiler options setting stuff.
> 
> I just tried putting it in a function or macro on lgc/CMakeLists.txt then invoking it from lgc/tool/lgc/CMakeLists.txt (which is add_subdirectoried from lgc/CMakeLists.txt), but it said "unknown cmake command". Is there a way to define a function or macro such that it is visible from other add_subdirectoried cmake files?

Defining a function should work as long as it comes before the add_subdir command.

trenouf:
> Defining a function should work as long as it comes before the add_subdir command.

D'oh! User error. Problem between keyboard and chair. (I had a function name mismatch.)

trenouf:
#570 now adjusted to use a function in lgc/CMakeLists.txt, called by lgc/tool/lgc/CMakeLists.txt.

nhaehnle:
Gah... combination of email threading fail and user error leading to a stale comment... sorry.

Flakebi:
I pushed a new version. Now `-DLLPC_ENABLE_WERROR=ON` can be used to add the `-Werror` flag to llpc and lgc builds, so this flag should be added to our jenkins builds.

amdvlk-admin:
### Test summary for commit 03650aa3a7ddf0d9645ff06bb5fab712e1bcbbc4

<li>PAL: dev b5e6dde2a0aa5dd2ce6d794e58d89ef715570f2d</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev f02738a5cf6c158c8c5b25a5048951f3059cabe2</li>
</ul></details>

<li>Failed: 0/955 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

kuhar:
Nice. Maybe we should also a clang configuration to the public CI system once clang support is reasonably stable?

Flakebi:
I think the result of using gcc or clang should be the same, especially since they should both use the same set of flags.
However, clang tends to build faster than gcc and faster CI is always nice :)
As a heads up, I *think* that building with clang also needs lld, at least it failed for me with the default linker, telling me I should run `ranlib` on some file.
`lld` can be used with `-DLLVM_USE_LINKER=lld -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld -DCMAKE_SHARED_LINKER_FLAGS=-fuse-ld=lld`.

kuhar:
> I think the result of using gcc or clang should be the same, especially since they should both use the same set of flags.
The biggest difference will be in the warnings they emit, which should be useful to know about.

Flakebi:
> > I think the result of using gcc or clang should be the same, especially since they should both use the same set of flags.
>
> The biggest difference will be in the warnings they emit, which should be useful to know about.

Right, I agree.
I think testing both, release and debug builds could also be worth it. I just got a build failure with `-Werror` in a release build because a variable is only used inside an `assert`, so it works fine in debug mode but counts as unused in release mode.

Flakebi:
Can someone review and merge this please? Maybe @trenouf?

lgc/CMakeLists.txt
```diff
@@ -32,67 +32,59 @@ add_dependencies(LLVMlgc
...
-            -Wno-switch
-            -Wno-parentheses
-            -Wno-maybe-uninitialized
+            -fno-strict-aliasing
+            -fvisibility-inlines-hidden
+            -Wall
+            -Wno-covered-switch-default
+            -Wno-delete-incomplete
             -Wno-delete-non-virtual-dtor
+            -Wno-error=deprecated-declarations
```
trenouf, line 56: Can we move -Wno-error=xxx (of which I think this is the only one) up into the "if(LLPC_ENABLE_WERROR)" clause above?

llpc/CMakeLists.txt
```diff
@@ -33,72 +33,65 @@ if(ICD_BUILD_LLPC)
...
-            -Wno-parentheses
-            -Wno-maybe-uninitialized
+        target_compile_options("${PROJECT_NAME}" PRIVATE
+            -fno-strict-aliasing
+            -fvisibility-inlines-hidden
+            -Wall
+            -Wno-covered-switch-default
+            -Wno-delete-incomplete
             -Wno-delete-non-virtual-dtor
+            -Wno-error=deprecated-declarations
```
trenouf, line 61: Same.

Flakebi:
Thanks Tim, I fixed your remarks. Jenkins tests succeeded on 2 nodes and failed on one, so I think this can be merged.

amdvlk-admin:
### Test summary for commit e5923e7b1e448c2b09bf26992e2988916a93c66b

<li>PAL: dev e2f8eb0ffe21751b300995e795965d37e8d6dd67</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev f02738a5cf6c158c8c5b25a5048951f3059cabe2</li>
</ul></details>

<li>Failed: 0/958 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

<li>Passed: 2199/2200 (100.0%)</li>
<li><details><summary>Failed: 1/2200 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.api.object_management.multithreaded_shared_resources.compute_pipeline
```
</p></details></li>


</ul></details>


</ul></details>

JaxLinAMD:
merge process is started, please don't force push.

lgc/CMakeLists.txt
```diff
@@ -32,67 +32,59 @@ add_dependencies(LLVMlgc
 )
 
 ### Cached Project Options #############################################################################################
-option(LLPC_BUILD_RENOIR   "LLPC support for Renoir?"   ON)
+option(LLPC_BUILD_RENOIR  "LLPC support for Renoir?"    ON)
```
JaxLinAMD, line 35: @Flakebi , why did you remove a space here? is there any coding standard about this? if so,  why not remove one space between "LLPC support for Renoir?" and "ON"?

Flakebi: I wanted to align it with the line below that I inserted, that’s all.
I guess there is no standard about the amount of spaces in there, it varies greatly. The only system behind it that I see is that lines that are near each other are aligned to each other.

JaxLinAMD: got it.not a problem then.

lgc/CMakeLists.txt
```diff
@@ -32,67 +32,71 @@ add_dependencies(LLVMlgc
...
+                -Wno-extra
+                -Wno-maybe-uninitialized
+                -Wno-pedantic
+            )
+            if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 8.4)
+                target_compile_options("${PROJECT_NAME}" PRIVATE
+                    -Wno-class-memaccess
+                )
+            endif()
+        elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
trenouf, line 91: Sorry, missed this in the original review, but this needs to be MATCHES rather than STREQUAL. Some versions of Clang are called "SomethingClang" rather than just "Clang". Same for the one in llpc/CMakeLists.txt. I'll do a new PR for this.

PR #543, opened 2020-04-06
Coding standards change

trenouf:

PR #563, opened 2020-04-10
CodingStandards: Separator line at top of class/func/method comment

trenouf:
Change-Id: I86cd75fef4e4a391039a172a5e128d0604772fbd

### DIFF
--- a/docs/CodingStandards.md
+++ b/docs/CodingStandards.md
@@ -132,7 +132,11 @@ or `return`.
 Despite the look of much LLVM code, the LLVM coding standard does actually recommend
 placing a comment at the top of a function/method. :-)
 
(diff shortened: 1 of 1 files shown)


trenouf:
This is in response to Rex's comment over on #543:

> Since you have a good opportunity to reformat the code, I also want to make those items clear:
> (1) A separate line "// =====" is always required for a class definition and any function implementation.
> (2) A separate line "// -----" could be added to separate member functions and data member in the same access section (public, protected, private) within a class.

This commit documents (1). I don't really see the need for (2); it's obvious which ones are methods and which ones are fields anyway from the m_ prefix. I don't know of anywhere in LLVM doing it (and LLVM doesn't even use the m_ prefix).

JaxLinAMD:
retest this please

PR #564, opened 2020-04-10
.clang-format: Don't merge out-of-line func/method into single line

trenouf:
Merging out-of-line func/method to a single line was an unintentional
change.

Change-Id: I79a7d0a8730997371260ad9c313afe48b036b453


JaxLinAMD:
retest this please

trenouf:
This passed CTS on one agent and spuriously crashed on the other. It has had a more recent successful test as part of #568.

PR #570, opened 2020-04-13
LGC lit test framework

trenouf:
(Only the top three commits are for this PR. Others are from earlier unmerged PRs.)

These three commits add an -emit-lgc option, an lgc command-line utility, and a lit test framework (with an initial test) for LGC.

### DIFF
--- a/lgc/builder/BuilderRecorder.cpp
+++ b/lgc/builder/BuilderRecorder.cpp
@@ -292,6 +292,8 @@ StringRef BuilderRecorder::getCallName(Opcode opcode) {
     return "subgroup.write.invocation";
   case SubgroupMbcnt:
     return "subgroup.mbcnt";
+  case Count:
+    break;
   }
   llvm_unreachable("Should never be called!");
   return "";
@@ -309,9 +311,9 @@ BuilderRecorderMetadataKinds::BuilderRecorderMetadataKinds(llvm::LLVMContext &co
 //
 // @param builderContext : Builder context
 // @param pipeline : PipelineState, or nullptr for shader compile
-BuilderRecorder::BuilderRecorder(LgcContext *builderContext, Pipeline *pipeline)
+BuilderRecorder::BuilderRecorder(LgcContext *builderContext, Pipeline *pipeline, bool omitOpcodes)
     : Builder(builderContext), BuilderRecorderMetadataKinds(builderContext->getContext()),
-      m_pipelineState(reinterpret_cast<PipelineState *>(pipeline)) {
+      m_pipelineState(reinterpret_cast<PipelineState *>(pipeline)), m_omitOpcodes(omitOpcodes) {
   m_isBuilderRecorder = true;
 }
 
@@ -1837,8 +1839,15 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
     auto funcTy = FunctionType::get(resultTy, {}, true);
     func = Function::Create(funcTy, GlobalValue::ExternalLinkage, mangledName, module);
 
-    MDNode *const funcMeta = MDNode::get(getContext(), ConstantAsMetadata::get(getInt32(opcode)));
-    func->setMetadata(opcodeMetaKindId, funcMeta);
+    // Add opcode metadata to the function, so that BuilderReplayer does not need to do a string comparison.
+    // We do not add that metadata if doing -emit-lgc, so that a test constructed with -emit-lgc will rely
+    // on the more stable lgc.create.* name rather than the less stable opcode.
+    if (!m_omitOpcodes) {
+      MDNode *const funcMeta = MDNode::get(getContext(), ConstantAsMetadata::get(getInt32(opcode)));
+      func->setMetadata(opcodeMetaKindId, funcMeta);
+    }
+
+    // Add requested attributes, plus nounwind.
     func->addFnAttr(Attribute::NoUnwind);
     for (auto attrib : attribs)
       func->addFnAttr(attrib);
@@ -1848,4 +1857,28 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
   auto call = CreateCall(func, args, instName);
 
   return call;
+}
+
+// =====================================================================================================================
+// Get the recorded call opcode from the function name. Asserts if not found.
+// This does not have to be particularly efficient, as it is only used with the lgc command-line utility.
+//
+// @param name : Name of function declaration
+// @return : Opcode
+BuilderRecorder::Opcode BuilderRecorder::getOpcodeFromName(StringRef name) {
+  assert(name.startswith(BuilderCallPrefix));
+  name = name.slice(strlen(BuilderCallPrefix), StringRef::npos);
+  unsigned bestOpcode = 0;
+  unsigned bestLength = 0;
+  for (unsigned opcode = 0; opcode != Opcode::Count; ++opcode) {
+    StringRef opcodeName = getCallName(static_cast<Opcode>(opcode));
+    if (name.startswith(opcodeName)) {
+      if (opcodeName.size() > bestLength) {
+        bestLength = opcodeName.size();
+        bestOpcode = opcode;
+      }
+    }
+  }
+  assert(bestLength != 0 && "No matching lgc.create.* name found!");
+  return static_cast<Opcode>(bestOpcode);
 }
--- a/lgc/builder/BuilderRecorder.cpp
+++ b/lgc/builder/BuilderRecorder.cpp
@@ -1867,7 +1867,7 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
 // @return : Opcode
 BuilderRecorder::Opcode BuilderRecorder::getOpcodeFromName(StringRef name) {
   assert(name.startswith(BuilderCallPrefix));
-  name = name.slice(strlen(BuilderCallPrefix), StringRef::npos);
+  name = name.drop_front(strlen(BuilderCallPrefix));
   unsigned bestOpcode = 0;
   unsigned bestLength = 0;
   for (unsigned opcode = 0; opcode != Opcode::Count; ++opcode) {
--- a/lgc/test/CMakeLists.txt
+++ b/lgc/test/CMakeLists.txt
@@ -1,0 +1,59 @@
+##
+ #######################################################################################################################
+ #
+ #  Copyright (c) 2020 Advanced Micro Devices, Inc. All Rights Reserved.
+ #
+ #  Permission is hereby granted, free of charge, to any person obtaining a copy
+ #  of this software and associated documentation files (the "Software"), to deal
+ #  in the Software without restriction, including without limitation the rights
+ #  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ #  copies of the Software, and to permit persons to whom the Software is
+ #  furnished to do so, subject to the following conditions:
+ #
+ #  The above copyright notice and this permission notice shall be included in all
+ #  copies or substantial portions of the Software.
+ #
+ #  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ #  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ #  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ #  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ #  SOFTWARE.
+ #
+ #######################################################################################################################
+
+configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.py.in
+  ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py
+  MAIN_CONFIG
+  ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg.py
+  )
+
+set(LGC_TEST_DEPENDS lgc)
+add_custom_target(lgc-test-depends DEPENDS ${LGC_TEST_DEPENDS})
+set_target_properties(lgc-test-depends PROPERTIES FOLDER "Tests")
+
+# required by lit.site.cfg.py.in
+set(LGC_TEST_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
+# required by configure_lit_site_cfg
+set(LLVM_LIT_OUTPUT_DIR ${LLVM_TOOLS_BINARY_DIR})
+configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.py.in
+  ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py
+  MAIN_CONFIG
+    ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg.py
+)
+
+add_lit_testsuite(check-lgc "Running the LGC regression tests"
+  ${CMAKE_CURRENT_BINARY_DIR}
+  ${exclude_from_check_all}
+  DEPENDS ${LGC_TEST_DEPENDS}
+)
+set_target_properties(check-lgc PROPERTIES FOLDER "Tests")
+
+add_lit_testsuites(LLVM ${CMAKE_CURRENT_SOURCE_DIR}
+  ${exclude_from_check_all}
+  DEPENDS ${LLVM_TEST_DEPENDS}
+)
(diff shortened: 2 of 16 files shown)


lgc/tool/lgc/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+set(LLVM_LINK_COMPONENTS
+    lgc
+    Core
+    AMDGPUAsmParser
+    AMDGPUCodeGen
+    AMDGPUDisassembler
+)
+
+add_llvm_tool(lgc
+    lgc.cpp
```
amdrexu, line 34: I am not sure if in the future, we will have a true lgc.cpp in LGC library. This is why we use amdllpc as the standalone tool name rather than llpc. Absolutely, lgc as the name is apparently elegant.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+      "\n"
+      "The options section below lists all options available, including all LLVM ones.\n"
+      "The most relevant ones for the lgc utility are:\n"
+      "\n"
+      "  -extract=<index>         - Extract a single input module from multi-module\n"
+      "                             input. Index is 1-based\n"
+      "  -o=<filename>            - Output filename ('-' for stdout); defaults to the input\n"
+      "                             filename with the directory removed and the extension\n"
+      "                             replaced to match the output type\n"
+      "  -mcpu=<cpu-name>         - Target a specific GPU e.g. 'gfx900'\n"
```
amdrexu, line 108: I understand this is to keep consistent with llc. Is it possible to rename it to other more clear names, such as -gfxip, -gpu.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+      "  -filetype=<value>        - Choose an output file type:\n"
+      "   =asm                    -   Emit an assembly ('.s') file (default)\n"
```
amdrexu, line 110: ".s" file extension is not clear to me. Is there any reason? I prefer ".asm"

lgc/tool/lgc/lgc.cpp
trenouf: I prefer to keep it as -mcpu as that option already exists in the llvm code that is linked in to lgc. As well as being consistent with llvm tools.

lgc/tool/lgc/lgc.cpp
trenouf: .s is the standard extension for assembly source in unix and thus in llvm.

trenouf:
Rebased.

amdvlk-admin:
### Test summary for commit bc2ba9efe084342d0059178ea3c26bb46e02bdd9

<li>PAL: dev bb7398b1c7509a30ba4f7d947b5ad5549a4d58db</li>
<li>SPVGEN: dev dc2b146de03ef54f188172d8fa2f9be708cbfaca</li>
<li>XGL: dev ffd39443fec0ce821b7702173eba7184855c7b5e</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 7b14353794b408b8d7fd87ae9a9f4f553e6d4782</li>
</ul></details>

<li>Failed: 0/955 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 38418/38419 (100.0%)</li>
<li><details><summary>Failed: 1/38419 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.instruction.graphics.16bit_storage.uniform_float_16_to_64.uniform_buffer_block_vector_float_const_idx_6_geom
```
</p></details></li>


</ul></details>

</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit a80b5970b6d7316030673c886d28a8a31dab9cc3

<li>PAL: dev bb7398b1c7509a30ba4f7d947b5ad5549a4d58db</li>
<li>SPVGEN: dev dc2b146de03ef54f188172d8fa2f9be708cbfaca</li>
<li>XGL: dev ffd39443fec0ce821b7702173eba7184855c7b5e</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 7b14353794b408b8d7fd87ae9a9f4f553e6d4782</li>
</ul></details>

<li>Failed: 0/955 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 38367/38368 (100.0%)</li>
<li><details><summary>Failed: 1/38368 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.instruction.graphics.16bit_storage.uniform_float_64_to_16.uniform_scalar_float_unspecified_rnd_mode_tesse
```
</p></details></li>


</ul></details>

</ul></details>


</ul></details>

trenouf:
Ping for review.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+        if (remaining[notSpacePos] == '!')
+          hadMetadata = true;
+        else if (hadMetadata && remaining.slice(notSpacePos, StringRef::npos).startswith("target")) {
+          // End the current split module and go on to the next one.
+          separatedAsms.back() = separatedAsms.back().slice(0, remaining.data() - separatedAsms.back().data());
+          separatedAsms.push_back(remaining);
+          hadMetadata = false;
+        }
+      }
+      auto nlPos = remaining.find_first_of('\n');
```
amdrexu, line 165: nlPos -> newLinePos

lgc/test/lgc-tool.lgc
```diff
@@ -0,0 +1,39 @@
+; Simple test that checks that the lgc command works with multiple modules in
```
amdrexu, line 1: This could be put in this directory as a a tutorial sample (sample.lgc). Do you have any plan on the naming schema of those tests (or maybe hierarchy on different GPUs)? This will quickly help the developers to determine if there are tests existing. The purposes are two: (1) Avoid add new tests and just extend existing ones (2) check if a test is missing and should be added as a new one. I don't like the naming in LLVM tests/CodeGen/AMDGPU. Besides the intrinsic names, it is not easy to quickly find a test that serves certain functionalities.

lgc/builder/BuilderRecorder.cpp
```diff
@@ -1849,3 +1858,27 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
...
+
+// =====================================================================================================================
+// Get the recorded call opcode from the function name. Asserts if not found.
+// This does not have to be particularly efficient, as it is only used with the lgc command-line utility.
+//
+// @param name : Name of function declaration
+// @return : Opcode
+BuilderRecorder::Opcode BuilderRecorder::getOpcodeFromName(StringRef name) {
+  assert(name.startswith(BuilderCallPrefix));
+  name = name.slice(strlen(BuilderCallPrefix), StringRef::npos);
```
kuhar, line 1870: nit: this is the same as `name.drop_front(strlen(BuilderCallPrefix))`, no?

lgc/test/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+ #  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ #  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ #  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ #  SOFTWARE.
+ #
+ #######################################################################################################################
+
+cmake_minimum_required(VERSION 3.3)
```
kuhar, line 26: LLVM requires cmake >= 3.4.3. Does it make sense to specify anything lower here?

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include "llvm/IR/Verifier.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/SourceMgr.h"
+
+using namespace lgc;
+using namespace llvm;
+
+// Input sources
+static cl::list<std::string> InFiles(cl::Positional, cl::OneOrMore, cl::ValueRequired,
```
kuhar, line 45: nit: Could all of the statics be in a anonymous namespace instead?

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+// =====================================================================================================================
+// Checks whether the output data is actually ISA assembler text
+//
+// @param data : Input data to check
+static bool isIsaText(StringRef data) {
+  // This is called by the lgc standalone tool to help distinguish between its three output types of ELF binary,
+  // LLVM IR assembler and ISA assembler. Here we use the fact that ISA assembler is the only one that starts
+  // with a tab character.
+  return data.startswith("\t");
```
kuhar, line 81: Is there anything less brittle we could easily check?
I can imagine passing an LLVM textual IR as an input file that happens to start with a tab too.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+// =====================================================================================================================
+// Main code of LGC standalone tool
+//
+// @param argc : Count of command-line arguments
+// @param argv : Command-line arguments
+int main(int argc, char **argv) {
+  const char *progName = sys::path::filename(argv[0]).data();
+  LLVMContext context;
+  LgcContext::initialize();
+  cl::ParseCommandLineOptions(argc, argv, "lgc: command-line tool for LGC, the LLPC middle-end compiler\n"
```
kuhar, line 92: You can use a raw string literal here to make this long string a bit more pretty.
But won't this help get printed out automatically based on the cl options? It seems very repetitive to specify these options 3 times: as `llvm::cl::opt`, in comments above, and in the help text again.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+      "  -mcpu=<cpu-name>         - Target a specific GPU e.g. 'gfx900'\n"
+      "  -filetype=<value>        - Choose an output file type:\n"
+      "   =asm                    -   Emit an assembly ('.s') file (default)\n"
+      "   =obj                    -   Emit an ELF ('.elf') file\n"
+      "  -emit-llvm               - Emit LLVM IR assembly before running back-end\n"
+      "  -pal-abi-version=<uint>  - PAL pipeline ABI version to compile for (default\n"
+      "                             latest known)\n"
+      "  -verify-ir               - Verify IR after each pass\n"
+      );
+
```
kuhar, line 117: Would it make sense to enable `llvm::PrettyStackTraceProgram` here too? I expect the lgc tool to be used mostly for debugging/development, so increasing debuggability seems desirable to me.

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+  // with a tab character.
+  return data.startswith("\t");
+}
+// =====================================================================================================================
+// Main code of LGC standalone tool
+//
+// @param argc : Count of command-line arguments
+// @param argv : Command-line arguments
+int main(int argc, char **argv) {
+  const char *progName = sys::path::filename(argv[0]).data();
```
kuhar, line 89: nit: const char *const progName

lgc/tool/lgc/lgc.cpp
```diff
@@ -0,0 +1,259 @@
...
+        continue;
+      }
+
+      SmallString<64> outFileName(OutFileName);
+      if (outFileName.empty()) {
+        // Determine the output filename by taking the input filename, removing the directory,
+        // removing the extension, and adding a default extension that depends on the output contents.
+        const char *ext = ".s";
+        if (isElfBinary(outBuffer)) {
+          ext = ".elf";
```
kuhar, line 234: Can we have these extension names as top-level `llvm::StringLiteral` constants?

lgc/builder/BuilderRecorder.cpp
```diff
@@ -1849,3 +1858,27 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
 
   return call;
 }
+
+// =====================================================================================================================
+// Get the recorded call opcode from the function name. Asserts if not found.
+// This does not have to be particularly efficient, as it is only used with the lgc command-line utility.
+//
+// @param name : Name of function declaration
+// @return : Opcode
```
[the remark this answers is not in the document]
trenouf: Good point; we should add something to the standards doc. However I think we should only ask for it in new code; it's not worth changing all existing code to add it.

lgc/tool/lgc/lgc.cpp
trenouf: I prefer nlPos as NL is a well-known abbreviation for newline.

lgc/test/lgc-tool.lgc
trenouf: This isn't just a sample; it serves a useful purpose for checking that the tool works, especially its -extract option.

I take your point about naming of tests, but I don't really know the answer yet. Let's leave it flat for the moment, and then move tests around later when we have a better idea of what kinds of test we have.

lgc/test/CMakeLists.txt
trenouf: No. :-) I guess it makes no sense to specify anything in lgc/CMakeLists.txt and its subdirs now it is an llvm external project.

lgc/tool/lgc/lgc.cpp
trenouf: Not sure if there is something better, but note that this check is used for output, not input, so it's a bit more canonical.

lgc/tool/lgc/lgc.cpp
trenouf: lgc -help just prints all of the options, so the ones that are particularly relevant to lgc get lost. That's why I added this bit here. But I think I saw some other tool have a separate section for its own options at the top the other day, so maybe there is a generic mechanism for doing that that I haven't found yet.

lgc/tool/lgc/lgc.cpp
trenouf: We don't generally have a style of consting local variables (although a couple of places written by Neil do it), and I don't think we want to start doing it.

lgc/tool/lgc/lgc.cpp
nhaehnle: Interesting, I never felt that that option helps debuggability at all. All it does is make llvm-lit run *a lot* slower when processing many files with crashes... not that it really matters either way.

lgc/tool/lgc/lgc.cpp
nhaehnle: I find the practice of putting string constants out-of-line makes code a lot harder to read, especially when they're perfectly well-known strings like ".s".

Though one piece of bike-shedding I'd have here is that if we follow the practice of calling assembly files ".s", why aren't we calling ELF files ".o"? ;)

trenouf:
> 1. What's the purpose of having multiple modules in a single file? It seems like it'd be preferable to just consume standard .ll files.

Two things:

1a. I would like to be able to put multiple tiny tests in the same file. In the llc world, you can just have multiple functions in the same file. That doesn't work for lgc if what you're trying to test is something in the metadata, or you want all the tests to be fragment shaders.

1b. The multiple modules combined with the -extract option is because I would like to be able to have a single file that runs a test that separately compiles a VS and an FS and then links them.

Of course having multiple modules in a file is backward-compatible with having a standard single .ll file. And, with a standard .ll file, you can bugpoint it.

> 2. Having multiple tools could get confusing. Is the idea to extend `lgc` to also consume .pipe files?

No; my plan is that lgc is specifically for LGC, the middle-end. amdllpc continues to be the tool for the whole of LLPC.

lgc/tool/lgc/lgc.cpp
trenouf: That would be nice, but I think we have to follow the driver's practice of calling a pal pipeline elf ".elf".

Maybe it would be nice to use ".o" for an unlinked compiled shader, before linking, but the problem with that is that the only way of telling the difference is to parse the elf and find some bit of pal metadata (that we haven't defined yet). Both kinds of elf are "relocatable" in elf standard terms.

nhaehnle: Yeah, it's not a big deal.

amdrexu: Because PAL has an internal ELF analyzer that accepts ".elf" as the inputs. It is better to keep the output file extension consistent with this.

lgc/tool/lgc/lgc.cpp
trenouf: I figured out how to do this properly.

trenouf:
V8: Used option categories, and other review comments

lgc/tool/lgc/lgc.cpp
kuhar: I found it useful when browsing logs from other machines (think CI) where I cannot quickly rerun a failing test and attach to it.

lgc/tool/lgc/lgc.cpp
kuhar: OK

lgc/tool/lgc/lgc.cpp
kuhar: Ack, fine for me.

PR #581, opened 2020-04-16
Proposed LGC interface for shader compilation and linking

trenouf:
The top two commits define an LGC interface for shader compilation generating an unlinked ELF, and ELF linking.

For now this is only for discussion.

### DIFF
--- a/lgc/interface/lgc/ElfLinker.h
+++ b/lgc/interface/lgc/ElfLinker.h
@@ -1,0 +1,95 @@
+/*
+ ***********************************************************************************************************************
+ *
+ *  Copyright (c) 2020 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in all
+ *  copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ *
+ **********************************************************************************************************************/
+/**
+ ***********************************************************************************************************************
+ * @file  ElfLinker.h
+ * @brief LLPC header file: LGC interface for linking unlinked half-pipeline ELFs into pipeline ELF
+ ***********************************************************************************************************************
+ */
+#pragma once
+
+#include "lgc/CommonDefs.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace llvm {
+class raw_pwrite_stream;
+} // namespace llvm
+
+namespace lgc {
+
+// =====================================================================================================================
+// The public API of the LGC interface for ELF linking.
+// The ElfLinker object is created by calling Pipeline::getElfLinker(). The ElfLinker internally refers back to
+// its Pipeline, and thus uses pipeline state for glue code generation, adding to PAL metadata, and resolving
+// relocs.
+class ElfLinker {
+public:
+  virtual ~ElfLinker() {}
+
+  // Get information on the glue code that will be needed for the link. It is an implementation detail how many
+  // chunks of glue there might be and what they are for, but, for information, they will be some subset of:
+  // - A CS prolog
+  // - A VS prolog ("fetch shader")
+  // - A vertex-processing epilog ("parameter export shader")
+  // - An FS epilog ("color export shader")
+  //
+  // Returns an array (possibly 0 length) with an entry for each chunk of glue code, where an entry
+  // is a StringRef that the client can hash for its cache lookup. If it gets a cache hit, it should provide the
+  // found blob to ElfLinker::addGlue. If it does not get a cache hit, the client can call ElfLinker::compileGlue to
+  // retrieve the compiled glue code to store in the cache.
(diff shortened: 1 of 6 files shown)


amdrexu:
I might need an interface that could extract GPU codes from a shader module ELF (just shader module, not pipeline) compiled by back-end and inject it to a patcher (independent of LLPC/LGC) that will insert additional "harmless" data dump instructions (buffer stores) and reconstruct this shader module ELF. Extra processing might be the modifications of jump label symbols (BB_X and its offsets), certain PAL ABI metadata, and shader registers (such as VGPR and SGPR usage).

This could be done within LGC as a patch pass. Or be done when front-end get the compilation results and modify it. I guess the latter is not elegant and need builder to provide additional interfaces.

lgc/interface/lgc/ElfLinker.h
```diff
@@ -0,0 +1,84 @@
...
+
+  // Compile a particular chunk of glue code and retrieve its blob. The returned blob remains valid until the first
+  // of calling link() or the ElfLinker's parent Pipeline being destroyed. It is optional to call this; any chunk
+  // of glue code that has not had one of addGlue() or compileGlue() done by the time link() is called will be
+  // internally compiled. The client only needs to call this if it wants to cache the glue code's blob.
+  //
+  // @param glueIndex : Index into the array that was returned by getGlueInfo()
+  // @return : The blob. A zero-length blob indicates that a recoverable error occurred, and link() will also return
+  //           and empty ELF blob.
+  virtual llvm::StringRef compileGlue(unsigned glueIndex) = 0;
```
nhaehnle, line 74: StringRef isn't really the right type for this, is it?

lgc/interface/lgc/ElfLinker.h
```diff
@@ -0,0 +1,84 @@
...
+  // - A CS prolog
+  // - A VS prolog ("fetch shader")
+  // - A vertex-processing epilog ("parameter export shader")
+  // - An FS epilog ("color export shader")
+  //
+  // Returns an array (possibly 0 length) with an entry for each chunk of glue code, where an entry
+  // is a StringRef that the client can hash for its cache lookup. If it gets a cache hit, it should provide the
+  // found blob to ElfLinker::addGlue. If it does not get a cache hit, the client can call ElfLinker::compileGlue to
+  // retrieve the compiled glue code to store in the cache.
+  virtual llvm::ArrayRef<llvm::StringRef> getGlueInfo() = 0;
```
nhaehnle, line 57: Should this just be a hash then instead of a StringRef?

lgc/interface/lgc/Pipeline.h
```diff
@@ -581,17 +582,32 @@ class Pipeline {
...
-  // Like other Builder methods, on error, this calls report_fatal_error, which you can catch by setting
-  // a diagnostic handler with LLVMContext::setDiagnosticHandler.
+  //
+  // Three ways this can exit:
+  // 1. Return true for success.
+  // 2. Return false for recoverable failure because irLink asked for an "unlinked" half-pipeline and there is some
+  //    reason why the module cannot be compiled in that way. The client typically then does a whole-pipeline
+  //    compilation instead. (This cannot happen if the call to irLink did not ask for an unlinked half-pipeline.)
+  // 3. Other failures cause exit by report_fatal_error, which the client can catch by setting a diagnostic handler
+  //    with LLVMContext::setDiagnosticHandler.
```
nhaehnle, line 595: Other failures shouldn't really happen if the input is valid. Perhaps out-of-memory conditions, but can those really be caught with setDiagnosticHandler?

If false means "unlinked" didn't work, perhaps a Result enum would be clearer here.

lgc/state/Compiler.cpp
```diff
@@ -128,17 +128,31 @@ Module *PipelineState::irLink(ArrayRef<std::pair<Module *, ShaderStage>> modules
...
-// Like other Builder methods, on error, this calls report_fatal_error, which you can catch by setting
-// a diagnostic handler with LLVMContext::setDiagnosticHandler.
+//
+// Three ways this can exit:
+// 1. Return true for success.
+// 2. Return false for recoverable failure because irLink asked for an "unlinked" half-pipeline and there is some
+//    reason why the module cannot be compiled in that way. The client typically then does a whole-pipeline
+//    compilation instead.
+// 3. Other failures cause exit by report_fatal_error, which the client can catch by setting a diagnostic handler
+//    with LLVMContext::setDiagnosticHandler.
```
nhaehnle, line 141: Similar concern here about "other failures".

lgc/interface/lgc/ElfLinker.h
trenouf: It seems to be the done thing. It is common to pass around a MemoryBufferRef, which is a StringRef with an attached name for error reporting etc. So the use of StringRef to represent a read-only binary blob whose memory you don't own does seem right.

lgc/interface/lgc/ElfLinker.h
trenouf: I did wonder about that, but that ties the front-end (and/or driver) into a particular size of hash that might disagree with other stuff it does. So I thought it would be better to let the front-end do the hashing in the same way it does other hashing.

lgc/interface/lgc/Pipeline.h
trenouf: Yes, other failures tend to be because input is invalid, which llpc does already catch. I guess when they're most useful (and not caught, but just reported to stderr) is in a standalone tool.

lgc/interface/lgc/ElfLinker.h
nhaehnle: Ah I see, you're right.

lgc/interface/lgc/ElfLinker.h
nhaehnle: Fair enough.

lgc/interface/lgc/Pipeline.h
trenouf: I don't want to do a Result enum as it seems over-the-top for the two conditions we can return here ("success" or "you need to do full pipeline compile").

LLVM generally does not provide a recoverable error mechanism from running a pass, so we can just assume that any error that wasn't supposed to happen (e.g. a pass thinks that something is invalid) will cause a report_fatal_error, or an assert, or setting your computer on fire.

nhaehnle: Can we just reword the comment then? "Other failures" sounds too much like the client should *expect* them and *must* handle them, when really, they're internal compiler errors that should never happen.

So I wouldn't even list them as a 3rd item here, just maybe a note that internal compiler errors are reported as assertions and/or via llvm's report_fatal_error.

trenouf:
V2: Added early check for shader compilability. Added error message
        getting. link method now takes raw_pwrite_stream. Changed
        Pipeline::getElfLinker to Pipeline::createElfLinker, with caller
        responsible for freeing.
    V3: Addressed review comment.

trenouf:
V4: Fixed clang-format.

trenouf:
Ping for approval, as I need to merge this before #673.

trenouf:
Rebased

trenouf:
V5: Fixed clang-format again.
    V6: Fixed lgc.cpp compile failure.

amdvlk-admin:
### Test summary for commit 8299e876538e6e01bc892d542cc2b61cc71b7e3b

<li>LLVM-PROJECT: amd-gfx-gpuopen-dev c57487b76215ee74b1038039f47b210a28cad65e</li>
</ul></details>

<li>Failed: 0/960 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

PR #584, opened 2020-04-17
Make use of IRBuilder::CreateZExtOrTrunc

jayfoad:


JaxLinAMD:
merge process is started, please don't force-push

PR #591, opened 2020-04-18
Put option defaults into LgcContext so driver does not need to set them

trenouf:
Before this commit, both the driver (compiler_solution_llpc.cpp) and
amdllpc.cpp needed to set the same set of LLVM options. With this
commit:

* The scalarizer-related options are LLPC options, so I changed the
option defaults to what the code above currently sets.
* The other (unconditionally-set) ones now have default values set in
LGCContext.
* I removed the amdllpc option-setting code as it is no longer needed.
The driver code for unconditional options can be removed too.
* That will leave the options set in the driver conditionally on the
app profile. They should be removed in the future by them becoming
tuning options.

A further benefit of this is that the lgc utility does not need extra
command line options to behave in a way that is consistent with running
LGC as part of amdllpc.

Change-Id: I4353af365dd239cba08959b6d1e4c606d14431e3

### DIFF
--- a/lgc/state/LgcContext.cpp
+++ b/lgc/state/LgcContext.cpp
@@ -40,6 +40,7 @@
 #include "llvm/CodeGen/CommandFlags.h"
 #include "llvm/IR/IRPrintingPasses.h"
 #include "llvm/InitializePasses.h"
+#include "llvm/Support/CodeGen.h"
 #include "llvm/Support/TargetRegistry.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Target/TargetOptions.h"
@@ -71,6 +72,25 @@ static cl::opt<bool> EmitLgc("emit-lgc", cl::desc("Emit LLVM assembly suitable f
 
 // -show-encoding: show the instruction encoding when emitting assembler. This mirrors llvm-mc behaviour
 static cl::opt<bool> ShowEncoding("show-encoding", cl::desc("Show instruction encodings"), cl::init(false));
+
+// =====================================================================================================================
+// Set default for a command-line option, but only if command-line processing has not happened yet, or did not see
+// an occurrence of this option.
+//
+// @param name : Option name
+// @param value : Default option value
+static void setOptionDefault(const char *name, StringRef value) {
+  auto optIterator = cl::getRegisteredOptions().find(name);
+  assert(optIterator != cl::getRegisteredOptions().end() && "Failed to find option to set default");
+  cl::Option *opt = optIterator->second;
+  if (opt->getNumOccurrences())
+    return;
+  // Setting MultiArg means that addOccurrence will not increment the option's occurrence count, so the user
+  // can still specify it to override our default here.
+  bool setFailed = opt->addOccurrence(0, opt->ArgStr, value, /*MultiArg=*/true);
+  assert(!setFailed && "Failed to set default for option");
+  ((void)setFailed);
+}
 
 // =====================================================================================================================
 // Initialize the middle-end. This must be called before the first LgcContext::Create, although you are
@@ -110,6 +130,16 @@ void LgcContext::initialize() {
   initializeStatePasses(passRegistry);
   initializeBuilderReplayerPass(passRegistry);
   initializePatchPasses(passRegistry);
+
+  // Initialize some command-line option defaults.
+  setOptionDefault("filetype", "obj");
+  setOptionDefault("unroll-max-percent-threshold-boost", "1000");
+  setOptionDefault("pragma-unroll-threshold", "1000");
+  setOptionDefault("unroll-allow-partial", "1");
+  setOptionDefault("simplifycfg-sink-common", "0");
+  setOptionDefault("amdgpu-vgpr-index-mode", "1"); // force VGPR indexing on GFX8
+  setOptionDefault("amdgpu-atomic-optimizations", "1");
+  setOptionDefault("use-gpu-divergence-analysis", "1");
 }
 
 // =====================================================================================================================
--- a/llpc/context/llpcPipelineContext.cpp
+++ b/llpc/context/llpcPipelineContext.cpp
@@ -86,6 +86,15 @@ static cl::opt<bool> EnableSiScheduler("enable-si-scheduler", cl::desc("Enable t
 // -subgroup-size: sub-group size exposed via Vulkan API.
 static cl::opt<int> SubgroupSize("subgroup-size", cl::desc("Sub-group size exposed via Vulkan API"), cl::init(64));
 
+// -enable-shadow-desc: enable shadow descriptor table
+static cl::opt<bool> EnableShadowDescriptorTable("enable-shadow-desc", cl::desc("Enable shadow descriptor table"));
+
+// -shadow-desc-table-ptr-high: high part of VA for shadow descriptor table pointer.
+// Default of 2 is for use by standalone amdllpc.
+static cl::opt<unsigned> ShadowDescTablePtrHigh("shadow-desc-table-ptr-high",
+                                                cl::desc("High part of VA for shadow descriptor table pointer"),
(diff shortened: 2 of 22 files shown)


trenouf:
In the Jenkins run, it looks like one agent passed CTS and the others did not run.

amdrexu:
I think we need to discuss the stability of CI in the meeting. Recently, spurious failures lead to lots of unnecessary job re-submissions. Normal verification is disturbed.

llpc/context/llpcPipelineContext.cpp
```diff
@@ -70,14 +70,14 @@ static cl::opt<unsigned> WavesPerEu("waves-per-eu", cl::desc("Maximum number of
...
-                                      cl::desc("Enable the optimization for load scalarizer."), cl::init(false));
+                                      cl::desc("Enable the optimization for load scalarizer."), cl::init(true));
 
 // The max threshold of load scalarizer.
 static const unsigned MaxScalarThreshold = 0xFFFFFFFF;
 
 // -scalar-threshold: Set the vector size threshold for load scalarizer.
 static cl::opt<unsigned> ScalarThreshold("scalar-threshold", cl::desc("The threshold for load scalarizer"),
-                                         cl::init(MaxScalarThreshold));
+                                         cl::init(3));
```
amdrexu, line 80: MaxScalarThreshold definition could be removed.

llpc/tool/amdllpc.cpp
```diff
@@ -373,139 +373,72 @@ static Result init(int argc, char *argv[], ICompiler **ppCompiler) {
...
-            token = std::strtok(nullptr, ".");
-          }
-
-          ParsedGfxIp.major = tokens[0] ? std::strtoul(tokens[0], nullptr, 10) : 0;
-          ParsedGfxIp.minor = tokens[1] ? std::strtoul(tokens[1], nullptr, 10) : 0;
-          ParsedGfxIp.stepping = tokens[2] ? std::strtoul(tokens[2], nullptr, 10) : 0;
-
-          delete[] gfxIp;
+    // Before we get to LLVM command-line option parsing, we need to find the -gfxip option value.
+    for (int i = 1; i != argc; ++i) {
```
amdrexu, line 377: int -> unsigned, better to use unsigned unless we want to a signed comparison.

llpc/tool/amdllpc.cpp
```diff
@@ -373,139 +373,72 @@ static Result init(int argc, char *argv[], ICompiler **ppCompiler) {
...
-
-      if (!envString) {
-        envString = getenv("HOME");
-        if (envString) {
-          snprintf(shaderCacheFileRootDir, sizeof(shaderCacheFileRootDir), "%s/.cache", envString);
-          envString = &shaderCacheFileRootDir[0];
-        }
-      }
-#endif
+    constexpr unsigned maxFilePathLen = 512;
```
amdrexu, line 413: maxFilePathLen -> MaxFilePathLen

llpc/tool/amdllpc.cpp
```diff
@@ -373,139 +373,72 @@ static Result init(int argc, char *argv[], ICompiler **ppCompiler) {
...
+    //   1. Find AMD_SHADER_DISK_CACHE_PATH to keep backward compatibility.
+    const char *envString = getenv("AMD_SHADER_DISK_CACHE_PATH");
 
 #ifdef WIN_OS
-      //   2. Find LOCALAPPDATA.
-      if (envString == nullptr) {
-        envString = getenv("LOCALAPPDATA");
-      }
+    //   2. Find LOCALAPPDATA.
+    if (envString == nullptr) {
```
amdrexu, line 409: Since you touch this part of codes, maybe you can fix some minor coding issues in original codes.
->
```
if (!envString)
  envString = getenv("LOCALAPPDATA");
```

llpc/tool/amdllpc.cpp
trenouf: I tried that originally, but that gives a warning on one compiler (an error because we have warnings-as-errors) because it is compared with argc, which is an int.

amdrexu: I see. I didn't notice argc was a signed one.

llpc/context/llpcPipelineContext.cpp
trenouf: It's still used at line 335 in the case that the command-line options are not used and the tuning options enable the scalarizer without specifying a threshold.

trenouf:
V3: Review comments

amdvlk-admin:
### Test summary for commit 432dd29d2ff83a526bf344dc71ac772f0931590b

<li>PAL: dev b5e6dde2a0aa5dd2ce6d794e58d89ef715570f2d</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 5ae05ed8bd4f232c8580db96c021cad1e5c300e3</li>
</ul></details>

<li>Failed: 0/955 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/state/LgcContext.cpp
```diff
@@ -69,6 +70,21 @@ static cl::opt<bool> EmitLlvmBc("emit-llvm-bc", cl::desc("Emit LLVM bitcode inst
...
+//
+// @param name : Option name
+// @param value : Default option value
+template <typename T> static void setOptionDefault(const char *name, T value) {
+  auto optIterator = cl::getRegisteredOptions().find(name);
+  assert(optIterator != cl::getRegisteredOptions().end());
+  cl::Option *opt = optIterator->second;
+  if (opt->getNumOccurrences())
+    return;
+  *static_cast<cl::opt<T> *>(opt) = value;
```
kuhar, line 85: What happens when a wrong type is passed here? For example, the original option is `cl::opt<bool>` but `0` (int) is passes instead.

trenouf: It could crash horribly. Well, int vs bool would be ok, but e.g. int vs std::string would not.

kuhar: Is there a safer alternative then? Could we remove static from options' definitions and a list of `extern` options, or keep the statics but provide external storage for these options end make them extern instead?

trenouf: They're options in LLVM, and we don't really have a good justification for changing them upstream in that way.

I think it's much more likely that an upstream option would just disappear, rather than suddenly change type. In that case, this code silently does nothing.

Maybe I could make it safer by checking that the option is of the kind I expect. I could see whether that is possible.

jfroy:
This also turns out to fix a use-after-free bug in amdllpc where the `shaderCacheFileRootDir` char array was inserted in the `newArgs` vector then going out of scope before being referenced during argument parsing.

It would be nice if we could add at least asan shaderdb runs to the CI pipeline...

lgc/state/LgcContext.cpp
kuhar: > Maybe I could make it safer by checking that the option is of the kind I expect. I could see whether that is possible.

This sounds like a good idea to me.
Another way would be to expose these configuration knobs through llvm pass creation function arguments.

With the cast as-is, this seems to me like a disaster waiting to happen.

Flakebi:
I’m working on running tests with asan (and ubsan). Though I didn’t catch this bug with a shaderdb run…

lgc/state/LgcContext.cpp
trenouf: Found a typer-safer way of doing it.

trenouf:
V4: Review comment: Made it typesafe

Also snuck in a new commit "Simplified shadow descriptor table option handling".

trenouf:
Ping for approval

trenouf:
Thanks Rex. I'll wait for Jakub's timezone to see if he has comments.

PR #592, opened 2020-04-20
Upstream re-factor for getVectorElementType and CreateShuffleVector d…

dstutt:
…eprecation

Upstream LLVM is removing/deprecating getVectorElementType and
getVectorNumElements (in favour of a cast to VectorType and use of
getElementType and getNumElements)

CreateShuffleVector with unsigned indices is also being deprecated so adjusted
those as well.


dstutt:
retest this please

trenouf:
I will merge this.

PR #600, opened 2020-04-21
Fixed compatibility breakage in ElfReader

trenouf:
The change "Make Descriptor Offsets Relocatable" also fixed some method
names in ElfReader to the new camelBack standard. However that breaks
AMD internal builds, because we have internal code using those methods
for which we need to maintain some backward compatibility.

Also fixed a couple of other build problems.

Change-Id: I5617636cbd91ca55493bec41d9e2b669605bd84b


trenouf:
Fixed clang-format error.

JacobHeAMD:
retest this please

PR #602, opened 2020-04-21
Tidy up descriptor offset relocation changes

trenouf:
* Moved emitRelocationConstant from Internal to
BuilderBase::CreateRelocationConstant
* Also subsumed the reformatting of the same change from #594.

Change-Id: Ifc3e332af22e0cdac138517cf363622caaca11f3

PR #588, opened 2020-04-18
Switch PatchDescriptorLoad offsets to bytes

trenouf:
In the recent PatchDescriptorLoad revamp, the code deals with descriptor
offsets in a descriptor table in dwords. This commit changes it to deal
in bytes, as that makes it easier to add on future code to allow the
offset to be a reloc for shader compilation.

Change-Id: I96066dca3764f9ddb7d7865414cf9cd48e5bd475

PR #603, opened 2020-04-21
Fix assertion failure in PielineVsFs_RelocConst.pipe test

csyonghe:


amdvlk-admin:

trenouf:
Do the other two tests need the same fix? I was getting asserts on those too.

csyonghe:
No they don't. I don't get any assertion errors on the other tests.

trenouf:
Actually it looks like I broke the other two tests in #588. I'll fix it.

trenouf:
I've put my fix into #602.

trenouf:
I will merge this once #602 is approved and merged.

PR #609, opened 2020-04-22
Relocation for Root Descriptor offset

csyonghe:
This PR adds relocation for root descriptor offsets.

There are two places where I am not certain if the change is doing the right thing:
1. I disabled `useFixedLayout` when relocatable shader elf is being built. This makes sure that we always emit descriptor set value during initial compute shader elf generation.
2. I am not sure whether the existing code determines whether or not to spill a root descriptor based on PipelineInfo. If that this case, should we make the spill/non-spill decision a relocatable entry as well? This PR does not address this issue.

### DIFF
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -29,6 +29,7 @@
  ***********************************************************************************************************************
  */
 #include "PatchEntryPointMutate.h"
+#include "../interface/lgc/LgcContext.h"
 #include "Gfx6Chip.h"
 #include "Gfx9Chip.h"
 #include "lgc/state/IntrinsDefs.h"
@@ -308,7 +309,8 @@ FunctionType *PatchEntryPointMutate::generateEntryPointType(uint64_t *inRegMask)
   unsigned availUserDataCount = maxUserDataCount - userDataIdx;
   unsigned requiredRemappedUserDataCount = 0; // Maximum required user data
   unsigned requiredUserDataCount = 0;         // Maximum required user data without remapping
-  bool useFixedLayout = (m_shaderStage == ShaderStageCompute);
+  bool useFixedLayout =
+      (m_shaderStage == ShaderStageCompute && !m_pipelineState->getLgcContext()->buildingRelocatableElf());
   bool reserveVbTable = false;
   bool reserveStreamOutTable = false;
   bool reserveEsGsLdsSize = false;
@@ -550,7 +552,9 @@ FunctionType *PatchEntryPointMutate::generateEntryPointType(uint64_t *inRegMask)
         assert(node->sizeInDwords == 1);
 
         auto shaderOptions = &m_pipelineState->getShaderOptions(m_shaderStage);
-        if (shaderOptions->updateDescInElf && m_shaderStage == ShaderStageFragment) {
+        if (shaderOptions->updateDescInElf &&
+            (m_shaderStage == ShaderStageFragment || m_shaderStage == ShaderStageVertex ||
+             m_shaderStage == ShaderStageCompute)) {
           // Put set number to register first, will update offset after merge ELFs
           // For partial pipeline compile, only fragment shader needs to adjust offset of root descriptor
           // If there are more individual shader compile in future, we can add more stages here
--- a/llpc/util/llpcElfWriter.cpp
+++ b/llpc/util/llpcElfWriter.cpp
@@ -196,6 +196,60 @@ void ElfWriter<Elf>::mergeMapItem(llvm::msgpack::MapDocNode &destMap, llvm::msgp
 }
 
 // =====================================================================================================================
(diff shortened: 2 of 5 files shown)


amdvlk-admin:

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -29,6 +29,7 @@
  ***********************************************************************************************************************
  */
 #include "PatchEntryPointMutate.h"
+#include "../interface/lgc/LgcContext.h"
```
trenouf, line 32: This should be #include "lgc/LgcContext.h" (and then put in the right place).

trenouf:
> 1. I disabled `useFixedLayout` when relocatable shader elf is being built. This makes sure that we always emit descriptor set value during initial compute shader elf generation.

I think that is fine(ish), but that is why we need linking to generate a compute prolog. `useFixedLayout` implements the hardware (I think) restriction that a compute pipeline has to have its user data sgprs reflect a prefix of the root layout, without holes or reordering. But you don't know the root layout in shader compilation, so linking probably needs to generate the prolog to receive the registers in the right order, possibly load some missing ones, and pass them to the shader as it expects them. The "ish" is that, in the case of spilling where it looks like the spill pointer has to go in s15, we could still ensure that in shader compilation.

> 2. I am not sure whether the existing code determines whether or not to spill a root descriptor based on PipelineInfo. If that this case, should we make the spill/non-spill decision a relocatable entry as well? This PR does not address this issue.

I think it can decide which ones to spill without pipeline info.

csyonghe:
> > 1. I disabled `useFixedLayout` when relocatable shader elf is being built. This makes sure that we always emit descriptor set value during initial compute shader elf generation.
> 
> I think that is fine(ish), but that is why we need linking to generate a compute prolog. `useFixedLayout` implements the hardware (I think) restriction that a compute pipeline has to have its user data sgprs reflect a prefix of the root layout, without holes or reordering. But you don't know the root layout in shader compilation, so linking probably needs to generate the prolog to receive the registers in the right order, possibly load some missing ones, and pass them to the shader as it expects them. The "ish" is that, in the case of spilling where it looks like the spill pointer has to go in s15, we could still ensure that in shader compilation.
> 
> > 1. I am not sure whether the existing code determines whether or not to spill a root descriptor based on PipelineInfo. If that this case, should we make the spill/non-spill decision a relocatable entry as well? This PR does not address this issue.
> 
> I think it can decide which ones to spill without pipeline info.

Who defines this "prefix of the root layout"? By `PipelineLayout`? If so I guess each PipelineLayout corresponds to a prolog then. If we have some way to enforce that a compute shader always states the full pipeline layout in its parameter definition (regardless of whether they are actually used in the shader body), then it is possible that we don't need this linking step, is that right?

csyonghe:
Just out of curiosity, what is the downside to always derive the sgpr number solely from the descriptor set number? e.g. s0 is always for descriptor set 0. If the user specifies a really big number, just spill it.

nhaehnle:
> Just out of curiosity, what is the downside to always derive the sgpr number solely from the descriptor set number? e.g. s0 is always for descriptor set 0. If the user specifies a really big number, just spill it.

Pipeline layout compatibility in Vulkan is defined such that layouts are compatible if they have the same *prefix* of descriptor set layouts, but the *exact same* push constant ranges. This means we can optimize the user data layout to align with pipeline layouts by putting push constants before descriptor sets.

It's arguably up to the driver whether it wants to do that, but in any case we should support it in the compiler.

llpc/util/llpcElfWriter.cpp
```diff
@@ -195,6 +195,57 @@ void ElfWriter<Elf>::mergeMapItem(llvm::msgpack::MapDocNode &destMap, llvm::msgp
...
+static void updateRootDescriptorRegisters(Context *context, msgpack::Document &document) {
+  auto pipeline = document.getRoot().getMap(true)[Util::Abi::PalCodeObjectMetadataKey::Pipelines].getArray(true)[0];
+  auto registers = pipeline.getMap(true)[Util::Abi::PipelineMetadataKey::Registers].getMap(true);
+  const unsigned mmSpiShaderUserDataVs0 = 0x2C4C;
+  const unsigned mmSpiShaderUserDataPs0 = 0x2c0c;
+  const unsigned mmComputeUserData0 = 0x2E40;
+  unsigned userDataBaseRegisters[] = {mmSpiShaderUserDataVs0, mmSpiShaderUserDataPs0, mmComputeUserData0};
+  unsigned psUserDataCount = context->getGfxIpVersion().major < 9 ? 16 : 32;
+  for (auto baseRegister : userDataBaseRegisters) {
+    for (unsigned i = 0; i < psUserDataCount; ++i) {
```
nhaehnle, line 212: The userDataCount must be per-stage just like the base register: compute still only has 16 user data registers.

csyonghe:
Is it possible that we get these changed merged in first?

JaxLinAMD:
retest this please

PR #611, opened 2020-04-23
Redesign NGG GS work flow

amdrexu:
The processing is something like this:

 ```
NGG_GS() {
   Initialize thread/wave info

   if (threadIdInWave < vertCountInWave)
     Run ES

   if (threadIdInSubgroup < primCountInSubgroup)
     Initialize primitive connectivity data (0x80000000)
   if (threadIdInSubgroup < maxVertOutInSubgroup)
     Initialize draw flags of output vertices
   if (threadIdInSubgroup < waveCount + 1)
     Initialize per-wave and per-subgroup count of output vertices
   Barrier

   if (threadIdInWave < primCountInWave)
     Run GS
   Barrier

   if (culling && threadIdInSubgroup < primCountInSubgroup)
     Do culling and reset draw flags of output vertices
   Barrier

   if (threadIdInSubgroup < maxVertOutInSubgroup)
     Read draw flags of output vertices and compute draw mask

   if (threadIdInWave < waveCount - waveId)
     Accumulate per-wave and per-subgroup count of output vertices
   Barrier
   Read per-subgroup count of output vertices

   if (threadIdInSubgroup < maxVertOutInSubgroup &&
       vertex compaction && vertex is drawed)
     Setup vertex thread ID map (uncompacted -> compacted)
     Setup vertex thread ID reverse map (compacted -> uncompacted)

   if (waveId == 0)
     NGG allocation request (GS_ALLOC_REQ)
   Barrier

   if (threadIdInSubgroup < primCountInSubgroup)
     Do primitive connectivity data export

   if (threadIdInSubgroup < vertCountInSubgroup)
     Run copy shader
 }
```

The major differences compared with original one are:

1. Create three LDS regions to store draw flags of output vertices, vertex ID
   mapping (uncompacted -> compacted, compacted -> uncompacted). This is to
   simplify vertex compaction even if culling isn't enabled. Note that
   different GS primitive threads can emit different numbers of vertices (see
   the case VK.geometry.basic.output_10_and_100). Thus, vertex compaction is
   still needed. Original design avoids vertex ID mapping by using subgroup
   inclusive add. This will inevitably introduce two loops in later phases to
   correct vertex ID for primitive export and calculate vertex attribute data
   offset for vertex export.

2. Simplify GS mutation. With new design, GS doesn't return any info
   (originally GS return output vertex/primitive count info) and we don't have
   to mutate GS by cloning a new function. Function cloning is time-consuming.

3. Redesign the GS-VS ring layout. GS-VS ring is managed by shader and
   hareware for legacy pipeline. For NGG, the ring is fully controlled by
   shader. Original layout is unable to directly fetch vertex attribute data
   in copy shader by vertex thread ID in multi-stream cases. This needs
   additional vertex attribute offset calculation in a loop. We now group all
   vertices of the same stream.

The new layout is something like this (shader takes over it):

```
  +----------+----+----------+----+----------+----+----------+
  | Vertex 0 | .. | Vertex N | .. | Vertex 0 | .. | Vertex N |
  +----------+----+----------+----+----------+----+----------+
  |<------ Primitive 0 ----->| .. |<------ Primitive M ----->|
  |<----------------------- Stream i ----------------------->|

  +----------+----------+----------+----------+
  | Stream 0 | Stream 1 | Stream 2 | Stream 3 |
  +----------+----------+----------+----------+
  |<--------------- GS-VS ring -------------->|

  (M = prims_per_subgroup, N = max_vertices)
```

By contrast, GS-VS ring layout of non-NGG is something like this (conform to
hardware design):

```
  +----------+----+----------+----+----------+----+----------+
  | Vertex 0 | .. | Vertex N | .. | Vertex 0 | .. | Vertex N |
  +----------+----+----------+----+----------+----+----------+
  |<-------- Stream 0 ------>| .. |<-------- Stream 3 ------>|
  |<---------------------- Primitive i --------------------->|

  +-------------+----+-------------+
  | Primitive 0 | .. | Primitive M |
  +-------------+----+-------------+
  |<--------- GS-VS ring --------->|

  (M = prims_per_subgroup, N = max_vertices)
```

4. Primitive connectivity data are initialized to 0x80000000 before GS
   execution. However, the initialization is based on GS threads in wave. In
   the case dEQP-VK.geometry.basic.output_128, this will add many LDS write
   instructions (ds_write2_b32) for each output primitive. Rather, we can use
   the threads of output primitive to do this. This is the same case for
   output vertex draw flags and output vertex count per wavs.

5. The case dEQP-VK.geometry.layered.1d_array.render_different_content
   introduce oustanding vertices with this sequence 0 -> 1 -> CUT -> 2 ->
   3 -> 4 -> 5 for triangle_strip. Original design had additional processing
   to handle vertex 0 and 1. With the introduction of draw flags, we can now
   easily identify such vertices and needn't special processing in GS_EMIT
   and GS_CUT.

Change-Id: I5b52834e7d715f129532ad32c90090b42087fd92

### DIFF
--- a/lgc/patch/NggPrimShader.cpp
+++ b/lgc/patch/NggPrimShader.cpp
@@ -66,6 +66,30 @@ NggPrimShader::NggPrimShader(PipelineState *pipelineState)
   m_hasTcs = m_pipelineState->hasShaderStage(ShaderStageTessControl);
   m_hasTes = m_pipelineState->hasShaderStage(ShaderStageTessEval);
   m_hasGs = m_pipelineState->hasShaderStage(ShaderStageGeometry);
+
+  // NOTE: For NGG GS mode, we change data layout of output vertices. They are grouped by vertex streams now.
+  // Vertices belonging to different vertex streams are in different regions of GS-VS ring. Here, we calculate
+  // the base offset of each vertex streams and record them. See NggPrimShader::exportGsOutput for detail.
+  if (m_hasGs) {
+    unsigned vertexItemSizes[MaxGsStreams] = {};
+    auto resUsage = m_pipelineState->getShaderResourceUsage(ShaderStageGeometry);
+    for (unsigned i = 0; i < MaxGsStreams; ++i)
+      vertexItemSizes[i] = 4 * resUsage->inOutUsage.gs.outLocCount[i];
+
+    unsigned gsVsRingItemSizes[MaxGsStreams] = {};
+    const auto &geometryMode = m_pipelineState->getShaderModes()->getGeometryShaderMode();
+    for (unsigned i = 0; i < MaxGsStreams; ++i)
+      gsVsRingItemSizes[i] = vertexItemSizes[i] * geometryMode.outputVertices;
+
+    const unsigned gsPrimsPerSubgroup =
+        resUsage->inOutUsage.gs.calcFactor.gsPrimsPerSubgroup * geometryMode.invocations;
+    unsigned gsStreamBase = 0;
+    for (unsigned i = 0; i < MaxGsStreams; ++i) {
+      m_gsStreamBases[i] = gsStreamBase;
+      gsStreamBase += gsVsRingItemSizes[i] * gsPrimsPerSubgroup;
+    }
+  } else
+    memset(m_gsStreamBases, 0, sizeof(m_gsStreamBases));
 }
 
 // =====================================================================================================================
@@ -564,7 +588,7 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
     {
       m_builder->SetInsertPoint(expPrimBlock);
 
-      doPrimitiveExport();
+      doPrimitiveExportWithoutGs();
       m_builder->CreateBr(endExpPrimBlock);
     }
 
@@ -1514,7 +1538,7 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
     {
       m_builder->SetInsertPoint(expPrimBlock);
 
-      doPrimitiveExport(vertexCompact ? cullFlag : nullptr);
+      doPrimitiveExportWithoutGs(vertexCompact ? cullFlag : nullptr);
       m_builder->CreateBr(endExpPrimBlock);
     }
 
@@ -1658,13 +1682,6 @@ void NggPrimShader::constructPrimShaderWithGs(Module *module) {
 
   const unsigned waveCountInSubgroup = Gfx9::NggMaxThreadsPerSubgroup / waveSize;
 
-  const auto resUsage = m_pipelineState->getShaderResourceUsage(ShaderStageGeometry);
-  const unsigned rasterStream = resUsage->inOutUsage.gs.rasterStream;
-  assert(rasterStream < MaxGsStreams);
-
-  const auto &calcFactor = resUsage->inOutUsage.gs.calcFactor;
-  const unsigned maxOutPrims = calcFactor.primAmpFactor;
-
   auto entryPoint = module->getFunction(lgcName::NggPrimShaderEntryPoint);
 
   auto arg = entryPoint->arg_begin();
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by trenouf** [comment]
> and we don't have to mutate GS by cloning a new function. Function cloning is time-consuming.

Why are we cloning a function anyway? PatchEntryPointMutate mutates a function by creating the new one then moving the code across from the old one. No cloning needed.

**Comment by amdrexu** [comment]
`Why are we cloning a function anyway? PatchEntryPointMutate mutates a function by creating the new one then moving the code across from the old one. No cloning needed.`
This is because we want to return some primitive/vertex data after GS execution to the main primitive shader like this:
```
NGG_GS() {
  value = GS();
}
GS() {
}
ES() {
}
```
The void type of GS is changed: `void GS() -> Type GS()`. After that, GS becomes a new subfunction with new return type and old GS is removed. Now, all these are not needed any more.

**Comment by trenouf** [comment]
Good that that is not needed any more.

But my point is that, to mutate the arg or return types of a function, you don't need to clone it. That is too expensive; it copies all the code inside the function. Instead, create the new function then move the basic blocks across to it. None of that matters as you have removed it. :-)

**Comment by amdrexu** [comment]
I see we use `CloneFunctionInto` in `PatchEntryPointMutate::processShade`. Is it expected as your description? NGG non-GS still does function cloning by calling `CloneFunctionInto` because we want to separate pos0 data computation and deferred non-pos0 data with two subfunctions from original ES.

**Comment by trenouf** [comment]
> I see we use `CloneFunctionInto` in `PatchEntryPointMutate::processShade`. Is it expected as your description?

Oh, my mistake. I didn't notice that.

My new code for mutating CS subfunctions does it the way I describe: it creates the new empty function, moves basic blocks across to it, and changes uses of existing args to use the new function args instead.

I think we should change the existing PatchEntryPointMutate code to do that too.

**Comment by amdrexu** [comment]
I see. This explains why I encounter slow compilation if a API VS shader is large enough :)

File: lgc/patch/NggPrimShader.cpp [hunk 1]
```diff
@@ -3356,84 +3163,169 @@ Function *NggPrimShader::mutateGsToVariant(Module *module) {
...
+
+  auto reverseMapVertThreadIdBlock = createBlock(expVertBlock->getParent(), ".reverseMapVertThreadId");
+  reverseMapVertThreadIdBlock->moveAfter(expVertBlock);
 
-  // Mutate copy shader entry-point, handle GS output import
+  auto expVertContinueBlock = createBlock(expVertBlock->getParent(), ".expVertContinue");
+  expVertContinueBlock->moveAfter(reverseMapVertThreadIdBlock);
+
+  auto vertCompacted = m_builder->CreateICmpULT(m_nggFactor.vertCountInSubgroup, m_nggFactor.maxVertOutInSubgroup);
+  m_builder->CreateCondBr(vertCompacted, reverseMapVertThreadIdBlock, expVertContinueBlock);
```
nhaehnle, line 3193: This branch seems to be redundant with the equivalent branch in `constructPrimShader`.

nhaehnle:
> As a second corollary, there is a potential algorithmic change that would allow you to eliminate the OutVertThreadIdMap and OutVertDataFlag regions. The basic idea is as follows:
[...]
> 5. For prim exports, we first look up OutVertThreadIdReverseMap to map from compacted (vertex == primitive) index to uncompacted one. Then read OutPrimData to determine whether the primitive is enabled and what its winding should be. Then reconstruct the prim export dword.

As a further optimization, reading OutVertThreadIdReverseMap would only have to be done once, and the result could be shared between both primitive export and vertex export.

amdrexu:
I really appreciate your thought provoking suggestions :) I will reply the comments in following days. We can have more discussion.

lgc/patch/NggPrimShader.cpp
amdrexu: This is to avoid query vertex ID map if you don't need vertex compaction. The condition is based on this `actualNumEmittedVertices < expectedNumEmittedVertices`. With this condition, reconstruction of primitive connectivity data is unnecessary.

amdrexu:
I will re-consider your proposal and give your feedback. I think it is very innovative to emit more null primitives to make output primitive count equal to output vertex count. This really simplifies the handling. I didn't realize primitive export was not costly compared with vertex export. Because GS throughput is basically decided by vertsPerSubgroup, I think it is feasible to emit more primitives and let PA drop them.

amdrexu:
> Furthermore, if emitCounter was replaced by something like currentPrimVertCounter, then you could remove the flipVertOrder flag. 

Correct. Your proposal is good.

> Finally, I am somewhat concerned about the correctness of this code with regard to streamout and multiple streams.

The implementation uses different counters for different streams. Only raster stream writes vertex count info, primitive connectivity data, and vertex attribute data to LDS. Other streams are skipped. But GS-VS ring is allocated the space for those streams. If we want to support transform feedback, we can allow non-raster stream to write such info to LDS. Each stream will not interfere with each other because their info data are isolated. CTS has 10+ multi-stream cases to test different stream emit/cut and no issue is encountered.

lgc/patch/NggPrimShader.cpp
```diff
@@ -2068,112 +1929,84 @@ void NggPrimShader::constructPrimShaderWithGs(Module *module) {
...
 
     m_nggFactor.vertCountInSubgroup = vertCountInSubgroup;
 
-    auto firstWaveInSubgroup = m_builder->CreateICmpEQ(m_nggFactor.waveIdInSubgroup, m_builder->getInt32(0));
-    m_builder->CreateCondBr(firstWaveInSubgroup, allocReqBlock, endAllocReqBlock);
+    auto vertCompacted = m_builder->CreateICmpULT(vertCountInSubgroup, m_nggFactor.maxVertOutInSubgroup);
+    auto outVertValid = m_builder->CreateICmpULT(m_nggFactor.threadIdInSubgroup, m_nggFactor.maxVertOutInSubgroup);
+    auto vertThreadIdMapSetup = m_builder->CreateAnd(vertCompacted, outVertValid);
+    vertThreadIdMapSetup = m_builder->CreateAnd(vertThreadIdMapSetup, drawFlag);
+    m_builder->CreateCondBr(vertThreadIdMapSetup, setupVertThreadIdMapBlock, endSetupVertThreadIdMapBlock);
```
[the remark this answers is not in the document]
nhaehnle: Ah, I see, I understand now the skipping of compaction in that case. I like it!

lgc/patch/NggPrimShader.cpp
nhaehnle: Got it, thanks!

nhaehnle:
> The implementation uses different counters for different streams. Only raster stream writes vertex count info, primitive connectivity data, and vertex attribute data to LDS. Other streams are skipped. But GS-VS ring is allocated the space for those streams. If we want to support transform feedback, we can allow non-raster stream to write such info to LDS. Each stream will not interfere with each other because their info data are isolated. CTS has 10+ multi-stream cases to test different stream emit/cut and no issue is encountered.

Yeah, I saw the different counters, but I still don't fully understand it. What I didn't get is how the counters for the different streams are then used to affect the copy-shader behavior. I see the per-stream counters in `mutateGs`, but then they are no longer accessible from `constructPrimShaderWithGs`.

Actually, I think that we aren't supporting stream-out with NGG in the driver at this point, so all CTS tests with streamout would run with NGG disabled.

I think what we can do for this particular change is, leave a note in a comment that streamout is not properly supported yet.

amdrexu:
> Yeah, I saw the different counters, but I still don't fully understand it. What I didn't get is how the counters for the different streams are then used to affect the copy-shader behavior. I see the per-stream counters in mutateGs, but then they are no longer accessible from constructPrimShaderWithGs.

Yes, the counters after GS execution are inaccessible. The counters are used to write vertices of non-raster stream to GS-VS ring. Stream based offset in GS-VS ring is stored in the new member `m_streamBase[]`. What haven't done in my change is to allocate LDS spaces to calculate vertex counts of non-raster streams (maybe still need compaction or vertex mapping). In copy shader, I think the emulated structure in the future might be:

```
if (threadIdInSubgroup < vertCountInSubgroup) {
  if (threadIdInSubgroup < vertCountInStream0) {
  // Handle stream 0
  } else if (threadIdInSubgroup  < vertCountInStream0 + vertCountInStream1) {
  // Handle stream 1
  } else if (threadIdInSubgroup < vertCountInStream0 + vertCountInStream1 + vertCountInStream2) {
  // Handle stream 2
  } else
  // Handle stream 3
  }
}
```

> I think what we can do for this particular change is, leave a note in a comment that streamout is not properly supported yet.

Yes, NGG is turned off when transform feedback is tested. There is a note in PatchResouceCollect when we decide NGG options. Will make notes in NggPrimShader.

amdrexu:
@nhaehnle I finally follow your suggestion to rewrite the processing algorithm. It does simplify the final ISA codes and reduce the code size by ~20%. Since primData[N] corresponds to the starting vertex N, I don't write any other data to it except the winding info ( 0 or 1). This is enough to decide whether the primitive is valid one or not.

amdvlk-admin:
### Test summary for commit 4e556fdbfe79396bd84c4e6506dd2dca5a478b66

<li>XGL: dev 278ee18285659b865ddc65f3a156c98264c8c801</li>
</ul></details>

<li>Failed: 0/958 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 103198/103199 (100.0%)</li>
<li><details><summary>Failed: 1/103199 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.api.object_management.multithreaded_shared_resources.graphics_pipeline
Stack trace: result: 0xfffffffffffffff3 at vktApiObjectManagementTests.cpp:1827
```
</p></details></li>


</ul></details>

</ul></details>


</ul></details>

JaxLinAMD:
@nhaehnle , can you review the code please?

JaxLinAMD:
retest this please

PR #612, opened 2020-04-23
Moved ResourceUsage.cpp to state

trenouf:
I missed this one when separating non-builder stuff out of the builder
directory into state.

Change-Id: I063d94c8f27112c8f5e768155245d93efa97837a


trenouf:
This only ran CTS on one agent, but that passed.

PR #596, opened 2020-04-20
Cleanup amdllpc.cpp

kuhar:
This simplifies error handling in amdllpc.cpp:main and factors out windows path handling into a helper function.

PR #618, opened 2020-04-24
Reorganize the contents of doCulling()

amdrexu:
Add two helpers to encapsulate the operations of fetching vertex position data
and cull distance sign mask: fetchVertexPositionData() and
fetchCullDistanceSignMask(). The motivation of doing this is to support future
NGG GS culling. The whole processing of doCulling() could be totally shared.

Change-Id: I4a6aefc5b2547ad82c5afadce3df892b69f0f643

### DIFF
--- a/lgc/patch/NggPrimShader.cpp
+++ b/lgc/patch/NggPrimShader.cpp
@@ -2372,84 +2372,96 @@ void NggPrimShader::initWaveThreadInfo(Value *mergedGroupInfo, Value *mergedWave
 Value *NggPrimShader::doCulling(Module *module) {
   Value *cullFlag = m_builder->getFalse();
 
-  // Skip culling if it is not requested
+  Value *primData = nullptr;
+  if (m_hasGs) {
+    // NOTE: GS could produce invalid primitives during the processing of GS_CUT. If the primitive is not a valid one,
+    // we initialize the cull flag to TRUE to skip later culling.
+    primData = readPerThreadDataFromLds(m_builder->getInt32Ty(), m_nggFactor.threadIdInSubgroup, LdsRegionOutPrimData);
+    cullFlag = m_builder->CreateICmpEQ(primData, m_builder->getInt32(NullPrim));
+  }
+
+  // Skip following culling if it is not requested
   if (!enableCulling())
     return cullFlag;
 
-  auto esGsOffset0 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                                {
-                                                    m_nggFactor.esGsOffsets01,
-                                                    m_builder->getInt32(0),
-                                                    m_builder->getInt32(16),
-                                                });
-  auto vertexId0 = m_builder->CreateLShr(esGsOffset0, 2);
+  Value *vertexId0 = nullptr;
+  Value *vertexId1 = nullptr;
+  Value *vertexId2 = nullptr;
 
-  auto esGsOffset1 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                                {
-                                                    m_nggFactor.esGsOffsets01,
-                                                    m_builder->getInt32(16),
-                                                    m_builder->getInt32(16),
-                                                });
-  auto vertexId1 = m_builder->CreateLShr(esGsOffset1, 2);
+  if (m_hasGs) {
+    assert(primData);
+    assert(m_pipelineState->getShaderModes()->getGeometryShaderMode().outputPrimitive ==
+           OutputPrimitives::TriangleStrip);
 
-  auto esGsOffset2 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                                {
-                                                    m_nggFactor.esGsOffsets23,
-                                                    m_builder->getInt32(0),
-                                                    m_builder->getInt32(16),
-                                                });
-  auto vertexId2 = m_builder->CreateLShr(esGsOffset2, 2);
+    // Primitive data layout [31:0]
+    //   [31]    = null primitive flag
+    //   [28:20] = vertexId2 (in bytes)
+    //   [18:10] = vertexId1 (in bytes)
+    //   [8:0]   = vertexId0 (in bytes)
+    vertexId0 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
+                                           {primData, m_builder->getInt32(0), m_builder->getInt32(9)});
 
-  Value *vertexId[3] = {vertexId0, vertexId1, vertexId2};
-  Value *vertex[3] = {};
+    vertexId1 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
+                                           {primData, m_builder->getInt32(10), m_builder->getInt32(9)});
 
-  const auto regionStart = m_ldsManager->getLdsRegionStart(LdsRegionPosData);
-  assert(regionStart % SizeOfVec4 == 0); // Use 128-bit LDS operation
-  auto regionStartVal = m_builder->getInt32(regionStart);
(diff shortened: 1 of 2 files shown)


amdvlk-admin:
### Test summary for commit a7ee3110ca2f19c9b897ce535a6e862dc661995b

<li>PAL: dev e2f8eb0ffe21751b300995e795965d37e8d6dd67</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev f02738a5cf6c158c8c5b25a5048951f3059cabe2</li>
</ul></details>

<li>Failed: 0/958 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 41570/41571 (100.0%)</li>
<li><details><summary>Failed: 1/41571 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.type.scalar.i64.add_tessc
```
</p></details></li>


</ul></details>

</ul></details>


</ul></details>

lgc/patch/NggPrimShader.cpp
```diff
@@ -5998,6 +6010,70 @@ Value *NggPrimShader::doDppUpdate(Value *oldValue, Value *srcValue, unsigned dpp
...
+                                                  {cullDistance, m_builder->getInt32(31), m_builder->getInt32(1)});
+      signBit = m_builder->CreateShl(signBit, i);
+      signMask = m_builder->CreateOr(signMask, signBit);
+
+      return signMask;
+    }
+  } else {
+    // ES-only
+    return readPerThreadDataFromLds(m_builder->getInt32Ty(), vertexId, LdsRegionCullDistance);
+  }
```
kuhar, line 6074: Same as above.
In addition, it seems like the for loop does only 1 iteration?

lgc/patch/NggPrimShader.cpp
```diff
@@ -5998,6 +6010,70 @@ Value *NggPrimShader::doDppUpdate(Value *oldValue, Value *srcValue, unsigned dpp
...
+    const unsigned rasterStream = inOutUsage.gs.rasterStream;
+
+    const unsigned vertexItemSize = 4 * inOutUsage.gs.outLocCount[rasterStream];
+    auto vertexOffset = m_builder->CreateMul(vertexId, m_builder->getInt32(vertexItemSize));
+    vertexOffset = m_builder->CreateShl(vertexOffset, 2);
+
+    return importGsOutput(VectorType::get(m_builder->getFloatTy(), 4), loc, 0, rasterStream, vertexOffset);
+  } else {
+    // ES-only
+    return readPerThreadDataFromLds(VectorType::get(m_builder->getFloatTy(), 4), vertexId, LdsRegionPosData);
```
kuhar, line 6032: Can you flip the if condition and return early in case of `ES-only`?

amdrexu: Yes, it is a good practice to use early returns for multiple simple checks. Here, I still make a if-else is just to make the logic clear: two branches, ES-GS and ES-only. And we will do a middle-scale change to modify ES-only NGG in the future. And the ES-only branch might not be just a simple return after that. More calculation might be involved :)

lgc/patch/NggPrimShader.cpp
amdrexu: The for loop depends on cull distance array size. Cull distance is often declared as `float gl_CullDistance[]`. The processing is to handle each cull distance values and check its sign. Then, the signs are changed to a bit mask for future use. The bit layout is like this:

`Sign of CullDistance[7] |  Sign of CullDistance[6] | ... | Sign of CullDistance[0]`

lgc/patch/NggPrimShader.cpp
kuhar: Ack, thanks for the extra context.
In that case, I'd simply keep the same same if condition but remove the else, since the `if-then` must return anyway.

lgc/patch/NggPrimShader.cpp
kuhar: Taking a step back: just by looking at the code, the last line of the for loop is a `return`, and there is no `continue` or `break`. This code looks suspicious because it will always return at the end of the first iteration. I don't get it. I'd guess that the `return` was supposed to be outside of the loop, since each iteration updates a variable defined just outside of the loop.

lgc/patch/NggPrimShader.cpp
kuhar: To elaborate on what my point is: reaching the end of a non-void function without returning a value is an undefined behavior in C++, and these things are annoying to debug as they propagate internationally. When I look at a function and see a terminator at the end (`return` or `llvm_unreachable`), I don't have to worry at all about this particular kind of UB. When you write `if (...) { ... return X; } else { ... return Y; } }<end-of-function>` my worry is that even though the code looks correct right now, one of the return statements may be become conditional in the future, and the function could exit without returning.
I think this is why LLVM uses this style.

lgc/patch/NggPrimShader.cpp
amdrexu: Oh. Thanks a lot for catching this. It is my coding error.

lgc/patch/NggPrimShader.cpp
amdrexu: Thanks for your valuable suggestions. I will change the code to make it always have return.

lgc/patch/NggPrimShader.cpp
```diff
@@ -5998,6 +6010,76 @@ Value *NggPrimShader::doDppUpdate(Value *oldValue, Value *srcValue, unsigned dpp
...
 }
 
+// =====================================================================================================================
+// Fetches the position data for the specified vertex ID.
+//
+// @param vertexId : Vertex thread ID in sub-group.
+Value *NggPrimShader::fetchVertexPositionData(Value *vertexId) {
+  Value *posData = nullptr;
+
+  if (m_hasGs) {
```
kuhar, line 6020: I want to preface this comment by saying that I don't want to nitpick too much and won't block this PR in the current state if you are happy with it.

I think the current revision of this function is not a strict improvement over the previous one. Instead of two returns for each execution path, now we have one with extra data-flow. When reading or debugging code like this, one of the basic question for me is 'what values does this function return'. To answer this in the current form, you have to look at the single `return` and then track back all possible assignments to `posData`, and have to reason about control- and data-flow at the same time. This is not a big deal at all for this function because it's quite simple, but can lead complicated spaghetti as the code grows. I tried to address this exact issue in one of my recent PR's: #596. I think that the code would be much more maintainable if each execution path had it's own return statement and there was no infeasible control flow path with no return at all, like in the first revision of this PR.

If you structured the code like this:
```
Value *Class::foo() {
   if (...) {
    // ES-GS
    // ...
    return bar(...);
  }
  
  // ES-only
  return baz(...);
}
```
the code would look much easier to reason about and debug for me. If I wanted to know which execution path was taken, I could just set breakpoints on the two returns, and I wouldn't have to worry about the function not returning in some other execution path. Again, this is not a big deal at all in this particular case, but this is a more general observation that I made when reading a lot of existing code that declares a `Result result = Result::Success` and the beginning and then keeps branching on `results` -- in my view, a lot of code like this would benefit from a similar refactoring as in #596.

amdrexu: It is no nitpicking. I think it is OK to discuss the coding philosophy. What you have seen in original codes is caused by PAL style. The style requires only one return is allowed for each function. The reason is to have only one common exit in a function for debugging. You don't have to place many breakpoints on all returns to catch the run-time calculated return values. So the logic is often something like this:
```
func()
{
    result = Initial Value;
    if (result == Success)
    {
        // Do something
        result = Y;
    }
    if (result == Success)
    {
        // Do something
        result = Z;
    }
    ...
    return result;
}
```
Admittedly, different developers have different strategies. I will change the codes as you requested.

amdrexu: I finally adopt your suggestion in the initial review. Take ES-only path as a special one
```
func() {
  if (!m_hasGs) {
    // ES-only
  }
  // ES-GS
}

```

PR #634, opened 2020-04-27
[CI] Add CI configurations for assertions, clang, and sanitizers

kuhar:
This extends the current public CI infrastructure with 5 new build
configurations. Each of configuration has its own base image cached
in GCR. This way each incremental CI build doesn't have to know about
the precise CMake invocation and resumes from whatever configuration
its base imgage used.

GitHub actions offer 10 free concurrent CI jobs, so this should
significanly slow down the public CI, with 4 workers still unused.

### DIFF
--- a/.github/workflows/build-amdvlk-docker.yml
+++ b/.github/workflows/build-amdvlk-docker.yml
@@ -6,16 +6,18 @@ on:
 
 jobs:
   build-and-push-amdvlk:
-    name: Branch ${{ matrix.branch }}, base image ${{ matrix.base-image }}, config ${{ matrix.config }}
+    name: "Features: ${{ matrix.feature-set }}, assertions: ${{ matrix.assertions }}"
     if: github.repository == 'GPUOpen-Drivers/llpc'
     runs-on: ${{ matrix.host-os }}
     strategy:
       matrix:
-        host-os:      ["ubuntu-latest"]
-        base-image:   ["gcr.io/stadia-open-source/amdvlk_dev_release:nightly"]
-        branch:       [dev]
-        config:       [Release]
-        generator:    [Ninja]
+        host-os:        ["ubuntu-18.04"]
+        image-template: ["gcr.io/stadia-open-source/amdvlk_%s%s_%s:nightly"]
+        branch:         [dev]
+        config:         [Release]
+        assertions:     ["OFF", "ON"]
+        feature-set:    ["+gcc", "+clang"]
+        generator:      [Ninja]
     steps:
       - name: Checkout LLPC
         run: |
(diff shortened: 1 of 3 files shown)


amdvlk-admin:

JaxLinAMD:
@kuhar , how about always enable assertion and sanitzers? then we only have two build variants:
gcc, clang

Flakebi:
Looks good in general, thanks for working on this.

I have a few notes on the sanitizers:
- I need to get some patches into the xgl cmake configuration before the driver compiles with sanitizers enabled, so I think it makes sense to enable this CI configuration in a later PR.
- `libclang-common-6.0-dev` (or `libclang-common-9.0-dev`) is needed for the asan runtime libraries. We need to use the shared asan version, the statically linked version is not supported for shared libraries as far as I know.
- To use it, we need to set `LD_PRELOAD=/usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.asan-x86_64.so` (or corresponding 9 version).
- Using tablegen in debug+sanitizer mode is really slow (it might be better in release mode), `-DLLVM_OPTIMIZED_TABLEGEN=1` improves that but it needs [D78913](https://reviews.llvm.org/D78913) in LLVM.
- The flag is called `-DLLVM_USE_SANITIZER=…` (note the missing s) which cmake warns about [hidden in the 5000 line log](https://github.com/kuhar/llpc/runs/623268521?check_suite_focus=true#step:6:1704) ;)

kuhar:
> @kuhar , how about always enable assertion and sanitzers? then we only have two build variants:
> gcc, clang
Having assertions always on is not a good idea. They affect ABI and can hide bugs -- a simple example is code being correct only when assertions are executed, or vice versa. When it comes to sanitizers, I think that in practice they can complain about some issues that are pretty low-priority to fix, and it would be good to know that even though they complain, regular builds continue working.

kuhar:
> Looks good in general, thanks for working on this.
> 
> I have a few notes on the sanitizers:

Thanks for pointing the issues out.

If you are happy with the general direction of this PR, perhaps it would be best to commit only the clang part now and add builds with assertions and sanitizers later, once the CMake build properly support them?

kuhar:
I updated the PR to run without sanitizers for now. The most important thing for me is that we have a mechanism to support these in the future and that there's a clear way to add more 'features' as we need.

@Flakebi Can we introduce 2 new flags:
1. `XYZ_ENABLE_ASSERTIONS` that enables assertions also in release mode, and probably forwards to `LLVM_ENABLE_ASSERTIONS`.
2. `XYZ_ENABLE_SANITIZER` that works like the LLVM one but sets everything up like you described.
Where `XYZ` is `XGL` or `LLPC`, whichever is more appropriate. I'm not very familiar with how the CMake builds are setup for amdvlk and would appreciate if someone else took over that part of the CI setup.

Flakebi:
> I updated the PR to run without sanitizers for now. The most important thing for me is that we have a mechanism to support these in the future and that there's a clear way to add more 'features' as we need.

Looks good!

1. `XYZ_ENABLE_ASSERTIONS`: Sounds good, enabling `LLVM_ENABLE_ASSERTIONS`, `PAL_ENABLE_PRINTS_ASSERTS` and removing `NDEBUG` from the release definitions somehow should do this. (Although PAL assertions shouldn’t change anything as we cannot test PAL code on GitHub anyway)
2. `XYZ_ENABLE_SANITIZER`: I’m not quite sure how to integrate that into the cmake declarations because I don’t want to duplicate all the code from LLVM for `LLVM_ENABLE_SANITIZER`, which already sets global flags. So far I was just reusing `LLVM_ENABLE_SANITIZER`.

I’ll try to get these flags into xgl.

kuhar:
> 2\. `XYZ_ENABLE_SANITIZER`: I’m not quite sure how to integrate that into the cmake declarations because I don’t want to duplicate all the code from LLVM for `LLVM_ENABLE_SANITIZER`, which already sets global flags. So far I was just reusing `LLVM_ENABLE_SANITIZER`.

You mentioned something about having to set `LD_PRELOAD` -- I thought that this can be setup in CMake?

Flakebi:
> You mentioned something about having to set LD_PRELOAD -- I thought that this can be setup in CMake, in addition to forwarding the flag to the llvm one. Is that right?

Actually, ignore that one, I think it only plays a role when using a shared library (amdvlk64.so), for example when running the CTS, not when running unit tests.

PR #638, opened 2020-04-28
Add PipelineState patching in llpcElfWriter linking

csyonghe:
This PR adds patching of pipeline states into the value of PA_CL_CLIP_CNTL register during the linking step of relocatable pipeline compilation.

Three pipeline settings are implemented here: u`srClipPlaneMask`, `zClipEnable` and `rasterizaerDiscardEnable`.

### DIFF
--- a/llpc/test/shaderdb/PipelineVsFs_MetaPatching.pipe
+++ b/llpc/test/shaderdb/PipelineVsFs_MetaPatching.pipe
@@ -1,0 +1,106 @@
+
+// This test case checks that metadata patching works correctly during vs/fs pipeline linking.
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -use-relocatable-shader-elf -v %gfxip %s | FileCheck -check-prefix=SHADERTEST1 %s
+; SHADERTEST1: PA_CL_CLIP_CNTL 0x000000000D480005
+; END_SHADERTEST
+
+[VsGlsl]
+#version 450
+#extension GL_ARB_separate_shader_objects : enable
+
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

trenouf:
Please rebase.

trenouf:
> I have to say that I'm not too fond of this, and would want us to choose a different path here.
> 
> The programming of registers for the PAL metadata generally happens in Gfx{6,9}ConfigBuilder.cpp. Can we make sure that we keep it that way, perhaps by refactoring that code in such a way that it can be used both during regular pipeline compilation **and** during linking?

#629 creates a new PalMetadata class in LGC, and I plan to move setting of PAL metadata that comes from pipeline state into a method in PalMetadata, used both from pipeline compilation and from linking. (I will be moving linking into LGC.) So can we put this change on hold for now please?

nhaehnle:
> #629 creates a new PalMetadata class in LGC, and I plan to move setting of PAL metadata that comes from pipeline state into a method in PalMetadata, used both from pipeline compilation and from linking. (I will be moving linking into LGC.) So can we put this change on hold for now please?

Yes, that does seem like a good step in the right direction.

csyonghe:
Please let me know when the planned changes went through and is ready to work on this PR again.

I see #629 merged, but that change doesn't seem to include the CNTL registers.

trenouf:
Hi Yong. #671, #581 and #673 (all in review) between them implement the equivalent of this, but keeping common code for setting these items between the pipeline compile case and the linker case.

JaxLinAMD:
retest this please

JaxLinAMD:
need a rebase

s-perron:
Implemented in other PRs.

PR #644, opened 2020-04-29
Fix stride relocation regression

csyonghe:
The descriptor stride was at some point refactored to be number of bytes instead of number of dwords, but the fixup logic has not been updated with the refactoring. This PR patches up this issue.


amdvlk-admin:

JaxLinAMD:
retest this please

PR #645, opened 2020-04-29
Use the correct ShuffleVector constructor

kuhar:
The old and deprecated constructor uses uint32 masks.
The new one takes int indices.

PR #648, opened 2020-04-30
Fix another use of deprecated CreateShuffleVector overload.

jayfoad:
This one crept in after PR #592.


kuhar:
This is a duplicate of #645, no?

jayfoad:
> This is a duplicate of #645, no?

Great minds think alike!

PR #650, opened 2020-04-30
Rationalized user data usage

trenouf:
(Only the top two commits belong to this PR.)

Previously, user data usage (descriptors and push constants) worked like
this:

1. DescBuilder generated calls representing descriptor loads, push
   constant pointer
2. PatchResourceCollect noted which user data nodes are in use, and
   stored that information in ResourceUsage, an outside-IR structure.
3. PatchEntryPointMutate decided which user data nodes go in entry
   SGPRs, and stored the information in ResourceUsage
4. PatchPushConst attempted to optimize push const usage to use entry
   SGPRs instead of the spill table
5. PatchDescriptorLoad lowered the calls generated in (1) using the user
   data in SGPRs information from (3).

As part of reducing the amount of outside-IR data carried through the
passes, and also to enable a future commit that will be able to do all
this in a shader compilation with no user data layout, this commit
changes to a new scheme:

1. DescBuilder generates descriptor loads directly, in terms of calls
   representing descriptor _sets_ (and push constant pointer, and root
   descriptors as long as user data layout is available).
2. PatchEntryPointMutate directly gathers descriptor set (etc) usage,
   then decides what user data goes into entry SGPRs, then lowers the
   calls from (1) based on that. None of that usage or SGPR information
   needs to be stored outside this pass.


JaxLinAMD:
@trenouf , this pull request causes jenkins agents to hang, please first fix the hang on your local machine.

trenouf:
> @trenouf , this pull request causes jenkins agents to hang, please first fix the hang on your local machine.

OK, sorry for the trouble it caused.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -46,26 +51,89 @@ using namespace llvm;
...
 } // namespace cl
-
 } // namespace llvm
 
-namespace lgc {
+namespace {
+
+// =====================================================================================================================
+// The entry-point mutation pass
+class PatchEntryPointMutate : public Patch {
```
linqun, line 68: why do you need remove the name space?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
 
 // =====================================================================================================================
-// Generates the type for the new entry-point based on already-collected info in LLVM context.
+// Generates the type for the new entry-point based on already-collected info.
+// This is what decides what SGPRs and VGPRs are passed to the shader at wave dispatch:
+//
+// * (For a GFX9+ merged shader or NGG primitive shader, the 8 system SGPRs at the start are not accounted for here.)
+// * The "user data" SGPRs, up to 32 (GFX9+ non-compute shader) or 16 (compute shader or <=GFX8). Many of the values
+//   here are pointers, but are passed as a single 32-bit register and then expanded to 64-bit in the shader code:
+//   - The "global information table", containing various descriptors such as the inter-shader rings
```
linqun, line 607: In the initial development of LLPC,  we keep user data layout close to internal compiler's layout to speed up the development.
now, I think It is time to remove unnecessary field, include per-shader table and etc.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
+//   - The "per-shader table", which is added here but appears to be unused
+//   - The streamout table if needed
+//   - Nodes from the root user data layout, including pointers to descriptor sets. In a compute shader, these
+//     must have the same layout as the root user data layout, even if that would mean leaving gaps in register
+//     usage, and are limited to 10 SGPRs s2-s11
+//   - For a compute shader, the spill table pointer if needed, which always goes into s12. (This is possibly
```
linqun, line 613: It is a PAL limitation. it only support remap user data in graphics pipeline. I think PAL should remove this limitation in the future.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
+  //
+  // The UserDataArg for each arg pushed into these vectors contains:
+  // - argTy: The IR type of the arg
+  // - userDataValue: The PAL metadata value to be passed to PalMetadata::setUserDataEntry, or Invalid for none
+  // - argIndex: Pointer to the location where we will store the actual arg number, or nullptr
+  // - isPadding: Whether this is a padding argument for compute shader fixed layout
+  // - mustSpill: The node must be loaded from the spill table; do not attempt to unspill
+
+  SmallVector<UserDataArg, 8> userDataArgs;
+  SmallVector<UserDataArg, 4> extraUserDataArgs;
```
linqun, line 649: I prefer to use specialUserDataArgs to match your comments.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
   auto intfData = m_pipelineState->getShaderInterfaceData(m_shaderStage);
-  auto resUsage = m_pipelineState->getShaderResourceUsage(m_shaderStage);
+  auto &entryArgIdxs = intfData->entryArgIdxs;
+  entryArgIdxs.initialized = true;
+
+  // First we collect the user data args in two vectors:
+  // - userDataArgs: global table, per-shader table and streamout table, followed by the nodes from the root user
+  //   data layout (excluding vertex buffer and streamout tables). Some of these are marked mustSpill; some of them
+  //   may need to be spilled anyway due to running out of entry SGPRs
+  // - extraUserDataArgs: special values that go at the end, such as ViewId.
```
linqun, line 639: why do we need put ViewId in the end of user datas? is it limited by PAL or just to simplify the design?
If I remember correctly, one reason is some special user data can't be put in spill table in PAL. so what will happen if we put these in the beginning of user data?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
-
-      if (node->type == ResourceNodeType::StreamOutTableVaPtr) {
-        // Only the last shader stage before fragment (ignoring copy shader) needs a stream out table.
-        if ((m_pipelineState->getShaderStageMask() &
-             (shaderStageToMask(ShaderStageFragment) - shaderStageToMask(m_shaderStage))) ==
-            shaderStageToMask(m_shaderStage))
+  for (auto &node : m_pipelineState->getUserDataNodes()) {
+    if (node.type == ResourceNodeType::IndirectUserDataVaPtr) {
+      reserveVertexBufferTable = true;
+    } else if (node.type == ResourceNodeType::StreamOutTableVaPtr) {
```
linqun, line 716: internal compiler has removed node type StreamOutTableVaPtr, can we remove it too?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
+      // Only the last shader stage before fragment (ignoring copy shader) needs a stream out table.
+      if ((m_pipelineState->getShaderStageMask() &
```
linqun, line 718: Could you help to add more comment for the stream out stage check? this code is too tricky.

trenouf:
> ps: why do we need getInternalPerShaderTablePtr? if i remember correctly, we never use per pipeline table in LLPC.

You're right; nothing is using it. Perhaps I should remove it.

Can PAL cope with us not asking for it at all, and starting our user data at s1 instead of s2?

lgc/patch/SystemValues.h
```diff
@@ -162,12 +162,9 @@ class ShaderSystemValues {
...
   llvm::SmallVector<llvm::Value *, 8> m_descTablePtrs;       // Descriptor table pointers
   llvm::SmallVector<llvm::Value *, 8> m_shadowDescTablePtrs; // Shadow descriptor table pointers
-  llvm::Value *m_internalGlobalTablePtr = nullptr;
-  // Internal global table pointer
-  llvm::Value *m_internalPerShaderTablePtr = nullptr;
-  // Internal per shader table pointer
-  llvm::Value *m_spilledPushConstTablePtr = nullptr;
-  // Spilled push constant pointer
+  llvm::Instruction *m_internalGlobalTablePtr = nullptr;     // Internal global table pointer
+  llvm::Value *m_internalPerShaderTablePtr = nullptr;        // Internal per shader table pointer
```
amdrexu, line 148: Internal per-shader table is not used in LLPC. Maybe we can skip it now. If I remember correctly, it is still added as user_data_1 because PAL once expected us to provide it. I think you can make a separate change to have a try. If no CTS fails, it is safe to remove it and make s1 available to other uses.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -284,538 +352,512 @@ bool PatchEntryPointMutate::isResourceNodeActive(const ResourceNode *node, bool
...
-        break;
-      }
-      }
-    } else {
-      if (needSpill && intfData->spillTable.offsetInDwords == InvalidValue)
-        intfData->spillTable.offsetInDwords = node->offsetInDwords;
-      m_pipelineState->getPalMetadata()->setUserDataSpillUsage(node->offsetInDwords);
+    default:
+      // Any other descriptor type in the root table is accessed via the spill table.
+      mustSpill = true;
```
linqun, line 962: why do need set mustSpill for all other resource types? I think we don't have this limitations.
as far as I know, some RPM shader put the resource/sampler in root.

lgc/state/PalMetadata.cpp
```diff
@@ -99,3 +136,126 @@ void PalMetadata::record(Module *module) {
...
+// @param stage : ShaderStage
+unsigned PalMetadata::getUserDataReg0(ShaderStage stage) {
+  if (m_userDataRegMapping[stage] != 0)
+    return m_userDataRegMapping[stage];
+
+  // Mapping not yet initialized.
+  // Set up ShaderStage -> user data register mapping.
+  m_userDataRegMapping[ShaderStageCompute] = mmCOMPUTE_USER_DATA_0;
+  m_userDataRegMapping[ShaderStageFragment] = mmSPI_SHADER_USER_DATA_PS_0;
+  if (!m_pipelineState->getNggControl()->enableNgg)
```
linqun, line 154: please group them with gfx ip. thanks!

lgc/state/PalMetadata.cpp
```diff
@@ -99,3 +136,126 @@ void PalMetadata::record(Module *module) {
...
+  unsigned userDataReg = getUserDataReg0(stage);
+
+  // Assert that the supplied user data index is not too big.
+  assert(userDataIndex + dwordCount <= 32 &&
+         (m_pipelineState->getTargetInfo().getGfxIpVersion().major >= 9 || stage != ShaderStageCompute ||
+          userDataIndex + dwordCount <= 16) &&
+         "Out of range user data index");
+
+  // Update userDataLimit if needed.
+  if (userDataValue < 0x10000 && userDataValue + dwordCount > m_userDataLimit->getUInt())
```
linqun, line 221: What's the meaning of 0x10000?

lgc/state/PalMetadata.cpp
```diff
@@ -99,3 +136,126 @@ void PalMetadata::record(Module *module) {
...
+  // If there are root user data nodes but none of them are used, adjust userDataLimit accordingly.
+  if (m_userDataLimit->getUInt() == 0 && !m_pipelineState->getUserDataNodes().empty())
+    setUserDataLimitToMax();
+}
+
+// =====================================================================================================================
+// Set userDataLimit to maximum (the size of the root user data table, excluding vertex buffer and streamout).
+// This is called if spill is in use, or if there are root user data nodes but none of them are used (PAL does
+// not like userDataLimit being 0 if there are any root user data nodes).
+void PalMetadata::setUserDataLimitToMax() {
```
linqun, line 258: I prefer to use name setUserDataLimit. setUserDataLimitToMax is easy to causes confuse that this function will set userDataLimit to a pre-defined max value.

lgc/builder/DescBuilder.cpp
```diff
@@ -39,6 +40,14 @@
 using namespace lgc;
 using namespace llvm;
 
+namespace {
+const unsigned DescriptorSizeResource = 8 * sizeof(unsigned);
```
linqun, line 44: I think this info should be a part of gpu property

lgc/state/PalMetadata.cpp
```diff
@@ -45,19 +45,40 @@ namespace {
...
+//
+// Note on LS/HS confusion:
+// <=GFX8 claims LS registers are from 0x2D4C and HS registers are from 0x2D0C
+// GFX9 claims LS registers are from 0x2D0C, and the LS-HS merged shader uses them
+// GFX10 claims HS registers are from 0x2D0C, and the LS-HS merged shader uses them.
+// So here we call the registers from 0x2D0C "HS" and have the LS-HS merged shader using them, for
+// consistency. That contradicts the GFX9 docs, but has the same effect.
+//
+// First the ones that only apply up to GFX8
+constexpr unsigned int mmSPI_SHADER_USER_DATA_LS_0 = 0x2D4C;
```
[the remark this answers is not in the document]
trenouf: Using the register definitions in imported si_ci_vi_merged_offset.h and gfx9_plus_merged_offset.h is a pain, and overcomplicated given that register numbers do not change very much between architectures. Even with those imported files, there is still a mechanism to cope with what changes there are in gfx9Chip etc, and that is overcomplicated.

LLVM does not use those imported files. Given that a compiler only uses a small subset of the registers, it is better to just copy the definitions and cope with between-generation differences as and when needed. Code changes are needed for a new architecture that has register changes anyway.

So I am moving more towards the LLVM way of doing it.

Per the comment above, these particular register numbers did not change at all between gfx6-gfx10, as long as you ignore the anomaly of how the LS/HS ones are documented in gfx9.

There are separate definitions in the pipeline dumper, which cannot be compiled with LLPC/LGC code.

lgc/builder/DescBuilder.cpp
trenouf: OK, but I think I should not put the compact buffer and ycbcr sampler sizes in there as they are driver constructs and nothing to do with hardware.

lgc/state/PalMetadata.cpp
trenouf: Good point. :-) I'll use a named constant and improve the comment.

trenouf:
V4: Fixed compact buffer desc and constant indexing of root desc.
        Fixed GS spilled used data when VS merged in.
        Review comment: Put HW descriptor size consts into GPUProperty.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: There is no need for out-of-line method definitions to be done inside their namespace. Also, if you define a previously-declared function (lgc::createPatchEntryPointMutate here), the LLVM coding standards doc specifically says do not do that inside the namespace, so the build catches any error earlier.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: Good idea for later commits...

lgc/patch/PatchEntryPointMutate.cpp
trenouf: I don't know; I just copied what it did before. :-) I suspect it was to ensure it is in the same register in two shaders that are merged.

Putting at the beginning of user data should be fine in terms of not spilling. The global table and streamout table already do that.

But I propose not changing anything like that in this change.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: Yes -- a change I have not pushed into a PR yet stops looking in the user data layout for IndirectUserDataVaPtr (for vertex buffer) and StreamOutTableVaPtr.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: I think I added that limitation in an earlier change.

Because spir-v allows you to pass a pointer to an image or sampler descriptor through phi nodes and function calls, the earlier change properly separated the action of taking the pointer to the descriptor from the action of loading the descriptor.

With that separation, not spilling a root table image or sampler descriptor would require a peephole to link the two parts together. I did not implement that peephole because I thought that the only case of a root descriptor is a "dynamic" buffer descriptor.

What is an RPM shader? Some special internal shader not from the user program?

lgc/patch/SystemValues.h
trenouf: Good idea for a future change. Presumably we still need to start in s2 in compute shaders.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: I changed it to use PipelineState::getLastVertexProcessingState, a method I had already created in an earlier change.

JaxLinAMD:
retest this please

JaxLinAMD:
llpc lit test failed.

lgc/patch/PatchEntryPointMutate.cpp
amdrexu: RPM shaders are internal shaders, created in advance to execute fixed pipeline functions of graphics system. You can refer the shader sources and pipeline info in /pal/src/core/hw/gfxip/rpm/shaders from internal code base.

trenouf:
V5: Better array of root desc and push const handling. Various fixes.
    V6: Avoid part spilling on CS fixed layout.

Reviewers might like to revisit this. :-) The top commit "Rationalized user data usage" has substantial changes to handle push const and array of root descriptors better. From the commit message:

    The unspilling algorithm has better results than the old one in a couple
    of cases:
    
    * An array of descriptors in the root table: As long as the array is not
      variably indexed, each entry is considered separately, so unused ones
      do not take an SGPR, and it might be that some entries are unspilled
      and some are spilled. (However, if the array is variably indexed, or
      this is a compute shader with fixed layout, the array is completely
      spilled.)
    * Similarly, if the push constant is a struct or array that is not
      variably indexed and is not loaded as the whole aggregate, the
      elements are considered individually. (However, if some part of the
      push constant is variably indexed, then that part is spilled, and if
      parts of the push constant are separately used on compute shader with
      fixed layout, then the whole push constant is spilled.)

trenouf:
Sorry, that one was completely borked, and I have probably crashed at least one of the test agents. I'll try and figure out what needs rebooting.

trenouf:
V8: CS fixed layout fix: don't unspill stuff after spilled desc array or
        partial-use push const

trenouf:
Rebased.

amdvlk-admin:
### Test summary for commit a8c589222d535d6f880d633465f4a6effbac2c02

<li>XGL: dev 278ee18285659b865ddc65f3a156c98264c8c801</li>
</ul></details>

<li>Passed: 959/959 (100.0%)</li>
<li>Failed: 0/959 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1069,6 +1506,42 @@ uint64_t PatchEntryPointMutate::pushFixedShaderArgTys(SmallVectorImpl<Type *> &a
...
+    m_userDataUsage[stage] = std::make_unique<UserDataUsage>();
+  return &*m_userDataUsage[stage];
+}
+
+// =====================================================================================================================
+// Get the shader stage that the given shader stage is merged into.
+// For GFX9+:
+// VS -> TCS (if it exists)
+// VS -> GS (if it exists)
+// TES -> GS (if it exists)
```
amdrexu, line 1526: Add a @param comment, describe the input stage is the original stage before merging.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1069,6 +1506,42 @@ uint64_t PatchEntryPointMutate::pushFixedShaderArgTys(SmallVectorImpl<Type *> &a
...
+ShaderStage PatchEntryPointMutate::getMergedShaderStage(ShaderStage stage) const {
+  if (m_pipelineState->getTargetInfo().getGfxIpVersion().major >= 9) {
+    switch (stage) {
+    case ShaderStageVertex:
+      if (m_pipelineState->hasShaderStage(ShaderStageTessControl))
+        return ShaderStageTessControl;
+      LLVM_FALLTHROUGH;
```
amdrexu, line 1533: This is very interesting. LLVM define such marker. Previously, PAL style requires a comment // Fall through for such use. But I think it is OK to use this marker. Better to write it to coding style doc

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -104,23 +105,121 @@ class PatchEntryPointMutate : public Patch {
...
   bool m_hasGs;                             // Whether the pipeline has geometry shader
   PipelineState *m_pipelineState = nullptr; // Pipeline state from PipelineStateWrapper pass
+  // Per-HW-shader-stage gathered user data usage information.
+  llvm::SmallVector<std::unique_ptr<UserDataUsage>, ShaderStageCount> m_userDataUsage;
+};
+
+// =====================================================================================================================
+// Utility class to generate code to extend an i32 address into a 64-bit pointer, using either PC or a given
+// constant high half value.
+class PointerExtender {
```
amdrexu, line 172: I think the pointer is used as address, maybe rename it to AddressExtender or something similar. I just believe pointer is too general.

lgc/builder/BuilderImpl.cpp
```diff
@@ -166,13 +166,53 @@ Instruction *BuilderImplBase::createWaterfallLoop(Instruction *nonUniformInst, A
...
+            continue;
+          }
+          // Variable GEP, to provide the index for the waterfall.
+          if (index || gep->getNumIndices() != 1)
+            break;
+          index = *gep->idx_begin();
+          base = gep->getPointerOperand();
+          continue;
+        }
+        if (auto extract = dyn_cast<ExtractValueInst>(base)) {
```
amdrexu, line 189: Consider changing it to extractValue. Extract/insert is OK in this context since the branch is very simple.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -104,23 +105,121 @@ class PatchEntryPointMutate : public Patch {
...
+    // calls must be kept.
+    bool pushConstSpill = false;
+    // Per-push-const-offset lists of loads from push const. We attempt to unspill these.
+    SmallVector<UserDataNodeUsage, 8> pushConstOffsets;
+    // Per-user-data-offset lists of lgc.root.descriptor calls
+    SmallVector<UserDataNodeUsage, 8> rootDescriptors;
+    // Per-descriptor-set lists of lgc.descriptor.set calls
+    SmallVector<UserDataNodeUsage, 8> descriptorSets;
+    // Minimum offset at which spill table is used.
+    unsigned spillUsage = UINT_MAX;
```
amdrexu, line 132: Actually, the spill threshold is set to UINT16_MAX by internal compiler :) Not an issue, but this could make the dump info consistent without any differences.

lgc/builder/DescBuilder.cpp
```diff
@@ -243,8 +511,105 @@ Value *DescBuilder::scalarizeIfUniform(Value *value, bool isNonUniform) {
...
+
+  sqBufRsrcWord3.bits.dstSelX = BUF_DST_SEL_X;
+  sqBufRsrcWord3.bits.dstSelY = BUF_DST_SEL_Y;
+  sqBufRsrcWord3.bits.dstSelZ = BUF_DST_SEL_Z;
+  sqBufRsrcWord3.bits.dstSelW = BUF_DST_SEL_W;
+  sqBufRsrcWord3.gfx6.numFormat = BUF_NUM_FORMAT_UINT;
+  sqBufRsrcWord3.gfx6.dataFormat = BUF_DATA_FORMAT_32;
+  assert(sqBufRsrcWord3.u32All == 0x24FAC);
+
+  // DWORD0
```
amdrexu, line 544: I plan to do a change clearing some coding issues. DWORD -> Dword or dword is in the list.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: I get a warning (error with -Werror) if I do not use it.

I just checked, and I was surprised that it is not mentioned in the LLVM coding standards doc. So, yes, we should add it to our doc.

lgc/builder/DescBuilder.cpp
trenouf: I have now changed these ones.

amdvlk-admin:
### Test summary for commit 46ec174fd259862a27e62ebd0669bae1da113678

</ul></details>

<li>Passed: 959/959 (100.0%)</li>
<li>Failed: 0/959 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

trenouf:
This passed CTS tests, but the result has not arrived back here.

PR #652, opened 2020-05-04
No user data layout for shader compile

trenouf:
Only the top two commits are for this PR. Other commits are from other unmerged PRs.

Building on those other PRs that enabled shader compilation without user data layout, this PR actually switches to doing shader compilation without user data layout. In LGC, relocs for user data are now only generated if there is no user data layout.

### DIFF
--- a/lgc/include/lgc/state/Defs.h
+++ b/lgc/include/lgc/state/Defs.h
@@ -139,5 +139,6 @@ static const unsigned InternalPerShaderTable = 0x10000001;
 static const unsigned DescRelocMagic = 0xA5A5A500;
 static const unsigned DescRelocMagicMask = 0xFFFFFF00;
 static const unsigned DescSetMask = 0x000000FF;
+static const unsigned DescRelocPushConst = 0xA5A5A600;
 
 } // namespace lgc
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -31,6 +31,7 @@
 
 #include "Gfx6Chip.h"
 #include "Gfx9Chip.h"
+#include "lgc/BuilderBase.h"
 #include "lgc/patch/Patch.h"
 #include "lgc/state/IntrinsDefs.h"
 #include "lgc/state/PalMetadata.h"
@@ -43,6 +44,7 @@
 #include "llvm/IR/IntrinsicsAMDGPU.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/MathExtras.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
@@ -130,6 +132,9 @@ private:
     SmallVector<UserDataNodeUsage, 8> descriptorSets;
     // Minimum offset at which spill table is used.
     unsigned spillUsage = UINT_MAX;
+    // Usage of vertex buffer table and streamout table
+    bool usesVertexBufferTable = false;
+    bool usesStreamOutTable = false;
   };
 
   // Gather user data usage in all shaders.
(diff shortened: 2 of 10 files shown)


lgc/include/lgc/state/Defs.h
```diff
@@ -140,5 +140,6 @@ static const unsigned InternalPerShaderTable = 0x10000001;
 static const unsigned DescRelocMagic = 0xA5A5A500;
 static const unsigned DescRelocMagicMask = 0xFFFFFF00;
 static const unsigned DescSetMask = 0x000000FF;
+static const unsigned DescRelocPushConst = 0xA5A5A600;
```
nhaehnle, line 143: I haven't followed up with this, but let's please not make things worse -- and since your moving of the linking code, we're now in a position to clean this up properly!

The right thing to do here is to extend PAL's Pal::Util::Abi::UserDataMapping. Let's define a range (0x8000000+?) for compiler-internal use, and then just reserve ranges starting from there, with proper thought given to how big those ranges should be (e.g., the descriptor set range should have 32 entries because that's the limit of descriptor sets per pipeline layout that the driver exposes).

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -454,18 +480,19 @@ void PatchEntryPointMutate::fixupUserDataUses(Module &module) {
...
         if (node) {
           byteOffset = builder.getInt32(node->offsetInDwords * 4);
           // Ensure we mark spill table usage.
           m_pipelineState->getPalMetadata()->setUserDataSpillUsage(node->offsetInDwords);
-        } else {
+        } else if (!m_pipelineState->isUnlinked()) {
           byteOffset = UndefValue::get(builder.getInt32Ty());
+        } else {
+          // Unlinked shader compilation: Use a reloc.
+          byteOffset = builder.CreateRelocationConstant("pushconst");
```
nhaehnle, line 494: Could you please add documentation on what relocation constants are used and what they mean?

lgc/include/lgc/state/Defs.h
trenouf: I have not moved the linker code yet. When I do, I intend to put these constants into a compiler/linker-private ABI file, and I agree we should put more thought into how they're defined. However I would like to do it in a future commit when I move the linker.

lgc/patch/PatchEntryPointMutate.cpp
trenouf: Yes. That's probably best in the new header file I'm proposing above (which can also define const strings for ones like this).

trenouf:
Looks like spurious failure on one CTS agent out of three.

trenouf:
> For standalone compiler, if we just have a shader stage to compile, how will you treat it since the user data layout info is no longer generated now? What we expect is trying to fill the missing info and make it work as the layout is specified if we do have such requirements (maybe via an command line option).

With this commit, amdllpcing a single shader will generate relocs for user data offsets. Do you mean you want to restore the previous auto-layout behavior with a command-line option? I can do that, but why do we want it?

amdrexu:
> I can do that, but why do we want it?

This is just to compare the compilation results with internal compiler. Because at the beginning of HW bring-up and extension development, we just hand-write simple GLSL shaders to investigate the compilation of certain new features. We don't have a full pipeline file from a real App test case at that time. Of course, this is not a high priority item but is useful.

trenouf:
> > I can do that, but why do we want it?
> 
> This is just to compare the compilation results with internal compiler. Because at the beginning of HW bring-up and extension development, we just hand-write simple GLSL shaders to investigate the compilation of certain new features. We don't have a full pipeline file from a real App test case at that time. Of course, this is not a high priority item but is useful.

I have now changed it so that -auto-desc-layout gets the old behavior.

amdvlk-admin:
### Test summary for commit 2945bdb240547cdccbdcdc0a6ae35b8d431c13c2

</ul></details>

<li>Passed: 959/959 (100.0%)</li>
<li>Failed: 0/959 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

trenouf:
Nicolai, are you ok to unblock this on the basis that I'm going to do the two things you said on a subsequent commit that moves linking into LGC?

trenouf:
Rebased

amdvlk-admin:
### Test summary for commit e07628099ed669adc2f5f25955754fbbaa489d6c

</ul></details>

<li>Passed: 959/959 (100.0%)</li>
<li>Failed: 0/959 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle:
> Nicolai, are you ok to unblock this on the basis that I'm going to do the two things you said on a subsequent commit that moves linking into LGC?

Yes, I think that's acceptable.

PR #660, opened 2020-05-08
Fix assertions on dEQP-VK.ssbo.phys.layout.random.descriptor_indexing.*

xuechen417:
- There are assertions of unsuported calls on dEQP-VK.ssbo.phys.layout.random.descriptor_indexing.3/11.
- Issue description:
OpDecorate %207 NonUniform
OpDecorate %249 NonUniform
%207 = OpLoad %mat3v4float %206 Aligned 16
//A dummy instruction is added after transShaderDecoration()
call void @spirv.NonUniform.V([3 x <4 x float>] %43)
%249 = OpLoad %mat4v4float %248 Aligned 16
//The first dummy intruction is changed after add the second dummy instruction in transShaderDecoration()
call void @2([3 x <4 x float>] %43)
call void @spirv.NonUniform.V([3 x <4 x float>] %91)
- The failing tests have arrayType varaibles with different element count decorated by NonUniform. In transShaderDecoration(), appendTypeMangling just mangles arrayType to V and ignores the element count and types. Hence, when create CallInst for the second decoration, it takes the name of the first one and make the first CallInst unsupported call.
- Fix: we modify printTypeName() to handle arrayType and structType in a finner scale like getTypeName() in lgc.

### DIFF
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -5183,11 +5183,30 @@ Function *SPIRVToLLVM::transFunction(SPIRVFunction *bf) {
 
 // Prints LLVM-style name for type to raw_ostream
 static void printTypeName(Type *ty, raw_ostream &nameStream) {
-  if (auto ptrTy = dyn_cast<PointerType>(ty)) {
-    nameStream << "p";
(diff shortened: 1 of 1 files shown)


amdrexu:
The name mangling is too coarse. It treats void/array/struct the same and qualifies them with "V". The change is follow what have been done in getTypeName() by Tim.

PR #675, opened 2020-05-11
Set the afn (ApproxFunc) fast-math flag for float operations.

jayfoad:
This is documented as allowing approximate calculations for intrinsics
like sin, log, sqrt etc. In practice it will allow generic middle-end
LLVM IR passes to transform pow(x,y) to exp(y*log(x)) and constant-fold
log(x) when x is a constant, and this doesn't decrease accuracy at all
because AMDGPU backend codegen implements pow with log and exp anyway.

It also allows the backend instruction selector to implement division as
multiplication by the reciprocal. This causes problems for some Vulkan
CTS tests which expect accurate double-precision results, so only set
this flag for float and not double.


llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1198,24 +1198,37 @@ Value *SPIRVToLLVM::transConvertInst(SPIRVValue *bv, Function *f, BasicBlock *bb
...
-      // AllowRessociation should be same with AllowContract
-      fmf.setAllowReassoc(allowContract);
-    }
-    // Enable "no NaN" and "no signed zeros" only if there isn't any floating point control flags
-    if (m_fpControlFlags.U32All == 0) {
-      fmf.setNoNaNs();
-      fmf.setNoSignedZeros(allowContract);
-    }
+  fmf.setAllowReciprocal();
+  if (ty->isTypeFloat(32))
```
trenouf, line 1215: This also does not allow afn for half-float. Is that intended? (Also brace this as it is multi-line including the comment.)

nhaehnle: We don't have fp16 transcendental operations anyway. Since approximate fp32 versions of transcendentals should have sufficient precision when applied to fp16 inputs, we should ensure that any backend lowering to fp32 applies the appropriate flags, but I think the right place to do that is in LLVM rather than here in the frontend.

nhaehnle: Huh, actually that's wrong -- we do have fp16 assembler instructions for some transcendentals,
but I believe those are of sufficient accuracy anyway.

PR #680, opened 2020-05-12
Fix the test

piotrAMD:
Make the test not check for explicit node id as
it is bound to change due to backend changes.


trenouf:
Looks like a spurious failure in one CTS run.

PR #686, opened 2020-05-14
[CI] Continue running remaining jobs despite failures.

kuhar:
This disables the default fail-fast behavior which kills all
github actions jobs after a single one fails. The motivation is to
better diagnose failures in the presence of `-Werror` and broken
trunk, like in
https://github.com/GPUOpen-Drivers/llpc/commit/ff90ae47205965cd9f8152ed86da6edeb081f41f.

For example, this allows all gcc builds to continue and report final
status even when clang builds fail.
Relevant docs:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast.

### DIFF
--- a/.github/workflows/build-amdvlk-docker.yml
+++ b/.github/workflows/build-amdvlk-docker.yml
@@ -10,6 +10,7 @@ jobs:
     if: github.repository == 'GPUOpen-Drivers/llpc'
     runs-on: ${{ matrix.host-os }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

kuhar:
This should be safe to merge without running CTS as no source/build files are changed.

kuhar:
On our end, we don't build with `-Werror` and use the `dev` branch, so this is not an issue at all. Fixing this CI is quick and easy :)

PR #704, opened 2020-05-20
Fix Game rage2 corruption

jiaolu:
Llpc assume "No Nan" optimization of fastmathflag when float control flag is not set,
therefore IsNan is removed in the later pass of "Combine redundant instructions".

In the case when spirv used IsNan op, the app writer already assume there could be occasion of float nan data.
so we can not set "No Nan" fastmathflag.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -306,6 +306,7 @@ struct ShaderModuleUsage {
   bool useHelpInvocation;      ///< Whether fragment shader has helper-invocation for subgroup
   bool useSpecConstant;        ///< Whether specializaton constant is used
   bool keepUnusedFunctions;    ///< Whether to keep unused function
+  bool useIsNan;               ///< Whether IsNan is used
 };
 
 /// Represents common part of shader module data
--- a/llpc/test/shaderdb/gfx9/ExtShaderFloat16_TestCommonFuncs_lit.frag
+++ b/llpc/test/shaderdb/gfx9/ExtShaderFloat16_TestCommonFuncs_lit.frag
@@ -49,25 +49,25 @@ void main()
 ; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIRV-to-LLVM translation results
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIR-V lowering results
-; SHADERTEST: = call reassoc nnan nsz arcp contract afn <3 x half> @llvm.fabs.v3f16(<3 x half>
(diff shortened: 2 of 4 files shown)


amdvlk-admin:
### Test summary for commit a1a12d8a5de7141ec7459a3b0e6cdc25091fce1a

<li>PAL: dev 85170384d0ee5447ac267d478f7a21bc036ea71c</li>
<li>SPVGEN: dev 704f4adddb6867d633431c2c9c2c0bfa6d294ce2</li>
<li>XGL: dev 9c4fe791366e730131a0d99a7519855068f216ef</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev c57487b76215ee74b1038039f47b210a28cad65e</li>
</ul></details>

<li>Failed: 0/960 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

jayfoad:
How about extending the disableFastMath hack in SpirvLowerAlgebraTransform to disable (some or all) fast math flags only on instructions that actually feed into an isnan call?

I agree that "translate OpIsNaN to something else that LLVM doesn't remove" would probably be the best fix, because it would match the SPIR-V semantics where isnan can reliably detect a nan result even from an instruction that's marked as not needing to handle nans properly.

jayfoad:
> translate OpIsNaN to something else that LLVM doesn't remove

We already have the llvm.amdgcn.class.f32 intrinsic that can detect nans. I wonder if that could be the "something else"?

jiaolu:
> llvm.amdgcn.class.f32

OpIsNan is translated to the llvm.amdgcn.class.f32, which is just the instruction removed in the "Combine redundant instructions"

jiaolu:
> SpirvLowerAlgebraTransform

this would be more complicated in implementation because isnan instruction itself does not have fastmathflag,  it's float instructions feeded to the isnan has no-nan fastmathflag.

jayfoad:
> > llvm.amdgcn.class.f32
> 
> OpIsNan is translated to the llvm.amdgcn.class.f32, which is just the instruction removed in the "Combine redundant instructions"

Maybe we could completely remove this part of `InstCombiner::visitCallInst`:
```
    if (Mask == (S_NAN | Q_NAN)) {
      // Equivalent of isnan. Replace with standard fcmp.
      Value *FCmp = Builder.CreateFCmpUNO(Src0, Src0);
      FCmp->takeName(II);
      return replaceInstUsesWith(*II, FCmp);
    }
```
I wonder if that would cause any other code quality problems?

> > SpirvLowerAlgebraTransform
> 
> this would be more complicated in implementation because isnan instruction itself does not have fastmathflag, it's float instructions feeded to the isnan has no-nan fastmathflag.

Yes, but `SpirvLowerAlgebraTransform::disableFastMath` already does that for you.

jiaolu:
> Yes, but SpirvLowerAlgebraTransform::disableFastMath already does that for you.
we still need to find OpIsNan to apply disableFastMath in lower pass, and disableFastMath itself is more computation intensive.

PR #728, opened 2020-05-28
Don't leave holes in PS input metadata

s-perron:
If there are missing entried for PS_INPUT_CNTL then the shader do not
work.  We need to make sure that if we use PS_INPUT_CNTL_<N> then we
also have an entry for PS_INTPU_CNTL_<M> for all M < N.

### DIFF
--- a/lgc/patch/Gfx9ConfigBuilder.cpp
+++ b/lgc/patch/Gfx9ConfigBuilder.cpp
@@ -1901,8 +1901,11 @@ void ConfigBuilder::buildPsRegConfig(ShaderStage shaderStage, T *pConfig) {
 
   for (unsigned i = 0; i < interpInfo->size(); ++i) {
     auto interpInfoElem = (*interpInfo)[i];
-    if (m_pipelineState->isUnlinked() && interpInfoElem.loc == InvalidFsInterpInfo.loc)
+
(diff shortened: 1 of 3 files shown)


amdvlk-admin:

lgc/patch/Gfx9ConfigBuilder.cpp
```diff
@@ -1901,8 +1901,6 @@ void ConfigBuilder::buildPsRegConfig(ShaderStage shaderStage, T *pConfig) {
 
   for (unsigned i = 0; i < interpInfo->size(); ++i) {
     auto interpInfoElem = (*interpInfo)[i];
-    if (m_pipelineState->isUnlinked() && interpInfoElem.loc == InvalidFsInterpInfo.loc)
```
trenouf, line 1904: 1. Is this a problem only because you're disabling packing on a shader compile?
2. Is the same fix needed in Gfx6ConfigBuilder.cpp (for gfx6,7,8)?

s-perron: 1) Yes, it is because we disable packing.  
2) Possibly I'll check.

JaxLinAMD:
retest this please

amdrexu:
The packing causes some game regression, so we disable this. After fixing the problems, packing will be re-enabled as the default mode.

JaxLinAMD:
retest this please

PR #729, opened 2020-05-29
Fix clang-10 warning

Flakebi:
lgc/elfLinker/ElfLinker.cpp:120:29: error: class with destructor marked 'final' cannot be inherited from [-Werror,-Wfinal-dtor-non-final-class]
  ~ElfLinkerImpl() override final;
                            ^


JaxLinAMD:
jenkins pipeline will be slow because of the maintenance.

JaxLinAMD:
retest this please

PR #731, opened 2020-05-29
Fix the user data limit calculation when linking

s-perron:
The limit must extend to the end of the descriptor not just to the start.


JaxLinAMD:
retest this please

PR #780, opened 2020-06-24
Explicitly include "llvm/Support/MemoryBuffer.h" for MemoryBufferRef

jayfoad:
Without this, this upstream LLVM commit b9a15c245f2 will break the
build.


trenouf:
I will merge this.

PR #870, opened 2020-08-03
NGG: Clarify the use of ES-GS offsets

amdrexu:
1. For merged shader, ES-GS offsets are packed together: 0,1 -> 01,
   2,3 -> 23, 4,5 -> 45. Recode the unpacked values for later uses.
   Also, replace the unpacking operation (ubfe) with and+shift.

2. The vertex indices within NGG subgroup are calculated from
   dividing ES-GS offset by ES-GS ring item size. Originally, the
   size was always 4, so we right-shift 2. But in the future,
   ES-GS ring item could be enlarged.

3. Add a helper function CreateUBfe to do bitfield extraction.

Change-Id: Ibdaf3fd21e3399d678d5baaff83a97b306050999

### DIFF
--- a/lgc/patch/NggPrimShader.cpp
+++ b/lgc/patch/NggPrimShader.cpp
@@ -544,8 +544,8 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
 
       initWaveThreadInfo(mergedGroupInfo, mergedWaveInfo);
 
-      // Record ES-GS vertex offsets info
-      m_nggFactor.esGsOffsets01 = esGsOffsets01;
+      // Record primitive connectivity data
+      m_nggFactor.primData = esGsOffsets01;
 
       if (distributePrimId) {
         auto primValid = m_builder->CreateICmpULT(m_nggFactor.threadIdInWave, m_nggFactor.primCountInWave);
@@ -572,7 +572,7 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
         // Distribute primitive ID
         auto vertexId0 =
             m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                       {m_nggFactor.esGsOffsets01, m_builder->getInt32(0), m_builder->getInt32(9)});
+                                       {m_nggFactor.primData, m_builder->getInt32(0), m_builder->getInt32(9)});
 
         unsigned regionStart = m_ldsManager->getLdsRegionStart(LdsRegionDistribPrimId);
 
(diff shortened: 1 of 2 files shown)


lgc/patch/NggPrimShader.cpp
```diff
@@ -1756,9 +1735,24 @@ void NggPrimShader::constructPrimShaderWithGs(Module *module) {
     m_nggFactor.primShaderTableAddrHigh = primShaderTableAddrHigh;
 
     // Record ES-GS vertex offsets info
-    m_nggFactor.esGsOffsets01 = esGsOffsets01;
-    m_nggFactor.esGsOffsets23 = esGsOffsets23;
-    m_nggFactor.esGsOffsets45 = esGsOffsets45;
+    auto esGsOffset0 = m_builder->CreateAnd(esGsOffsets01, 0xFFFF);
+    auto esGsOffset1 = m_builder->CreateLShr(esGsOffsets01, 16);
+    esGsOffset1 = m_builder->CreateAnd(esGsOffset1, 0xFFFF);
```
ruiling, line 1740: I think it is better to move the three lines into a small function. Something like "void getLoHiOfInt32(Value *, Value **, Value **)"? Then you can pass in the addresses of m_nggFactor.esGsOffsetX as the arguments.

lgc/patch/NggPrimShader.cpp
```diff
@@ -953,15 +959,12 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
...
         //   ES_GS_OFFSET01[15:0]  = vertexId0 (in dwords)
 
-        // Use vertex0 as provoking vertex to distribute primitive ID
-        auto esGsOffset0 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                                      {
-                                                          m_nggFactor.esGsOffsets01,
-                                                          m_builder->getInt32(0),
-                                                          m_builder->getInt32(16),
-                                                      });
+        const auto esGsRingItemSize =
```
ruiling, line 962: I think 'unsigned' is much clearer than 'auto' here, so we can easily know it is a constant value.

lgc/patch/NggPrimShader.cpp
```diff
@@ -1208,29 +1211,13 @@ void NggPrimShader::constructPrimShaderWithoutGs(Module *module) {
...
-      vertexId1 = m_builder->CreateLShr(vertexId1, 2);
-
-      Value *vertexId2 = m_builder->CreateIntrinsic(Intrinsic::amdgcn_ubfe, m_builder->getInt32Ty(),
-                                                    {
-                                                        m_nggFactor.esGsOffsets23,
-                                                        m_builder->getInt32(0),
-                                                        m_builder->getInt32(16),
-                                                    });
-      vertexId2 = m_builder->CreateLShr(vertexId2, 2);
+      const auto esGsRingItemSize =
```
ruiling, line 1214: 'unsigned'

lgc/patch/NggPrimShader.cpp
amdrexu: OK. I will correct this.

lgc/patch/NggPrimShader.cpp
amdrexu: OK. I will add a helper function to handle this.

lgc/patch/NggPrimShader.cpp
amdrexu: Will be corrected.

lgc/patch/NggPrimShader.cpp
amdrexu: I add a helper CreateUBfe to do bitfield extraction because it is a special case to get low part and high part from a 32-bit integer. Actually, we need more general uses in this source file and we have several code pieces to replace amdgcn_ubfe (will make a future commit to do this).

ruiling: sounds ok to me.

amdrexu:
retest this please

Issue #879, opened 2020-08-06
[CI] Sanitizer AMDVLK builds fail because of insufficient disk space


kuhar:
Started failing 7 days ago, the first failure was https://github.com/GPUOpen-Drivers/llpc/runs/927523361?check_suite_focus=true:
```
[2/2] Linking CXX executable compiler/llpc/llvm/bin/not
Error processing tar file(exit status 1): write /vulkandriver/builds/ci-build/pal/src/libpal.a: no space left on device
```

This fails very late in the build. We should reduce disk usage or disable them. Other build configurations are unaffected.

kuhar:
Sometimes fails with different files:
```
[250/250] Linking CXX shared library spvgen/spvgen.so
[1/2] Building CXX object compiler/llpc/llvm/utils/FileCheck/CMakeFiles/FileCheck.dir/FileCheck.cpp.o
[2/2] Linking CXX executable compiler/llpc/llvm/bin/FileCheck
[1/2] Building C object compiler/llpc/llvm/utils/count/CMakeFiles/count.dir/count.c.o
[2/2] Linking C executable compiler/llpc/llvm/bin/count
[1/2] Building CXX object compiler/llpc/llvm/utils/not/CMakeFiles/not.dir/not.cpp.o
[2/2] Linking CXX executable compiler/llpc/llvm/bin/not
Error processing tar file(exit status 1): write /vulkandriver/builds/ci-build/spvgen/spvgen.so: no space left on device
##[error]Process completed with exit code 1.
```

This only affects builds with assertions on and sanitizers; sanitizers and no assertions work fine.

kuhar:
I was watching the builds and it seems to me that too many llvm tools get built. It may be worth seeing if we can exclude them in CMake.
```
[2379/2512] Generating ../../bin/llvm-lib
[2380/2512] Generating ../../bin/llvm-dlltool
[2381/2512] Generating ../../bin/llvm-ranlib
[2382/2512] Linking CXX executable compiler/llpc/llvm/bin/llvm-nm
[2383/2512] Building CXX object compiler/llpc/llvm/lib/Target/AMDGPU/CMakeFiles/LLVMAMDGPUCodeGen.dir/SIISelLowering.cpp.o
[2384/2512] Building CXX object compiler/llpc/llvm/lib/Target/AMDGPU/Disassembler/CMakeFiles/LLVMAMDGPUDisassembler.dir/AMDGPUDisassembler.cpp.o
[...]

PR #883, opened 2020-08-10
[CI] Don't checkout full repos in github workflows

kuhar:
Save some disk space (34MiB) by keeping only the 'docker' subdirectory.

An alternative solution would be to do a sparse checkout, but
this requires very recent git versions and doesn't seem well
documented.

Bug: https://github.com/GPUOpen-Drivers/llpc/issues/879


kuhar:
Closing in favor of a bigger cleanup

PR #885, opened 2020-08-10
Extend descriptor set pointer to 64 bits at start of shader

trenouf:
This restores the behavior from before #650 was merged:
Rationalized user data usage

When a descriptor set pointer is extended from 32 to 64 bit using the
high word of PC, that code is now inserted at the start of the shader
(or just after the load, if it was spilled). That seems to cause less
confusion as to whether it is divergent, and saves lots of flat_loads
and v_readfirstlanes in a case I was looking at where image and sampler
descrioptors were first used after some control flow.

Change-Id: I164827622f6a493d39768b9be754342c615507ea


trenouf:
That failure was just running out of disk space.

PR #899, opened 2020-08-14
Fix VFX issues when parsing pipeline file dumped from driver

amdrexu:
- disableLoopUnroll is missing in ShaderOption
- memberCount of ShaderOption is wrong. Add two. One is for
  disableLoopUnroll. The other is for internally reserved.
- ShadowDescriptorTableUsage enumerants are not mapped to strings and
  VFX is unable to recognize them.

Change-Id: I83755f14846a985ce09483afe52e872fd54217e7


amdrexu:
retest this please

PR #905, opened 2020-08-20
Fix user data handling when reading off end of push const

trenouf:
This fixes a compiler crash introduced in #650
68414fe49 Rationalized user data usage

The crash happened when a shader uses a larger push constant than the
size indicated in the pipeline state.

Change-Id: Iec7ef011e89ed199b34609896699881b502b59c4

### DIFF
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -1057,7 +1057,7 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
       break;
     }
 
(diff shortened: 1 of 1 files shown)


JaxLinAMD:
reest this please

JaxLinAMD:
retest this please

JaxLinAMD:
retest this please

nhaehnle (commented):
Isn't that illegal according to the spec? As in, shouldn't that fail validation? We probably shouldn't just compile this without complaining loudly about it...

trenouf:
Yes it sounds like it is illegal. But we have at least one real game doing it, and it works on our closed source compiler, and on LLPC before #650.

PR #943, opened 2020-09-02
Revert "Fix-ups for now deprecated getNumElements on VectorType"

trenouf:
That change appeared to be causing CTS failures in internal testing.

This reverts commit 6107115c6a63e291a1b84bc175914d65efb96715.


trenouf:
Scratch that -- Maciej has confirmed that it was the LLVM upgrade that caused those failures.

PR #958, opened 2020-09-09
Remove GraphicsFuzz_ReplaceAllUsesWithNotValid.spvasm no longer works

dstutt:
Removing this test as a more recent LLVM fails to exhibit the issue. I think
that the issue is probably still there, but hidden by changes upstream.

Since there's already an issue filed for this (with a reproducer for the noted
versions of the source), I'm going to remove the test to allow merges and
promotions.

My assumption is that a test will be reinstated (that passes) once the original
issue is resolved.


amdrexu (approved):

trenouf:
I will merge without letting testing complete, as all it does is remove one test.

PR #990, opened 2020-10-05
Add pass that inserts stack bounds checks

Flakebi:
Some applications access stack objects out of bounds, leading to
segfaults and hangs.

The pass checks the upper bound of dynamic array indices. Out of bounds
stores will be omitted, loads return zero.

This pass replaces the `enable-scratch-bounds-checks` option of llvm.

Open question:
When should this option be activated? `enable-scratch-bounds-checks` is currently always enabled, though probably it should only be enabled for pipelines or apps that need it. For what it’s worth, I cannot reproduce the original issue anymore that lead to the implementation of `enable-scratch-bounds-checks` (it runs fine with disabled bounds checks).

Should this option also have a corresponding command line option for amdllpc?

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -70,6 +70,7 @@
 //* %Version History
 //* | %Version | Change Description                                                                                    |
 //* | -------- | ----------------------------------------------------------------------------------------------------- |
+//* |     43.2 | Add enableScratchBoundsCheck in PipelineOptions                                                       |
 //* |     43.1 | Add disableImageResourceCheck in PipelineOptions                                                      |
 //* |     43.0 | Removed the enumerant WaveBreakSize::DrawTime                                                         |
 //* |     42.0 | Removed tileOptimal flag from SamplerYcbcrConversion metadata struct                                  |
@@ -328,6 +329,8 @@ struct PipelineOptions {
                                    ///  then linking them, when possible.  When not possible this option is ignored.
   bool disableImageResourceCheck;  ///< If set, the pipeline shader will not contain code to check and fix invalid image
                                    ///  descriptors.
+  bool enableScratchBoundsCheck;   ///< If set, accessing a stack object out of bounds will be skipped for stores and
+                                   ///  return zero for loads.
   ShadowDescriptorTableUsage shadowDescriptorTableUsage; ///< Controls shadow descriptor table.
   unsigned shadowDescriptorTablePtrHigh;                 ///< Sets high part of VA ptr for shadow descriptor table.
   ExtendedRobustness extendedRobustness;                 ///< ExtendedRobustness is intended to correspond to the
--- a/lgc/patch/PatchBoundsCheckMemory.cpp
+++ b/lgc/patch/PatchBoundsCheckMemory.cpp
@@ -1,0 +1,226 @@
+/*
+ ***********************************************************************************************************************
+ *
+ *  Copyright (c) 2020 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in all
+ *  copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ *
+ **********************************************************************************************************************/
+/**
+ ***********************************************************************************************************************
+ * @file  PatchBoundsCheckMemory.cpp
+ * @brief LLPC source file: contains implementation of class lgc::PatchBoundsCheckMemory.
+ *
+ * This pass adds bounds checks to all stack/scratch accesses with dynamic indices. The pass looks at every
+ * getelementptr instruction, checking which of the used indices are not constants. For non-constant indices, it checks
+ * if they index into either a FixedVectorType or an ArrayType and gets their element count, which is the upper bound of
+ * the index.
+ *
+ * Example:
+ * We have a getelementptr, followed by a load.
+ *
+ * %elemPtr = getelementptr [16 x float], [16 x float] addrspace(5)* %array, i32 0, i32 %index
+ * %value = load float, float addrspace(5)* %elemPtr, align 4
+ *
+ * Behind the getelementptr, we will compute the condition if all dynamic indices are in-bounds:
+ *
+ * %elemPtr = getelementptr [16 x float], [16 x float] addrspace(5)* %array, i32 0, i32 %index
+ * %inBounds = icmp ult i32 %index, 16
+ *
+ * %value = load float, float addrspace(5)* %elemPtr, align 4
+ *
+ * We look at all users of the getelementptr and guard loads and stores with the in-bounds condition. Stores are skipped
+ * if out-of-bounds, loads will return zero.
+ *
+ * %elemPtr = getelementptr [16 x float], [16 x float] addrspace(5)* %array, i32 0, i32 %index
+ * %inBounds = icmp ult i32 %index, 16
+ *
+ * br i1 %inBounds, label %inBoundsBB, label %continueBB
+ *
+ * inBoundsBB:
+ * %loadValue = load float, float addrspace(5)* %elemPtr, align 4
+ * br label %continueBB
+ *
+ * continueBB:
+ * %value = phi float [ %loadValue, %inBoundsBB ], [ 0.000000e+00, %.entry ]
+ *
+ ***********************************************************************************************************************
+ */
+#include "PatchBoundsCheckMemory.h"
+#include "lgc/state/IntrinsDefs.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/BasicBlockUtils.h"
+
+#define DEBUG_TYPE "lgc-bounds-check-memory"
+
+using namespace lgc;
+using namespace llvm;
+
+namespace lgc {
+
+// =====================================================================================================================
+// Initializes static members.
+char PatchBoundsCheckMemory::ID = 0;
+
+// =====================================================================================================================
+// Pass creator, creates the pass of bounds check memory operations.
+ModulePass *createPatchBoundsCheckMemory() {
+  return new PatchBoundsCheckMemory();
+}
+
+static Value *getZeroConstant(Type *ty) {
+  if (ty->isIntegerTy()) {
+    return ConstantInt::get(ty, 0);
+  }
+  if (ty->isFloatingPointTy()) {
+    return ConstantFP::get(ty, 0.0);
+  }
+  return ConstantAggregateZero::get(ty);
+}
+
+// =====================================================================================================================
+PatchBoundsCheckMemory::PatchBoundsCheckMemory() : Patch(ID) {
+}
+
+// =====================================================================================================================
+// Executes this pass on the specified LLVM module.
+//
+// @param [in/out] module : LLVM module to be run on
+bool PatchBoundsCheckMemory::runOnModule(Module &module) {
+  LLVM_DEBUG(dbgs() << "Run the pass llpc-bounds-check-memory\n");
+
+  m_pipelineState = getAnalysis<PipelineStateWrapper>().getPipelineState(&module);
+
+  if (!m_pipelineState->getOptions().enableScratchBoundsCheck)
+    return false;
+
+  m_context = &module.getContext();
+  m_builder = std::make_unique<IRBuilder<>>(*m_context);
+
+  visit(module);
+
+  bool changed = false;
+  for (auto &inst : m_getElementPtrInsts)
+    changed |= addBoundsCheck(inst);
+  m_getElementPtrInsts.clear();
+
+  return changed;
+}
+
+// =====================================================================================================================
+// Visits "getelementptr" instruction.
+//
+// @param getElemPtr : "GetElementPtr" instruction
+void PatchBoundsCheckMemory::visitGetElementPtrInst(GetElementPtrInst &getElemPtr) {
+  auto pointerTy = getElemPtr.getPointerOperand()->getType();
+  if (pointerTy->getPointerAddressSpace() != ADDR_SPACE_PRIVATE)
+    return;
+
+  // Search for dynamic indices in the instruction and collect them in a GetElemPtrInfo struct
+  llvm::SmallVector<std::pair<Value *, uint64_t>, 1> dynIndices;
+  std::vector<Value *> indices;
+  for (const auto &index : getElemPtr.indices()) {
+    if (!isa<Constant>(index)) {
+      auto indexedTy = GetElementPtrInst::getIndexedType(pointerTy->getPointerElementType(), indices);
+      // Save the index value its the upper bound
+      if (auto vectorTy = dyn_cast<FixedVectorType>(indexedTy)) {
+        dynIndices.push_back({index, vectorTy->getNumElements()});
+      } else if (auto arrayTy = dyn_cast<ArrayType>(indexedTy)) {
+        dynIndices.push_back({index, arrayTy->getNumElements()});
+      } else {
+        llvm_unreachable("Unsupported indexed type for bounds checking");
+      }
+    }
+    indices.push_back(index);
+  }
+  if (dynIndices.empty())
+    return;
+  m_getElementPtrInsts.push_back({&getElemPtr, dynIndices});
+}
+
+// =====================================================================================================================
+// Checks whether each dynamic index in the specified "getelementptr" instruction is lower than its allowed upper bound.
+//
+// @param info : "GetElementPtr" instruction and its dynamic indices
+// @returns : True if the code was changed or false if unmodified
+bool PatchBoundsCheckMemory::addBoundsCheck(GetElemPtrInfo &info) {
+  auto &getElemPtr = *info.getElemPtr;
+
+  // Copy users
+  std::vector<Instruction *> users;
+  for (auto user : getElemPtr.users()) {
+    Instruction *inst = dyn_cast<Instruction>(user);
+    if (!inst)
+      continue;
+    // Look through bitcasts
+    if (isa<BitCastInst>(user)) {
+      for (auto user : inst->users()) {
+        if (Instruction *inst = dyn_cast<Instruction>(user))
+          users.push_back(inst);
+      }
+    } else {
+      users.push_back(inst);
+    }
+  }
+
+  if (users.empty())
+    return false;
+
+  // Insert after getElemPtr
+  m_builder->SetInsertPoint(getElemPtr.getParent(), ++getElemPtr.getIterator());
+
+  // Get condition for in-bounds pointer
+  Value *inBounds = nullptr;
+  // Iterate through all dynamic indices and check that each one is in bounds
+  for (auto indexBoundPair : info.dynIndices) {
+    auto indexTy = cast<IntegerType>(indexBoundPair.first->getType());
+    auto curInBounds = m_builder->CreateCmp(CmpInst::ICMP_ULT, indexBoundPair.first,
+                                            m_builder->getIntN(indexTy->getScalarSizeInBits(), indexBoundPair.second));
+    inBounds = inBounds ? m_builder->CreateAnd(inBounds, curInBounds) : curInBounds;
+  }
+
+  // Only execute loads and stores if in-bounds
+  for (auto user : users) {
+    BasicBlock *instBB = user->getParent();
+    Instruction *inBoundsTerminator = SplitBlockAndInsertIfThen(inBounds, user, false);
+    user->moveBefore(inBoundsTerminator);
+    BasicBlock *inBoundsBB = inBoundsTerminator->getParent();
+    BasicBlock *continueBB = cast<BranchInst>(inBoundsTerminator)->getSuccessor(0);
+
+    if (auto loadInst = dyn_cast<LoadInst>(user)) {
+      // Load returns zero if out-of-bounds
+      auto loadTy = loadInst->getType();
+      m_builder->SetInsertPoint(&continueBB->front());
+      auto *phi = m_builder->CreatePHI(loadTy, 2);
+      loadInst->replaceAllUsesWith(phi);
+      phi->addIncoming(loadInst, inBoundsBB);
+      phi->addIncoming(getZeroConstant(loadTy), instBB);
+    } else if (!isa<StoreInst>(user)) {
+      llvm_unreachable("Expected a load or store instruction");
+    }
+  }
+  return true;
+}
+
+} // namespace lgc
+
+// =====================================================================================================================
+// Initializes the pass of bounds check memory operations.
+INITIALIZE_PASS(PatchBoundsCheckMemory, DEBUG_TYPE, "Patch LLVM for memory operation bounds checks", false, false)
--- a/lgc/patch/PatchBoundsCheckMemory.h
+++ b/lgc/patch/PatchBoundsCheckMemory.h
@@ -1,0 +1,82 @@
+/*
+ ***********************************************************************************************************************
+ *
+ *  Copyright (c) 2020 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in all
+ *  copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ *
+ **********************************************************************************************************************/
+/**
+ ***********************************************************************************************************************
+ * @file  PatchBoundsCheckMemory.h
+ * @brief LLPC header file: contains declaration of class lgc::PatchBoundsCheckMemory.
+ ***********************************************************************************************************************
+ */
+#pragma once
+
+#include "lgc/patch/Patch.h"
+#include "lgc/state/PipelineState.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/InstVisitor.h"
+#include <utility>
+#include <vector>
+
+namespace llvm {
+class GetElementPtrInst;
+class StoreInst;
+} // namespace llvm
+
+namespace lgc {
+
+// =====================================================================================================================
+// The structure for getelementptr instruction which need to be checked.
+struct GetElemPtrInfo {
+  llvm::GetElementPtrInst *getElemPtr;
+  // Dynamic index value and maximum for each index.
+  llvm::SmallVector<std::pair<llvm::Value *, uint64_t>, 1> dynIndices;
+};
+
+// =====================================================================================================================
+// Represents the pass of bounds check memory operations.
+class PatchBoundsCheckMemory final : public Patch, public llvm::InstVisitor<PatchBoundsCheckMemory> {
+public:
+  PatchBoundsCheckMemory();
+
+  void getAnalysisUsage(llvm::AnalysisUsage &analysisUsage) const override {
+    analysisUsage.addRequired<PipelineStateWrapper>();
+  }
+
+  bool runOnModule(llvm::Module &module) override;
+  void visitGetElementPtrInst(llvm::GetElementPtrInst &getElementPtrInst);
+
+  static char ID; // ID of this pass
+
+private:
+  PatchBoundsCheckMemory(const PatchBoundsCheckMemory &) = delete;
+  PatchBoundsCheckMemory &operator=(const PatchBoundsCheckMemory &) = delete;
+
+  bool addBoundsCheck(GetElemPtrInfo &info);
+
+  std::vector<GetElemPtrInfo> m_getElementPtrInsts; // The memory accesses to patch.
+  std::unique_ptr<llvm::IRBuilder<>> m_builder;
+  llvm::LLVMContext *m_context;
+  PipelineState *m_pipelineState;
+};
+
+} // namespace lgc
--- a/lgc/test/StackBoundsCheck.lgc
+++ b/lgc/test/StackBoundsCheck.lgc
@@ -1,0 +1,133 @@
+; RUN: lgc -mcpu=gfx900 -stop-after=lgc-bounds-check-memory - <%s | FileCheck %s
+
+; lgc code generated with amdllpc llpc/test/shaderdb/OpLoad_StackBoundsCheck_struct.pipe -emit-lgc
+
+; ModuleID = 'lgcPipeline'
+source_filename = "lgcPipeline"
+target datalayout = "e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-ni:7"
+target triple = "amdgcn--amdpal"
+
+; Function Attrs: nounwind
+define spir_func void @lgc.shader.VS.main() local_unnamed_addr #0 !spirv.ExecutionModel !10 !lgc.shaderstage !10 {
+.entry:
+  %0 = call <4 x float> (...) @lgc.create.read.generic.input.v4f32(i32 0, i32 0, i32 0, i32 0, i32 0, i32 undef)
+  call void (...) @lgc.create.write.builtin.output(<4 x float> %0, i32 0, i32 0, i32 undef, i32 undef)
+  call void (...) @lgc.create.write.builtin.output(float undef, i32 1, i32 0, i32 undef, i32 undef)
+  call void (...) @lgc.create.write.builtin.output([1 x float] undef, i32 3, i32 4096, i32 undef, i32 undef)
+  call void (...) @lgc.create.write.builtin.output([1 x float] undef, i32 4, i32 4096, i32 undef, i32 undef)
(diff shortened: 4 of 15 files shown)


amdvlk-admin:
### Test summary for commit 7df315093f491b4da368775c9e413ecc610e563c

<li>PAL: dev c9937b277c491a55c689eec7cdd48fb238b8004c</li>
<li>SPVGEN: dev e1f1881d84cd1c7a6c1d661cbac0e619b4e110c1</li>
<li>XGL: dev c6c90450fb1abf5f414acf1e38a0f51a72c426c1</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 0510352269ff92dabcb91c46b1ad30b61299a7e7</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/131664)</summary><ul><li>Built with branch Vulkan-CTS-1.2.0.2-fetch-sources-fix</li>

<li>Passed: 24006/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19889/43895 (45.3%)</li>
</ul></details>

<li>Passed: 24005/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19890/43895 (45.3%)</li>
</ul></details>

<li>Passed: 22647/43874 (51.6%)</li>
<li>Failed: 0/43874 (0.0%)</li>
<li>Skipped: 21227/43874 (48.4%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ *
+ **********************************************************************************************************************/
+/**
+ ***********************************************************************************************************************
+ * @file  PatchBoundsCheckMemory.cpp
+ * @brief LLPC source file: contains implementation of class lgc::PatchBoundsCheckMemory.
```
kuhar, line 28: Could you make the description more descriptive? This repeats both the file line above and the actual filename.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+ */
+#include "PatchBoundsCheckMemory.h"
+#include "lgc/state/IntrinsDefs.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+
+#define DEBUG_TYPE "lgc-bounds-check-memory"
+
+using namespace llvm;
+using namespace lgc;
```
kuhar, line 39: unused?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+// =====================================================================================================================
+// Pass creator, creates the pass of bounds check memory operations.
+ModulePass *createPatchBoundsCheckMemory() {
+  return new PatchBoundsCheckMemory();
+}
+
+static Value *getTypeZero(Type *ty) {
+  if (ty->isIntegerTy()) {
+    return ConstantInt::get(ty, 0);
+  } else if (ty->isFloatTy() || ty->isHalfTy() || ty->isDoubleTy() || ty->isFP128Ty() || ty->isBFloatTy()) {
```
kuhar, line 56: LLVM coding style doesn't use else after return.

kuhar: Would isFloatingPointType work here?
https://llvm.org/doxygen/classllvm_1_1Type.html#aac5759c83abd6a4af236401a7cfe7a0f

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+    return ConstantFP::get(ty, 0.0);
+  } else {
```
kuhar, line 58: Remove else. Can we assert that `ty` is an aggregate type?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+bool PatchBoundsCheckMemory::runOnModule(Module &module) {
+  LLVM_DEBUG(dbgs() << "Run the pass llpc-bounds-check-memory\n");
+
+  m_pipelineState = getAnalysis<PipelineStateWrapper>().getPipelineState(&module);
+
+  if (!m_pipelineState->getOptions().enableScratchBoundsCheck)
+    return false;
+
+  m_context = &module.getContext();
+  m_builder.reset(new IRBuilder<>(*m_context));
```
kuhar, line 80:
```suggestion
  m_builder = std::make_unique<IRBuilder<>>(*m_context);
```

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+// @param getElemPtr : "GetElementPtr" instruction
+void PatchBoundsCheckMemory::visitGetElementPtrInst(GetElementPtrInst &getElemPtr) {
+  auto pointerTy = getElemPtr.getPointerOperand()->getType();
+  if (pointerTy->getPointerAddressSpace() != ADDR_SPACE_PRIVATE)
+    return;
+
+  // Check for dynamic indices in the instruction
+  llvm::SmallVector<std::pair<Value *, uint64_t>, 1> dynIndices;
+  std::vector<Value *> indices;
+  for (unsigned i = 1, operandCount = getElemPtr.getNumOperands(); i < operandCount; ++i) {
```
kuhar, line 104: Can this iterate over `getElemPtr.indices()`?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+    auto index = getElemPtr.getOperand(i);
+    if (!isa<Constant>(index)) {
```
kuhar, line 106: can you flip this condition to avoid nesting the code so much?
i.e., `if (isa<Constant>(index)) continue;`

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+// =====================================================================================================================
+// Visits "getelementptr" instruction.
+//
+// @param getElemPtr : "GetElementPtr" instruction
+void PatchBoundsCheckMemory::visitGetElementPtrInst(GetElementPtrInst &getElemPtr) {
+  auto pointerTy = getElemPtr.getPointerOperand()->getType();
+  if (pointerTy->getPointerAddressSpace() != ADDR_SPACE_PRIVATE)
+    return;
+
+  // Check for dynamic indices in the instruction
```
kuhar, line 101: Can you describe what this loop does? 'Checking' is a very general term.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+  if (dynIndices.empty())
+    return;
+  m_getElementPtrInsts.push_back({&getElemPtr, dynIndices});
+}
+
+// =====================================================================================================================
+// Checks whether each dynamic index in the specified "getelementptr" instruction is lower than its allowed upper bound.
+//
+// @param info : "GetElementPtr" instruction and its dynamic indinces
+// @returns : True of the code was changed or false if unmodified
```
kuhar, line 128: typo: of -> if

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+  std::vector<User *> users;
+  for (auto user : getElemPtr.users()) {
+    // Look through bitcasts
+    Instruction *inst = cast<Instruction>(user);
+    if (isa<BitCastInst>(inst)) {
+      for (auto user : inst->users())
+        users.push_back(user);
+    } else {
+      users.push_back(user);
+    }
```
kuhar, line 142: Can you use `inst->stripPointerCasts()` here?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  if (users.empty())
+    return false;
+
+  // Insert after getElemPtr
+  m_builder->SetInsertPoint(getElemPtr.getParent(), ++getElemPtr.getIterator());
+
+  // Get condition for out-of-bounds
```
kuhar, line 151: ... for out-of-bounds access?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+  Value *inBounds = nullptr;
+  for (auto p : info.dynIndices) {
+    auto indexTy = cast<IntegerType>(p.first->getType());
+    auto curInBounds =
+        m_builder->CreateCmp(CmpInst::ICMP_ULT, p.first, m_builder->getIntN(indexTy->getScalarSizeInBits(), p.second));
+    if (inBounds)
+      inBounds = m_builder->CreateAnd(inBounds, curInBounds);
+    else
+      inBounds = curInBounds;
```
kuhar, line 160: nit: I'd use the ternary operator.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  if (users.empty())
+    return false;
+
+  // Insert after getElemPtr
+  m_builder->SetInsertPoint(getElemPtr.getParent(), ++getElemPtr.getIterator());
+
+  // Get condition for out-of-bounds
+  Value *inBounds = nullptr;
+  for (auto p : info.dynIndices) {
```
kuhar, line 153: can you rename `p` to something like `somethingSomethingPair`?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+        m_builder->CreateCmp(CmpInst::ICMP_ULT, p.first, m_builder->getIntN(indexTy->getScalarSizeInBits(), p.second));
+    if (inBounds)
+      inBounds = m_builder->CreateAnd(inBounds, curInBounds);
+    else
+      inBounds = curInBounds;
+  }
+
+  // Only execute loads and stores if in-bounds
+  for (auto user : users) {
+    Instruction *inst = cast<Instruction>(user);
```
kuhar, line 165: You can avoid the casts by making users a vector of instructions.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+    auto instBB = inst->getParent();
+    auto inBoundsBB = instBB->splitBasicBlock(inst);
+    auto nextInst = ++inst->getIterator();
+    BasicBlock *continueBB = inBoundsBB->splitBasicBlock(nextInst);
```
kuhar, line 169: I think there should be an llvm function that splits a BB and adds a branch for the given condition, no?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    instBB->getTerminator()->eraseFromParent();
+    BranchInst::Create(inBoundsBB, continueBB, inBounds, instBB);
+
+    auto loadInst = dyn_cast<LoadInst>(user);
+    auto storeInst = dyn_cast<StoreInst>(user);
+    if (loadInst) {
```
kuhar, line 176: put these variable in if conditions

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+      auto loadTy = loadInst->getType();
+      auto *phi = PHINode::Create(loadTy, 2, "", &continueBB->front());
+      loadInst->replaceAllUsesWith(phi);
+      phi->addIncoming(loadInst, inBoundsBB);
+      phi->addIncoming(getTypeZero(loadTy), instBB);
+    } else if (storeInst) {
+    } else {
```
kuhar, line 183: ?

lgc/patch/PatchBoundsCheckMemory.cpp
kuhar: After going through the code, a diagram with how the instrumented IR looks like it would be useful to have either here or above the function that emits the instrumentation.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+    indices.push_back(index);
+  }
+  if (dynIndices.empty())
+    return;
+  m_getElementPtrInsts.push_back({&getElemPtr, dynIndices});
+}
+
+// =====================================================================================================================
+// Checks whether each dynamic index in the specified "getelementptr" instruction is lower than its allowed upper bound.
```
kuhar, line 125: What if the index is just one past the size of the upper bound but the pointer is never accessed?

llpc/test/shaderdb/OpLoad_StackBoundsCheck.pipe
```diff
@@ -0,0 +1,123 @@
...
+
+layout(location = 0) in vec4 color_in;
+layout(location = 0) out vec4 color_out;
+
+void main() {
+  vec2 localData[16];
+  for (int i = 0; i < 16; i++)
+    localData[i] = vec2(color_in.x + i + 5);
+  // Read out-of-bounds of the stack array
+  float v = localData[int(color_in.x * 255) + 5024].x;
```
kuhar, line 48: Just based on this test, it's unclear to me if this will insert one check in the loop (times unroll count) and one outside, or just outside. Can you add some comments that relates the checks to the code, or perhaps put the checks just above the code they correspond to?

llpc/test/shaderdb/OpLoad_StackBoundsCheck_struct.pipe
```diff
@@ -0,0 +1,137 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} pipeline patching results
+
+; SHADERTEST: = icmp ult i32 %[[#]], -16
```
kuhar, line 5: Why -16?

kuhar (changes requested):
Would it be possible to create lgc tests for this?
Based on the existing tests and the code, it's difficult to tell how the instrumentation looks like.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: It seems to be used by the `INITIALIZE_PASS` in the end.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: Right, thanks!

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: There is such an assertion right at the top of `ConstantAggregateZero::get`. I’d rather not copy it here because it checks for three different types.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: The `break` at the end of the if should not have been there, so I can’t do that without duplicating the `indices.push_back`.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: I want to look through bitcasts of the result of getelementptr/the users. `stripPointerCasts` does look at the used inputs to the instruction, which is the wrong direction as far as I understand.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: Oh there is, thanks for the hint.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: We only *change* the behavior of load and store instructions. If there are no loads or stores, the `inBounds` condition might be false but nothing changes.

llpc/test/shaderdb/OpLoad_StackBoundsCheck_struct.pipe
Flakebi: llvm rearranges the code that it skips the load if `int(color_in.y * 255) < -16`.

llpc/test/shaderdb/OpLoad_StackBoundsCheck.pipe
Flakebi: For the loop, llvm can statically show that all accesses are in-bounds so it will remove the checks. The only check left is the one that is actually out-of-bounds.
I think most (all?) shaderdb tests have their checks at the top or bottom of the file, should I move them around anyway?

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: Not a diagram but I added examplary IR with an explanation.

amdvlk-admin:
### Test summary for commit 3644f8902c8d67c8c3aaef181575b333dbfe9a42

<li>PAL: dev c9937b277c491a55c689eec7cdd48fb238b8004c</li>
<li>SPVGEN: dev e1f1881d84cd1c7a6c1d661cbac0e619b4e110c1</li>
<li>XGL: dev c6c90450fb1abf5f414acf1e38a0f51a72c426c1</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev bf987708faa5335d515b11a05d898442674fd6ed</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/131664)</summary><ul><li>Built with branch Vulkan-CTS-1.2.0.2-fetch-sources-fix</li>

<li>Passed: 24006/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19889/43895 (45.3%)</li>
</ul></details>

<li>Passed: 24005/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19890/43895 (45.3%)</li>
</ul></details>

<li>Passed: 22647/43874 (51.6%)</li>
<li>Failed: 0/43874 (0.0%)</li>
<li>Skipped: 21227/43874 (48.4%)</li>
</ul></details>


</ul></details>

tool/dumper/vkgcPipelineDumper.cpp
```diff
@@ -665,6 +665,7 @@ void PipelineDumper::dumpPipelineOptions(const PipelineOptions *options, std::os
   dumpFile << "options.scalarBlockLayout = " << options->scalarBlockLayout << "\n";
   dumpFile << "options.includeIr = " << options->includeIr << "\n";
   dumpFile << "options.robustBufferAccess = " << options->robustBufferAccess << "\n";
+  dumpFile << "options.robustBufferAccess = " << options->enableScratchBoundsCheck << "\n";
```
perlfu, line 668: Presumably duplicate "options.robustBufferAccess" should be "options.enableScratchBoundsCheck"?

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,226 @@
...
+//
+// @param info : "GetElementPtr" instruction and its dynamic indinces
```
perlfu, line 161: indinces -> indices

include/vkgcDefs.h
```diff
@@ -328,6 +328,8 @@ struct PipelineOptions {
                                    ///  then linking them, when possible.  When not possible this option is ignored.
   bool disableImageResourceCheck;  ///< If set, the pipeline shader will not contain code to check and fix invalid image
                                    ///  descriptors.
+  bool enableScratchBoundsCheck;   ///< If set, accessing a stack object out of bounds will be skipped for stores and
```
perlfu, line 331: Need to update the version history for this.
Is the intention that this be set default true in XGL to match current behaviour?
Should you also remove the targetFeatures += ",+enable-scratch-bounds-checks" code with this patch as well so that duplicate bounds checks are not generated?

perlfu (changes requested):

include/vkgcDefs.h
Flakebi: I tried to find something about out-of-bounds accesses in the SPIR-V and Vulkan specs but to no avail. The robustness extensions are only referring to buffer and image accesses but not about accessing a local (stack) array.
So I think an out-of-bounds OpLoad on a local variable is undefined behavior and we do not need to guard against it.

As the bounds checks incur a performance overhead, I would disable them by default and only enable them if needed.

Removing `+enable-scratch-bounds-checks` is included in this patch :)

lgc/patch/PatchBoundsCheckMemory.cpp
kuhar: Thanks, the example makes things much clearer.

lgc/patch/PatchBoundsCheckMemory.cpp
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * This pass adds bounds checks to all stack/scratch accesses with dynamic indices. The pass looks at every
+ * getelementptr instruction, checking which of the used indices are not constants. For non-constant indices, it checks
+ * if they index into either a FixedVectorType or an ArrayType and gets their element count, which is the upper bound of
+ * the index.
+ *
+ * Example:
+ * We have a getelementptr, followed by a load.
+ *
+ * %elemPtr = getelementptr [16 x float], [16 x float] addrspace(5)* %array, i32 0, i32 %index
```
kuhar, line 38: Does you insert checks regardless of the GEP `inbounds` annotation?

kuhar (commented):
This approach uses GEP types to find out-of-bounds accesses. Should these types be trusted in the first place? I'd expect LLVM passes to freely convert between types, e.g., it'd appear plausible for me for a pass to widen 1D array type to be a power of 2 bytes large, or choose the widest type for a set of pointers that point to arrays of the same element type but different length.

As an alternative, would it be possible to directly check if an accessed pointer if within bounds of its scratch-allocated object? IE something like: when the pointer variable starts with offset 0, check which object it points to (if there's more than one), and then check if any subsequent access if within bounds of that object.

llpc/test/shaderdb/OpLoad_StackBoundsCheck.pipe
kuhar: > I think most (all?) shaderdb tests have their checks at the top or bottom of the file, should I move them around anyway?
LIT doesn't care. Is the other (shadertest?) tool still in use? Not sure what limitations it has.

kuhar: IMO, It would be really valuable to have LGC tests for this, where you can see the IR before and after the pass.

llpc/test/shaderdb/OpLoad_StackBoundsCheck_struct.pipe
kuhar: I think this deserves a comment in code.

kuhar:
@trenouf Would it be possible to have make an LGC test such that it stops just after this pass and prints the module?

include/vkgcDefs.h
perlfu: Thanks, I agree on no default enable.
Comment about updating version history in the header still stands.

trenouf:
> @trenouf Would it be possible to have make an LGC test such that it stops just after this pass and prints the module?

You could write a test that does -stop-after=passname and redirect stderr to FileCheck. (I might have never got around to hooking up -stop-after in LGC's pass manager subclass; if so, it would be simple to hook it up.)

What we cannot yet in general do is a test that starts at a particular pass. That's because LGC still has some internal state in the PipelineState object that does not get serialized and deserialized.

lgc/patch/PatchBoundsCheckMemory.cpp
Flakebi: Yes

Flakebi:
Updated tests and version history. (-stop-after does not work perfectly, there might be side-effects because of different pass-managers?; @kuhar does that kinda match with what you want?)

> This approach uses GEP types to find out-of-bounds accesses. Should these types be trusted in the first place? I'd expect LLVM passes to freely convert between types, e.g., it'd appear plausible for me for a pass to widen 1D array type to be a power of 2 bytes large, or choose the widest type for a set of pointers that point to arrays of the same element type but different length.

I tried to add this pass early in the pipeline (i.e. shortly after translating SPIR-V to LLVM IR), so transformations that do that should happen afterwards.

> As an alternative, would it be possible to directly check if an accessed pointer if within bounds of its scratch-allocated object? IE something like: when the pointer variable starts with offset 0, check which object it points to (if there's more than one), and then check if any subsequent access if within bounds of that object.

I’m not sure if I understand what you mean. Do you mean checking the pointer that results from getelementptr? That was my first try, however for LLVM it is a lot easier to optimize away index checks than proving that the result of getelementptr does not exceed another pointer.
As this pass runs early in the pipeline, we should have the original type of the local variable available. I think the only optimizing passes that run before are the inliner and dead code elimination.

amdvlk-admin:
### Test summary for commit fe81b804ff5c225288681a98f62ad8c59d653510

<li>PAL: dev c9937b277c491a55c689eec7cdd48fb238b8004c</li>
<li>SPVGEN: dev e1f1881d84cd1c7a6c1d661cbac0e619b4e110c1</li>
<li>XGL: dev c6c90450fb1abf5f414acf1e38a0f51a72c426c1</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev bf987708faa5335d515b11a05d898442674fd6ed</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/131664)</summary><ul><li>Built with branch Vulkan-CTS-1.2.0.2-fetch-sources-fix</li>

<li>Passed: 24006/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19889/43895 (45.3%)</li>
</ul></details>

<li>Passed: 24005/43895 (54.7%)</li>
<li>Failed: 0/43895 (0.0%)</li>
<li>Skipped: 19890/43895 (45.3%)</li>
</ul></details>

<li>Passed: 22647/43874 (51.6%)</li>
<li>Failed: 0/43874 (0.0%)</li>
<li>Skipped: 21227/43874 (48.4%)</li>
</ul></details>


</ul></details>

lgc/test/StackBoundsCheck.lgc
```diff
@@ -0,0 +1,128 @@
+; RUN: lgc -mcpu=gfx900 -stop-after=lgc-bounds-check-memory - <%s | FileCheck %s
```
kuhar, line 1: Very cool, that's exactly the kind of test I was hoping for.

lgc/test/StackBoundsCheck.lgc
```diff
@@ -0,0 +1,128 @@
...
+
+; lgc code generated with amdllpc llpc/test/shaderdb/OpLoad_StackBoundsCheck_struct.pipe -emit-lgc
+
+; CHECK:      = fmul reassoc nnan nsz arcp contract afn float %[[#]], 2.550000e+02
```
kuhar, line 5: Consider interleaving these checks with or moving them closer to the input IR. Right now we have to scroll a lot to 'align' the input and output IR and see what's expected to change.

lgc/test/StackBoundsCheck.lgc
```diff
@@ -0,0 +1,128 @@
...
+; CHECK-NEXT: = fptosi float %[[#]] to i32
+; CHECK-NEXT: = icmp ult i32 %[[#]], -16
+; CHECK-NEXT: = xor i1 %[[#]], true
+; CHECK-NEXT: = call { i1, i64 } @llvm.amdgcn.if.i64(i1 %.inv)
+; CHECK-NEXT: = extractvalue { i1, i64 } %[[#]], 0
+; CHECK-NEXT: = extractvalue { i1, i64 } %[[#]], 1
```
kuhar, line 11: Can you reference concrete instruction names to make sure the operands are like you'd expect?

kuhar (commented):
Hi @Flakebi, sorry for the delay with the review, I was on a vacation at the second half of the last week and we had a Thanksgiving Day yesterday.

> Updated tests and version history. (-stop-after does not work perfectly, there might be side-effects because of different pass-managers?; @kuhar does that kinda match with what you want?)

Yes, makes it much easier to see what the pass actually does.

> > This approach uses GEP types to find out-of-bounds accesses. Should these types be trusted in the first place? I'd expect LLVM passes to freely convert between types, e.g., it'd appear plausible for me for a pass to widen 1D array type to be a power of 2 bytes large, or choose the widest type for a set of pointers that point to arrays of the same element type but different length.
> 
> I tried to add this pass early in the pipeline (i.e. shortly after translating SPIR-V to LLVM IR), so transformations that do that should happen afterwards.

> As this pass runs early in the pipeline, we should have the original type of the local variable available. I think the only optimizing passes that run before are the inliner and dead code elimination.

Ack. I think trusting types makes sense since we have (complete) control over the frontend.


> > As an alternative, would it be possible to directly check if an accessed pointer if within bounds of its scratch-allocated object? IE something like: when the pointer variable starts with offset 0, check which object it points to (if there's more than one), and then check if any subsequent access if within bounds of that object.
> 
> I’m not sure if I understand what you mean. Do you mean checking the pointer that results from getelementptr? That was my first try, however for LLVM it is a lot easier to optimize away index checks than proving that the result of getelementptr does not exceed another pointer.

Consider the following C code:
```
float *buffer = ...; // some stack memory
while (nd()) {
  // other code
  // ...
  ++buffer;
}
buffer[0] = 0.0f;
```
where `nd` is some arbitrary condition that the optimizer cannot see through (complex logic, condition over values loaded from memory, etc.).

My worry is that this code won't catch an OOB access, because the largest GEP offset (inside the loop) will be `1`, which is less than the declared aggregate dimensions. I can also imagine a case when you have two separate loops that will have a larger (but in bounds) offsets, but both together will go OOB. Will this pass work in cases like this?

(One more corner case: can the frontend/spir-v use negative GEP offsets?)

What I was trying to say is that if you can know that at the beginning the pointer `p0` points to offset 0 of some object, you could make an instrumentation that checks that whatever GEP instructions run, the final accessed pointer `pN` is sufficiently close to the initial one: `pN - p0 < sizeof(obj)`. I'd guess that in most cases you can precisely tell which object is being accessed, sometimes tell which set of stack objects it could be pointing to (pointer analysis), and sometimes not be able to tell. Is this accurate in this specific case that you are trying to solve?

Flakebi:
Thank you for your comments and reviews @kuhar, they are really helpful.
You’re right, multiple having chained getelementptrs doesn’t work that way.
Some other things I noticed: atomicrmv and memcpy is not handled.
As you said, finding (sets of) pointed-to objects could make it work in some cases, though in cases where we get a pointer through `OpFunctionParameter` it’s either going to be very expensive or impossible.
The original issue that lead to enable-scratch-bounds-checks should also be fixed with the current version of this patch as it is a simple local array out-of-bounds access. But I cannot reproduce the original issue anyway.
@perlfu, do you have a feeling on if this is enough or if it should check more?

I should probably use the scalar evolution analysis to check if a value is in range and not emit a check if it is. That should save the compiler some work.

kuhar:
> The original issue that lead to enable-scratch-bounds-checks should also be fixed with the current version of this patch as it is a simple local array out-of-bounds access. But I cannot reproduce the original issue anyway.
@perlfu, do you have a feeling on if this is enough or if it should check more?

IMO it could make sense to have a pass that handles a class (subset) of OOB accesses, as long as the code/documentation it upfront about it and the necessary assumptions (trustworthy types, absence of negative offsets). However, I such a pass would have to come with some supporting evidence for not breaking correct code (e.g., all accessed in-bounds but GEP types delcared too small for some reason, pointers stating before accessed objects).

lgc/patch/PatchBoundsCheckMemory.h
```diff
@@ -0,0 +1,82 @@
...
+} // namespace llvm
+
+namespace lgc {
+
+// =====================================================================================================================
+// The structure for getelementptr instruction which need to be checked.
+struct GetElemPtrInfo {
+  llvm::GetElementPtrInst *getElemPtr;
+  // Dynamic index value and maximum for each index.
+  llvm::SmallVector<std::pair<llvm::Value *, uint64_t>, 1> dynIndices;
```
nhaehnle, line 52: Careful: gep indices are interpreted as negative values.

nhaehnle (changes requested):
My apologies for not getting around to this earlier.

Can we have another discussion of the design of this, please? Pointers in LLVM IR are supposed to be untyped, and intermediate passes may potentially change types of getelementptrs in ways that ought to be equivalent, but lose some of the range data that this pass attempts to use implicitly.

Coming at this from scratch (heh), the first thing we should note is that there are really two kinds of bounds checking to consider:

1. Check that scratch accesses are in-bounds relative to the underlying memory object. So if we OpLoad from/OpStore to a pointer that was derived from an OpVariable declaring an array, we check exactly against the bounds of the array.

2. Check that scratch accesses cannot go out of a wave's scratch space. In other words, if you have two OpVariable's that are both lowered to scratch memory accesses, then they could still end up clobbering each other; but a scratch access from wave A can never clobber scratch space from wave B (or entirely different memory).

Do we even understand what the requirement is here? I would argue that option 2 would best be addressed with a backend option that sticks to using buffer instructions, because then we can just use the hardware's native bounds checking in a very convenient way.

Option 1 is what this change wants to do, but again the implementation runs against what LLVM wants to be. FWIW, here's a variant that makes sense to me (from a correctness point of view):

During translation from SPIR-V, augment to Private and Function storage class pointers by turning them into triples of `(pointer, lower bound, upper bound)`. Pass those triples through when translating access chains. Translate OpLoad and OpStore into `lgc.bounded.load|store` intrinsics. (Obviously, all of this is conditional on the bounds checking feature being enabled; and it also only needs to be done if the OpVariable is used by something other than OpLoad or OpStore -- keep in mind that `mem2reg` will only work well if native load/store are exposed instead of the lgc intrinsics.)

Add a pass that turns those intrinsics into native LLVM load/store instructions, if in-bounds can be statically proven. This pass should run relatively early in the pipeline, because exposing native load/store instructions will help with further optimizations.

Add a later pass that lowers the intrinsics into the control flow sequence you're using here. This should be late because the added control flow is rather painful. (Something else to consider is whether `llvm.masked.load/store` should be used for this -- it currently only works with vectors, but perhaps that ought to be changed.)

In any case, the first question to me is still: Which granularity of bounds checking is really needed here?

nhaehnle:
FWIW, after discussing this a bit more offline last week, I do think handling this in the frontend is the right approach. It boils down to the fact that this:
```
  %v = alloca ...
  %g = getelementptr ... %v, i32 %idx
  %s = store ... %g, %data
```
is already undefined behavior if `%idx` is out of bounds. Some intermediate transformation may decide to change this code based on such an assumption, which could break the intention of "out of bounds stores get dropped". So to do this right, we already need to produce an alternative code sequence when translating from SPIR-V.

Flakebi:
Okay :)
So the next question is, how far should the checks go?
Possibilities being e.g.
1. Check only simple array and object accesses (like the current state of the patch does)
1. All before + look through multiple stacked getelementptrs
1. All before + look through phis, adding an alias analysis, and check memory address for all of the possible pointed-to objects
1. All before + also use alias analysis for function calls that take a pointer
1. All before + catch out-of-bounds accesses through opaque pointers by implementing something similar to asan

From top to bottom, they
- Catch more out-of-bounds accesses
- Are harder to understand for llvm, thus less likely to get optimized out
- Are harder to implement

nhaehnle:
> Okay :)
> So the next question is, how far should the checks go?

If we're bothering with the feature at all, it should arguably be correct and watertight by design. This means covering all your bullets, and the way I'd see this working is:

1. SPIR-V-to-LLVM lowers relevant SPIR-V pointers to triples of `(ptr, base, bound)`. All pointer-related instructions are taught to operate on these triples (conditionally on the feature being enabled). This means e.g. `OpAccessChain` becomes a `getelementptr` on the `ptr` component while passing `base` and `bound` through unmodified; `OpPhi` is replicated for all elements of the triple; and so on.

2. `OpLoad` and `OpStore` are lowered to lgc intrinsics initially.

3. Use loose values for the triples; putting them into LLVM IR structs may feel natural, but passes have a harder time dealing with those.

4. Function prototypes obviously need to be adjusted as well.

5. As a pragmatic optimization, if it's possible in the SPIR-V reader to immediately determine that a `OpVariable` of `Function` storage class is only used by `OpLoad` and `OpStore` (i.e., no `OpAccessChain` etc.), avoid doing the triple thing from the beginning. This is to help out `mem2reg` in the most common case.

6. This pragmatic optimization could be extended to the case where we can guarantee that any `OpAccessChain`s are using constant indices.

7. Add a pass which reduces the lgc intrinsics introduced in step 2 to plain LLVM IR `load` and `store` in the common cases. Some thought needs to go into where this pass should be inserted in the compilation pipeline. It should probably occur after some initial optimizations (to uncover more information about any relevant `getelementptr` indices), but also not too late, so that subsequent optimizations can understand the simplified plain `load` and `store`.

8. Add a pass towards the end of the compilation pipeline which lowers the lgc intrinsics to bounds checks + control flow.

An alternative at the end would be:

8. Implement the intrinsics in the backend. (We could make the existing llvm.masked.{load/store} intrinsics work on scalars....)

amdrexu:
@Flakebi Not sure if this PR is still on progress.

Flakebi:
Yep, not really in progress. Nicolai’s proposal seems like a good way to go. I’ll close this pr then.

PR #992, opened 2020-10-06
Update llpc changes to anticipate upstream revert of old change

dstutt:
Upstream change:
[InstCombine] Revert rL226781 "Teach InstCombine to canonicalize loads which are
only ever stored to always use a legal integer type if one is available."
(PR47592)

This changes some behaviour that needs fixing in llpc test.
Interestingly, the resultant ISA from these tests is better than before since
there is one fewer buffer_load - even though the IR looks slightly worse in some
cases. However, the later stages in the backend to optimise loads and stores by
combining where possible seems to fix up the issues.


dstutt:
Note: test failures are expected - these are in anticipation of an upstream change.

trenouf (approved):
LGTM. I will merge this.

PR #1026, opened 2020-11-05
Fix dangling else warning (-Wdangling-else)

ehsannas:
TIL: [dangling else](https://en.wikipedia.org/wiki/Dangling_else) is when it's not clear which `if` statement an `else` statement corresponds to. This fixes a build warning (hence build error).

### DIFF
--- a/lgc/patch/PatchWorkarounds.cpp
+++ b/lgc/patch/PatchWorkarounds.cpp
@@ -106,11 +106,13 @@ void PatchWorkarounds::applyImageDescWorkaround(void) {
       if (isImage || isLastUse) {
         for (auto &use : func.uses()) {
           if (auto *callInst = dyn_cast<CallInst>(use.getUser())) {
-            if (callInst->isCallee(&use))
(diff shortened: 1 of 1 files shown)


amdvlk-admin:

lgc/patch/PatchWorkarounds.cpp
```diff
@@ -106,11 +106,13 @@ void PatchWorkarounds::applyImageDescWorkaround(void) {
       if (isImage || isLastUse) {
         for (auto &use : func.uses()) {
           if (auto *callInst = dyn_cast<CallInst>(use.getUser())) {
-            if (callInst->isCallee(&use))
-              if (isLastUse)
+            if (callInst->isCallee(&use)) {
+              if (isLastUse) {
```
amdrexu, line 110: This innermost if-else could have no {}. But it's OK if you like to add them.

amdrexu (approved):

JaxLinAMD:
retest this please

PR #1061, opened 2020-11-26
Report cache access information

jaebaek:
Based on the Vulkan spec, Vulkan application can get the cache hit/miss
information for the pipeline and all shader stages using
[Pipeline Creation Feedback](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap11.html#pipelines-creation-feedback).

Since LLPC uses/maintains the pipeline and the shader caches, the ICD
needs the help from LLPC to get the cache hit/miss information.
This commit lets LLPC report the information to the ICD i.e., XGL.

[This XGL PR](https://github.com/GPUOpen-Drivers/xgl/pull/68) reports
the cache hit/miss information through `VkPipelineCreationFeedbackCreateInfoEXT`
data structure based on the information given by LLPC.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -46,7 +46,7 @@
 #define LLPC_INTERFACE_MAJOR_VERSION 45
 
 /// LLPC minor interface version.
-#define LLPC_INTERFACE_MINOR_VERSION 0
+#define LLPC_INTERFACE_MINOR_VERSION 1
 
 #ifndef LLPC_CLIENT_INTERFACE_MAJOR_VERSION
 #if VFX_INSIDE_SPVGEN
@@ -70,6 +70,7 @@
 //* %Version History
 //* | %Version | Change Description                                                                                    |
 //* | -------- | ----------------------------------------------------------------------------------------------------- |
+//* |     45.1 | Add pipelineCacheAccess, stageCacheAccess(es) to GraphicsPipelineBuildOut/ComputePipelineBuildOut     |
 //* |     45.0 | Remove the member 'enableFastLaunch' of NGG state                                                     |
 //* |     44.0 | Rename the member 'forceNonPassthrough' of NGG state to 'forceCullingMode'                            |
 //* |     43.1 | Add disableImageResourceCheck in PipelineOptions                                                      |
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -777,8 +777,12 @@ Result Compiler::BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, Shad
 // @param context : Acquired context
 // @param shaderInfo : Shader info of this pipeline
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   CacheAccessInfo::CacheNotChecked
 Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                                 ElfPackage *pipelineElf) {
+                                                 ElfPackage *pipelineElf,
+                                                 MutableArrayRef<CacheAccessInfo> stageCacheAccesses) {
   LLPC_OUTS("Building pipeline with relocatable shader elf.\n")
   Result result = Result::Success;
 
@@ -787,6 +791,7 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
   context->getPipelineContext()->setUnlinked(true);
 
   ElfPackage elf[ShaderStageNativeStageCount];
+  assert(stageCacheAccesses.size() >= shaderInfo.size());
   for (unsigned stage = 0; stage < shaderInfo.size() && result == Result::Success; ++stage) {
     if (!shaderInfo[stage] || !shaderInfo[stage]->pModuleData)
       continue;
@@ -826,6 +831,7 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
       // Release Entry
       ReleaseCacheEntry(false, nullptr, &cacheEntry);
       LLPC_OUTS("Cache hit for shader stage " << getShaderStageName(static_cast<ShaderStage>(stage)) << "\n");
+      stageCacheAccesses[stage] = CacheAccessInfo::CacheHit;
       continue;
     }
 
@@ -837,9 +843,11 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
       auto data = reinterpret_cast<const char *>(elfBin.pCode);
       elf[stage].assign(data, data + elfBin.codeSize);
       LLPC_OUTS("Cache hit for shader stage " << getShaderStageName(static_cast<ShaderStage>(stage)) << "\n");
+      stageCacheAccesses[stage] = CacheAccessInfo::CacheHit;
       continue;
     }
     LLPC_OUTS("Cache miss for shader stage " << getShaderStageName(static_cast<ShaderStage>(stage)) << "\n");
+    stageCacheAccesses[stage] = CacheAccessInfo::CacheMiss;
 
     // There was a cache miss, so we need to build the relocatable shader for
     // this stage.
@@ -1424,16 +1432,21 @@ unsigned Compiler::ConvertColorBufferFormatToExportFormat(const ColorTarget *tar
 // @param shaderInfo : Shader info of this graphics pipeline
 // @param buildingRelocatableElf : Build the pipeline by linking relocatable elf
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   CacheAccessInfo::CacheNotChecked
 Result Compiler::buildGraphicsPipelineInternal(GraphicsContext *graphicsContext,
                                                ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                               bool buildingRelocatableElf, ElfPackage *pipelineElf) {
+                                               bool buildingRelocatableElf, ElfPackage *pipelineElf,
+                                               MutableArrayRef<CacheAccessInfo> stageCacheAccesses) {
   Context *context = acquireContext();
   context->attachPipelineContext(graphicsContext);
   Result result = Result::Success;
-  if (buildingRelocatableElf)
-    result = buildPipelineWithRelocatableElf(context, shaderInfo, pipelineElf);
-  else
+  if (buildingRelocatableElf) {
+    result = buildPipelineWithRelocatableElf(context, shaderInfo, pipelineElf, stageCacheAccesses);
+  } else {
     result = buildPipelineInternal(context, shaderInfo, /*unlinked=*/false, pipelineElf);
+  }
   releaseContext(context);
   return result;
 }
@@ -1508,6 +1521,11 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
       cacheResult = lookUpCaches(userCache, &hashId, &elfBin, &cacheEntry);
     else
       cacheEntryState = lookUpShaderCaches(appCache, &cacheHash, &elfBin, &shaderCache, &hEntry);
+
+    if ((cacheEntryState == ShaderEntryState::Ready) || (m_cache && (cacheResult == Result::Success)))
+      pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheHit;
+    else
+      pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheMiss;
   } else {
     cacheEntryState = ShaderEntryState::Compiling;
   }
@@ -1517,7 +1535,8 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
   if (cacheEntryState == ShaderEntryState::Compiling || (m_cache && cacheResult != Result::Success)) {
 
     GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
-    result = buildGraphicsPipelineInternal(&graphicsContext, shaderInfo, buildingRelocatableElf, &candidateElf);
+    result = buildGraphicsPipelineInternal(&graphicsContext, shaderInfo, buildingRelocatableElf, &candidateElf,
+                                           pipelineOut->stageCacheAccesses);
 
     if (result == Result::Success) {
       elfBin.codeSize = candidateElf.size();
@@ -1559,9 +1578,10 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
 // @param pipelineInfo : Pipeline info of this compute pipeline
 // @param buildingRelocatableElf : Build the pipeline by linking relocatable elf
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccess : Compute shader stage cache access result
 Result Compiler::buildComputePipelineInternal(ComputeContext *computeContext,
                                               const ComputePipelineBuildInfo *pipelineInfo, bool buildingRelocatableElf,
-                                              ElfPackage *pipelineElf) {
+                                              ElfPackage *pipelineElf, CacheAccessInfo *stageCacheAccess) {
   Context *context = acquireContext();
   context->attachPipelineContext(computeContext);
 
@@ -1569,10 +1589,13 @@ Result Compiler::buildComputePipelineInternal(ComputeContext *computeContext,
       nullptr, nullptr, nullptr, nullptr, nullptr, &pipelineInfo->cs,
   };
   Result result;
-  if (buildingRelocatableElf)
-    result = buildPipelineWithRelocatableElf(context, shadersInfo, pipelineElf);
-  else
+  if (buildingRelocatableElf) {
+    CacheAccessInfo stageCacheAccesses[ShaderStageCount] = {};
+    result = buildPipelineWithRelocatableElf(context, shadersInfo, pipelineElf, stageCacheAccesses);
+    *stageCacheAccess = stageCacheAccesses[ShaderStageCompute];
+  } else {
     result = buildPipelineInternal(context, shadersInfo, /*unlinked=*/false, pipelineElf);
+  }
   releaseContext(context);
   return result;
 }
@@ -1637,6 +1660,11 @@ Result Compiler::BuildComputePipeline(const ComputePipelineBuildInfo *pipelineIn
       cacheResult = lookUpCaches(userCache, &hashId, &elfBin, &cacheEntry);
     else
       cacheEntryState = lookUpShaderCaches(appCache, &cacheHash, &elfBin, &shaderCache, &hEntry);
+
+    if ((cacheEntryState == ShaderEntryState::Ready) || (m_cache && (cacheResult == Result::Success)))
+      pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheHit;
+    else
+      pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheMiss;
   } else
     cacheEntryState = ShaderEntryState::Compiling;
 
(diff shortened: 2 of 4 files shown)


amdvlk-admin:

jaebaek:
Related PR: https://github.com/GPUOpen-Drivers/xgl/pull/68

llpc/context/llpcCompiler.cpp
```diff
@@ -1561,18 +1571,24 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
...
   };
   Result result;
-  if (buildingRelocatableElf)
-    result = buildPipelineWithRelocatableElf(context, shadersInfo, pipelineElf);
-  else
+  if (buildingRelocatableElf) {
+    StageCacheInfo stageCacheAccesses[ShaderStageCount] = {};
+    result = buildPipelineWithRelocatableElf(context, shadersInfo, pipelineElf, stageCacheAccesses);
+    assert(llvm::all_of(
+        std::array<StageCacheInfo>(stageCacheAccesses, stageCacheAccesses[ShaderStageCompute]),
```
kuhar, line 1586: Why not define `stageCacheAccesses` as `std::array`? I'd also think that `any_of` should work with a plain C array of static size, no?

llpc/include/llpc.h
```diff
@@ -109,14 +109,26 @@ struct ShaderModuleBuildOut {
...
+enum StageCacheInfo : uint8_t {
+  Invalid = 0, ///< Stage cache miss/hit information is invalid.
+  CacheMiss,   ///< Stage cache miss.
+  CacheHit,    ///< Stage cache hit.
+};
+
 /// Represents output of building a graphics pipeline.
 struct GraphicsPipelineBuildOut {
   BinaryData pipelineBin; ///< Output pipeline binary data
+  // Google - To support http://go/teamgob/519403 and http://b/138657311.
```
kuhar, line 121: remove this comment

llpc/include/llpc.h
```diff
@@ -109,14 +109,26 @@ struct ShaderModuleBuildOut {
...
+  bool pipelineCacheHit;                               ///< Output if pipeline cache hits
+  StageCacheInfo stageCacheAccesses[ShaderStageCount]; ///< Output if stage cache hits
 };
 
 /// Represents output of building a compute pipeline.
 struct ComputePipelineBuildOut {
   BinaryData pipelineBin; ///< Output pipeline binary data
+  // Google - To support http://go/teamgob/519403 and http://b/138657311.
```
kuhar, line 129: remove this comment

llpc/include/llpc.h
```diff
@@ -109,14 +109,26 @@ struct ShaderModuleBuildOut {
...
+  CacheMiss,   ///< Stage cache miss.
+  CacheHit,    ///< Stage cache hit.
+};
+
 /// Represents output of building a graphics pipeline.
 struct GraphicsPipelineBuildOut {
   BinaryData pipelineBin; ///< Output pipeline binary data
+  // Google - To support http://go/teamgob/519403 and http://b/138657311.
+  bool pipelineCacheHit;                               ///< Output if pipeline cache hits
+  StageCacheInfo stageCacheAccesses[ShaderStageCount]; ///< Output if stage cache hits
```
kuhar, line 123: I can't parse these two comments. Maybe something like `Indicates whether pipeline cache access was a hit`?

llpc/include/llpc.h
```diff
@@ -109,14 +109,26 @@ struct ShaderModuleBuildOut {
...
 };
 
 /// Represents output of building a compute pipeline.
 struct ComputePipelineBuildOut {
   BinaryData pipelineBin; ///< Output pipeline binary data
+  // Google - To support http://go/teamgob/519403 and http://b/138657311.
+  bool pipelineCacheHit;           ///< Output if pipeline cache hits
+  StageCacheInfo stageCacheAccess; ///< Output if stage cache hits
```
kuhar, line 131: same as above

kuhar (changes requested):
I'm not sure if the public headers should be guarded by `LLPC_INTERFACE_MAJOR_VERSION`

llpc/include/llpc.h
s-perron: I agree.  You should output a more descriptive comment like:

```
bool pipelineCacheHit;                               // True if the pipeline was found in the cache.
StageCacheInfo stageCacheAccesses[ShaderStageCount]; // True if an unlinked shader was found in the cache for the stage.
```

llpc/context/llpcCompiler.cpp
```diff
@@ -778,7 +778,7 @@ Result Compiler::BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, Shad
 // @param shaderInfo : Shader info of this pipeline
 // @param [out] pipelineElf : Output Elf package
 Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                                 ElfPackage *pipelineElf) {
+                                                 ElfPackage *pipelineElf, StageCacheInfo *stageCacheAccesses) {
```
s-perron, line 781: You probably want to use an "ArrayRef" for this paramenter, and everywhere else it is passed internally.

llpc/context/llpcCompiler.cpp
s-perron: The problem is that he wants to leave out the element referring to the compute stage.  I don't think this is worth asserting. The program will still work fine even if the assert fails.  I should be up to `buildPipelineWithRelocatableElf` to make sure it did not do anything unexpected.  It is not the responsibility of each of the users of the function.

llpc/include/llpc.h
jaebaek: done

llpc/include/llpc.h
jaebaek: done

llpc/include/llpc.h
jaebaek: done

llpc/include/llpc.h
```diff
@@ -109,14 +109,26 @@ struct ShaderModuleBuildOut {
   ShaderModuleData *pModuleData; ///< Output shader module data (opaque)
 };
 
+enum StageCacheInfo : uint8_t {
+  Invalid = 0, ///< Stage cache miss/hit information is invalid.
```
s-perron, line 113: I don't know if I like "Invalid" for this name.  It gives the impression that something went wrong.  Something like "CacheUnchecked" might be better.

The comment should be clearer.  From this comment is it not clear why it should be invalid.  You could something like "The cache was not checked."

llpc/context/llpcCompiler.cpp
```diff
@@ -778,7 +778,7 @@ Result Compiler::BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, Shad
 // @param shaderInfo : Shader info of this pipeline
 // @param [out] pipelineElf : Output Elf package
```
s-perron, line 779: You did not add a comment for the new parameter.  Be sure to call out that you expect it to already be initialized.

llpc/context/llpcCompiler.cpp
```diff
@@ -1426,14 +1429,17 @@ unsigned Compiler::ConvertColorBufferFormatToExportFormat(const ColorTarget *tar
...
 Result Compiler::buildGraphicsPipelineInternal(GraphicsContext *graphicsContext,
                                                ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                               bool buildingRelocatableElf, ElfPackage *pipelineElf) {
+                                               bool buildingRelocatableElf, ElfPackage *pipelineElf,
+                                               GraphicsPipelineBuildOut *pipelineOut) {
```
s-perron, line 1433: Need to add a comment for the new parameter.

llpc/include/llpc.h
jaebaek: done

llpc/context/llpcCompiler.cpp
```diff
@@ -1426,14 +1429,17 @@ unsigned Compiler::ConvertColorBufferFormatToExportFormat(const ColorTarget *tar
...
   Context *context = acquireContext();
   context->attachPipelineContext(graphicsContext);
   Result result = Result::Success;
-  if (buildingRelocatableElf)
-    result = buildPipelineWithRelocatableElf(context, shaderInfo, pipelineElf);
-  else
+  if (buildingRelocatableElf) {
+    result = buildPipelineWithRelocatableElf(context, shaderInfo, pipelineElf, pipelineOut->stageCacheAccesses);
+    assert(pipelineOut->stageCacheAccesses[ShaderStageCompute] == StageCacheInfo::Invalid);
```
s-perron, line 1439: I don't see the point of this assert as I mentioned in another comment.

jaebaek (commented):
Thank you for your code review. I updated the code based on it.
In terms of `LLPC_INTERFACE_MAJOR_VERSION`, I used `grep` to understand how I have to use it but I have no idea. Could you please give me some advice how to use it?

s-perron (changes requested):
You should update the PR and commit descriptions to include the reason we are adding this.  It is part of implementing the Vulkan Pipeline creation feedback, and point to a change in XGL that will make use of this to implement that feature.  I think people will want to see the whole thing.

llpc/include/llpc.h
jaebaek: done.

jaebaek (commented):
Thank you for code review. PTAL!

llpc/context/llpcCompiler.cpp
kuhar: `ArrayRef` is only for referencing array-like continuous containers, e.g., `std::vector`, `llvm::SmallVector`, `std::array`, or plain C arrays of static size. Unlike the last two, `ArrayRef` doesn't have a size parameter in its type.

I'd recommend using `std::array` in the place where the source array is defined, and then passing it by `ArrayRef` or `std::array` reference.

kuhar: And `MutableArrayRef` is a mutable version of `ArrayRef` that allows to modify the underlying data.

kuhar:
@jaebaek 
> Thank you for your code review. I updated the code based on it.
> In terms of `LLPC_INTERFACE_MAJOR_VERSION`, I used `grep` to understand how I have to use it but I have no idea. Could you please give me some advice how to use it?

It's defined in `include/vkgcDefs.h`, you can see how it's used to guard some code in https://github.com/GPUOpen-Drivers/llpc/commit/9ce37dd671bfabaf30bcbd12c473e7414d8375f0.

llpc/context/llpcCompiler.cpp
```diff
@@ -61,7 +61,6 @@
 #include "llvm/Support/Timer.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/IPO/AlwaysInliner.h"
-
```
kuhar, line 64: undo this change

llpc/context/llpcCompiler.cpp
```diff
@@ -777,8 +776,12 @@ Result Compiler::BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, Shad
 // @param context : Acquired context
 // @param shaderInfo : Shader info of this pipeline
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   StageCacheInfo::CacheNotChecked
 Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                                 ElfPackage *pipelineElf) {
+                                                 ElfPackage *pipelineElf,
+                                                 StageCacheInfo stageCacheAccesses[ShaderStageCount]) {
```
kuhar, line 784: Don't use C arrays as function arguments. Prefer `llvm::ArrayRef`, or `llvm::MutableArrayRef` when the size is static but not known, `std::array<...>` when the size is static and known, or `llvm::SmallVectorImpl` when the size is dynamic.

llpc/context/llpcCompiler.cpp
```diff
@@ -1424,16 +1430,21 @@ unsigned Compiler::ConvertColorBufferFormatToExportFormat(const ColorTarget *tar
...
 // @param buildingRelocatableElf : Build the pipeline by linking relocatable elf
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   StageCacheInfo::CacheNotChecked
 Result Compiler::buildGraphicsPipelineInternal(GraphicsContext *graphicsContext,
                                                ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                               bool buildingRelocatableElf, ElfPackage *pipelineElf) {
+                                               bool buildingRelocatableElf, ElfPackage *pipelineElf,
+                                               StageCacheInfo stageCacheAccesses[ShaderStageCount]) {
```
kuhar, line 1439: Same as above

llpc/include/llpc.h
```diff
@@ -109,14 +109,24 @@ struct ShaderModuleBuildOut {
...
+enum StageCacheInfo : uint8_t {
+  CacheNotChecked = 0, ///< Stage cache is not checked.
+  CacheMiss,           ///< Stage cache miss.
+  CacheHit,            ///< Stage cache hit.
+};
+
 /// Represents output of building a graphics pipeline.
 struct GraphicsPipelineBuildOut {
   BinaryData pipelineBin; ///< Output pipeline binary data
+  bool pipelineCacheHit;  ///< Indicates whether pipeline cache access was a hit
```
kuhar, line 121: You use 3 values to describe stage cache accesses (`ShaderCacheInfo`) and only 2 values to describe pipeline cache access (hit/miss). Does the code always check pipeline cache, or does the vulkan API force us to return true/false here?

kuhar (changes requested):

include/vkgcDefs.h
```diff
@@ -43,7 +43,7 @@
 #endif
 
 /// LLPC major interface version.
-#define LLPC_INTERFACE_MAJOR_VERSION 45
+#define LLPC_INTERFACE_MAJOR_VERSION 46
```
jaebaek, line 46: @kuhar Does this version update need a special approval from AMD?

jaebaek (commented):
I squashed the commits and updated `LLPC_INTERFACE_MAJOR_VERSION`.

include/vkgcDefs.h
kuhar: We should wait for a review from AMD to confirm.

llpc/context/llpcCompiler.cpp
```diff
@@ -826,6 +830,7 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
       // Release Entry
       ReleaseCacheEntry(false, nullptr, &cacheEntry);
       LLPC_OUTS("Cache hit for shader stage " << getShaderStageName(static_cast<ShaderStage>(stage)) << "\n");
+      stageCacheAccesses[stage] = CacheAccessInfo::CacheHit;
```
kuhar, line 833: You probably want to assert somewhere near the top of the loop that `stageCacheAccess` has enough elements to avoid potential OOB errors if the stages change in the future.

llpc/context/llpcCompiler.h
```diff
@@ -122,13 +122,16 @@ class Compiler : public ICompiler {
...
   Result buildGraphicsPipelineInternal(GraphicsContext *graphicsContext,
                                        llvm::ArrayRef<const PipelineShaderInfo *> shaderInfo,
-                                       bool buildingRelocatableElf, ElfPackage *pipelineElf);
+                                       bool buildingRelocatableElf, ElfPackage *pipelineElf,
+                                       llvm::MutableArrayRef<CacheAccessInfo> stageCacheAccesses);
 
   Result buildComputePipelineInternal(ComputeContext *computeContext, const ComputePipelineBuildInfo *pipelineInfo,
-                                      bool buildingRelocatableElf, ElfPackage *pipelineElf);
+                                      bool buildingRelocatableElf, ElfPackage *pipelineElf,
+                                      CacheAccessInfo *stageCacheAccess);
```
kuhar, line 130: Consider taking `stageCacheAccess` by reference

kuhar (commented):
Looks fine to me, thanks for the fixes. Please wait for reviews from Steven and AMD.

kuhar (approved):

llpc/context/llpcCompiler.cpp
jaebaek: done

llpc/context/llpcCompiler.h
jaebaek: `ElfPackage *pipelineElf` uses a pointer for the output data.
Could you please explain why `stageCacheAccess` should be a reference?

include/vkgcDefs.h
```diff
@@ -70,6 +70,7 @@
 //* %Version History
 //* | %Version | Change Description                                                                                    |
 //* | -------- | ----------------------------------------------------------------------------------------------------- |
+//* |     46.0 | Add pipelineCacheAccess, stageCacheAccess(es) to GraphicsPipelineBuildOut/ComputePipelineBuildOut     |
```
amdrexu, line 73: Since you just add new members to structure info, I think a minor version bump is enough 45.1. If we remove some existing or change the interface method, we need to do a major version bump.

amdrexu (commented):
LGTM. Not sure if Steven or Tim will take a look.

llpc/context/llpcCompiler.h
kuhar: It makes it clear that there's a single non-null object behind it and there's no pointer arithmetic.

llpc/context/llpcCompiler.cpp
```diff
@@ -777,8 +777,12 @@ Result Compiler::BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, Shad
 // @param context : Acquired context
 // @param shaderInfo : Shader info of this pipeline
 // @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
```
s-perron, line 780: I don't know the technical definitions, but, if you expect that that the parameter is initialized, should it be an in/out parameter?  On the other hand, that initial value is never used, it is just so the caller can differentiate between the values that were set and those that were not.

llpc/context/llpcCompiler.cpp
```diff
@@ -1504,9 +1517,14 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
...
   if (!buildingRelocatableElf) {
-    if (m_cache)
+    if (m_cache) {
       cacheResult = lookUpCaches(userCache, &hashId, &elfBin, &cacheEntry);
-    else
+      if (cacheResult == Result::Success) {
+        pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheHit;
+      } else {
+        pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheMiss;
+      }
```
s-perron, line 1526: I think this might be in the wrong place.  The `lookUpShaderCaches` is also a cache lookup, and should set the `pipelineCacheAccess`.  Can you move setting this value to include the lookup in the other cache?

llpc/context/llpcCompiler.cpp
```diff
@@ -1633,9 +1656,14 @@ Result Compiler::BuildComputePipeline(const ComputePipelineBuildInfo *pipelineIn
...
```
s-perron, line 1665: Same as above.

s-perron (approved):
I just have a couple minor comment.  I'm fine even if they do not get fixed.

amdrexu (approved):

JaxLinAMD:
retest this please

amdvlk-admin:
### Test summary for commit b8edb5ede8d719c49f6c3c19431cc4b1195628c6

<li>PAL: dev eb28c7ef53329f838e693f233696823330e964c0</li>
<li>SPVGEN: dev e1f1881d84cd1c7a6c1d661cbac0e619b4e110c1</li>
<li>XGL: dev 63761ecfd2d3cfa61a756ce0a72e1ca1cdae333a</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev fde3d4e2bd90743606ca7b724bf62a53cacf4615</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/144351)</summary><ul><li>Built with version 1.2.2.2</li>

<li>Passed: 26321/48174 (54.6%)</li>
<li>Failed: 0/48174 (0.0%)</li>
<li>Skipped: 21853/48174 (45.4%)</li>
</ul></details>

<li>Passed: 26320/48174 (54.6%)</li>
<li>Failed: 0/48174 (0.0%)</li>
<li>Skipped: 21854/48174 (45.4%)</li>
</ul></details>

<li>Timeout or crash</li>

</ul></details>

<details open><summary>Rhel 8.2, Gfx103</summary><ul>

<li>Passed: 26219/48003 (54.6%)</li>
<li>Failed: 0/48003 (0.0%)</li>
<li>Skipped: 21784/48003 (45.4%)</li>
</ul></details>


</ul></details>

PR #1101, opened 2020-12-15
Separately report cache hit of internal cache

jaebaek:
`Compiler::BuildGraphicsPipeline/BuildComputePipeline()` reports cache
access results to the [Pipeline Creation Feedback](
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap11.html#pipelines-creation-feedback).
Since the pipeline cache and base pipeline acceleration are only two
options to report the cache access result (see
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPipelineCreationFeedbackFlagBitsEXT.html),
we have to separately report the cache hit of the internal cache.
This commit separates the cache hit of the internal cache from the
cache hit of the pipeline cache that is given by VkPipelineCache.

### DIFF
--- a/lgc/patch/PatchInOutImportExport.cpp
+++ b/lgc/patch/PatchInOutImportExport.cpp
@@ -2871,7 +2871,9 @@ void PatchInOutImportExport::patchTcsBuiltInOutputExport(Value *output, unsigned
     Value *tessFactorOffset = calcTessFactorOffset(isOuter, elemIdx, insertPos);
     storeTessFactorToBuffer(tessFactors, tessFactorOffset, insertPos);
 
-    assert(perPatchBuiltInOutLocMap.find(builtInId) != perPatchBuiltInOutLocMap.end());
+    if (perPatchBuiltInOutLocMap.count(builtInId) == 0)
+      return; // Avoid writing unused tessellation factor to Lds
+
     unsigned loc = perPatchBuiltInOutLocMap[builtInId];
 
     if (!elemIdx) {
--- a/lgc/patch/PatchResourceCollect.cpp
+++ b/lgc/patch/PatchResourceCollect.cpp
@@ -1761,10 +1761,9 @@ void PatchResourceCollect::mapBuiltInToGenericInOut() {
         const unsigned mapLoc = nextInOutUsage.perPatchBuiltInInputLocMap[BuiltInTessLevelOuter];
         inOutUsage.perPatchBuiltInOutputLocMap[BuiltInTessLevelOuter] = mapLoc;
         availPerPatchOutMapLoc = std::max(availPerPatchOutMapLoc, mapLoc + 1);
-      } else {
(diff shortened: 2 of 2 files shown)


amdvlk-admin:

llpc/context/llpcCompiler.cpp
```diff
@@ -846,7 +846,7 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
       auto data = reinterpret_cast<const char *>(elfBin.pCode);
       elf[stage].assign(data, data + elfBin.codeSize);
       LLPC_OUTS("Cache hit for shader stage " << getShaderStageName(static_cast<ShaderStage>(stage)) << "\n");
-      stageCacheAccesses[stage] = CacheAccessInfo::CacheHit;
+      stageCacheAccesses[stage] = CacheAccessInfo::InternalCacheHit;
```
s-perron, line 849: We should double check.  I just realized that if `userShaderCache` is not nullptr, then it could be an application cache that is being checked.

llpc/context/llpcCompiler.cpp
```diff
@@ -1520,14 +1520,16 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
...
+      if (cacheResult == Result::Success)
+        pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheHit;
+    } else {
       cacheEntryState = lookUpShaderCaches(appCache, &cacheHash, &elfBin, &shaderCache, &hEntry);
-
-    if ((cacheEntryState == ShaderEntryState::Ready) || (m_cache && (cacheResult == Result::Success)))
-      pipelineOut->pipelineCacheAccess = CacheAccessInfo::CacheHit;
-    else
+      if (cacheEntryState == ShaderEntryState::Ready)
+        pipelineOut->pipelineCacheAccess = CacheAccessInfo::InternalCacheHit;
```
s-perron, line 1530: Same as above, I think that if `appCache` is not nullptr, then it is checking an application cache.

llpc/context/llpcCompiler.cpp
jaebaek: I checked the XGL https://github.com/GPUOpen-Drivers/xgl.
As far as I checked, `pipelineInfo->pShaderCache` is never set by XGL and it should be `nullptr`.

llpc/context/llpcCompiler.cpp
jaebaek: Same as above

llpc/context/llpcCompiler.cpp
jaebaek: Updated

llpc/context/llpcCompiler.cpp
jaebaek: Updated

jaebaek:
I re-sent it in #1102 because of git-squash and force-pushing mistake.

PR #1110, opened 2020-12-25
Support spirv with capability linkage

jiaolu:
Spirv with capability linkage does not have entry function, so it is required to create
dummy entry function of compute shader stage for this spirv.

### DIFF
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -6136,12 +6136,11 @@ bool SPIRVToLLVM::translate(ExecutionModel entryExecModel, const char *entryName
 
   // Find the targeted entry-point in this translation
   auto entryPoint = m_bm->getEntryPoint(entryExecModel, entryName);
-  if (!entryPoint)
+
+  if (!entryPoint && !m_bm->hasCapability(CapabilityLinkage))
     return false;
 
-  m_entryTarget = m_bm->get<SPIRVFunction>(entryPoint->getTargetId());
-  if (!m_entryTarget)
-    return false;
+  m_entryTarget = (entryPoint != nullptr) ? m_bm->get<SPIRVFunction>(entryPoint->getTargetId()) : nullptr;
 
   m_execModule = entryExecModel;
   m_fpControlFlags.U32All = 0;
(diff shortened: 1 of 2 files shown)


llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -8121,6 +8124,21 @@ llvm::GlobalValue::LinkageTypes SPIRVToLLVM::transLinkageType(const SPIRVValue *
...
+  FunctionType *funcTy = FunctionType::get(builder->getVoidTy(), {}, false);
+  auto func = Function::Create(funcTy, GlobalValue::ExternalLinkage, "libraryEntry", m_m);
+  BasicBlock *entryBlock = BasicBlock::Create(*m_context, "", func);
+  builder->SetInsertPoint(entryBlock);
+  builder->CreateRetVoid();
+  std::vector<Metadata *> execModelMDs;
+  execModelMDs.push_back(ConstantAsMetadata::get(ConstantInt::get(builder->getInt32Ty(), ExecutionModelGLCompute)));
+  auto execModelMdNode = MDNode::get(*m_context, execModelMDs);
+  func->addMetadata(gSPIRVMD::ExecutionModel, *execModelMdNode);
+  func->setDLLStorageClass(GlobalValue::DLLExportStorageClass);
```
Flakebi, line 8138: IIRC @trenouf once said that `dllexport` is used to mark shader entry points (i.e. functions that are called by hardware in contrast to functions called by software), so I think this should not be marked as `dllexport` (but let’s wait for him to confirm).

jiaolu: spirv with capability linkage is library spirv,  hence it can not be the module with hardware entry function.
however we need to get the module information i.e. module shader stage, from entry function.
so i mark that dummy function as DLLExportStorageClass

Flakebi: Ah, I see.

I think, it makes sense to have no entry function in that case, that would make it clear that there is no entrypoint.
lgc (middle-end) and llvm should be fine without an entrypoint, the shader stage is stored in function metadata in lgc and in the calling convention in llvm.
As far as I can see, the parts that need to be changed are

1. the SPIR-V execution model: I guess we can put this onto every function instead of only on entrypoints, similar to how it is done in lgc
2. lowering globals currently inserts code in the entrypoint: I think for non-entrypoint modules, this needs to be handled differently anyway. Maybe by moving the global into function arguments?

jiaolu: our lowering pass assume the entry point a lot place in code, to rid of this assumption, i am afraid there are more than a few places to modify and adding code complexity.
It's about trade-off.  to add dummy entry point , which you can argue it is a little bit hack, is simple and easy without adding a lot of technical liability.

amdrexu (approved):

PR #1114, opened 2020-12-30
Mark push constant for spill for compute library

LLJJDD:
The value of spill_threshold should be unified across the pipeline, otherwise PAL may not correctly spill data into memory. Mark push constant for spill for compute library even it does not actually use it.

### DIFF
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -1315,10 +1315,13 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
           addUserDataArg(userDataArgs, node.offsetInDwords + dwordOffset, pushConstOffset.dwordSize,
                          &pushConstOffset.entryArgIdx, builder);
         }
+      } else {
+        // Mark push constant for spill for compute library.
+        userDataUsage->pushConstSpill = true;
       }
 
       // Ensure we mark the push constant's part of the spill table as used.
-      if (userDataUsage->pushConstSpill || isComputeWithCalls())
+      if (userDataUsage->pushConstSpill)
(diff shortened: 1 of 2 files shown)


lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1315,6 +1315,9 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
           addUserDataArg(userDataArgs, node.offsetInDwords + dwordOffset, pushConstOffset.dwordSize,
                          &pushConstOffset.entryArgIdx, builder);
         }
+      } else {
+        // Mark push constant for spill for compute library.
+        userDataUsage->pushConstSpill = true;
       }
 
       // Ensure we mark the push constant's part of the spill table as used.
```
Flakebi, line 1323: You can remove the check for `|| isComputeWithCalls()` in the line below (GitHub doesn’t let me comment on that line directly).

Flakebi (approved):
Looks good, thanks. Just a small remark.

If you have a case where it previously failed, could you add that as an lgc test case?

LLJJDD:
> Looks good, thanks. Just a small remark.
> 
> If you have a case where it previously failed, could you add that as an lgc test case?

My case is that there is a set of compute library in one pipeline, and one of them uses push constant. I'm not sure whether a lgc test can represent that.

Or just having a test to make sure push constant is spilled for any compute library in a pipeline that has push constant resource node?

Flakebi (approved):
Great, thanks

amdrexu (approved):

JaxLinAMD:
retest this please

JaxLinAMD:
@LLJJDD  your branch  is 11 commits behind GPUOpen-Drivers:dev, could you rebase that and force-push again?

LLJJDD:
Done.

PR #900, opened 2020-08-14
Link relocatable shaders with data in the .rodata section

s-perron:
The current relocatabale shader linking does not work when there is read
only data.  See https://github.com/GPUOpen-Drivers/llpc/issues/896 for
an example.  So a change is going into LLVM
(https://reviews.llvm.org/D85895) to place the data in the .rodata
section instead of the .text section.  This change is to have the elf
linker handle it correclty.

Fixes #896

PR #1122, opened 2021-01-11
Support merging ELF binaries containing relocs

xuechen417:
ELF binary containing reloc section isn't implemented in
ElfWriter<Elf>::mergeElfBinary. Thus, partial pipeline caching is
disabled as a workaround for this case to avoid regressions in the cts
tests VK.robustness.robustness2.*. mergeELfBinary merges the ELF binary
of fragment shader into ELF bianry of non-fragment shader. This change
will includes:
- Enable partial pipeline caching for reclos
- If the ELF binary of non-fragment binary contain relco, remove reloc
section and symbol
- If the ELF binary of fragment binary contain reloc, add reloc section
and symbol, and update reloc offset and the corresponding data.

Fixes: VK.robustness.robustness2.*

### DIFF
--- a/lgc/patch/PatchCheckShaderCache.cpp
+++ b/lgc/patch/PatchCheckShaderCache.cpp
@@ -89,8 +89,8 @@ bool PatchCheckShaderCache::runOnModule(Module &module) {
   Patch::init(&module);
 
   // NOTE: Global constants are either added to the end of the .text section, or in a separate .rodata section with
-  // relocs in the .text section to refer to them. We can't merge ELF binaries if relocs are used because llpcElfWriter
-  // doesn't know how to merge them.
+  // relocs in the .text section to refer to them. We can't merge ELF binaries if relocs are used in non-fragment shader
+  // stages.
   for (auto &global : module.globals()) {
     if (auto globalVar = dyn_cast<GlobalVariable>(&global)) {
       if (globalVar->isConstant()) {
@@ -102,7 +102,8 @@ bool PatchCheckShaderCache::runOnModule(Module &module) {
               vals.push_back(user);
               continue;
             }
-            return false;
+            if (getShaderStage(cast<Instruction>(user)->getFunction()) != ShaderStageFragment)
+              return false;
           }
         }
       }
--- a/llpc/util/llpcElfWriter.cpp
+++ b/llpc/util/llpcElfWriter.cpp
@@ -494,14 +494,20 @@ template <class Elf> void ElfWriter<Elf>::assembleSymbols() {
   if (m_symSecIdx == InvalidValue)
     return;
   auto strTabSection = &m_sections[m_strtabSecIdx];
-  auto newStrTabSize = 0;
+  unsigned newStrTabSize = 0;
   unsigned symbolCount = 0;
+  ElfSymbol *fragmentIsaSymbol = nullptr;
+  auto fragmentIsaSymbolName =
+      Util::Abi::PipelineAbiSymbolNameStrings[static_cast<unsigned>(Util::Abi::PipelineSymbolType::PsMainEntry)];
   for (auto &symbol : m_symbols) {
     if (symbol.nameOffset == InvalidValue)
       newStrTabSize += strlen(symbol.pSymName) + 1;
 
     if (symbol.secIdx != InvalidValue)
       symbolCount++;
+
+    if (strcmp(symbol.pSymName, fragmentIsaSymbolName) == 0)
+      fragmentIsaSymbol = &symbol;
   }
 
   if (newStrTabSize > 0) {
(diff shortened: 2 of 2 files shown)


amdvlk-admin:
### Test summary for commit fca4207c4961d92e58d59b056c9cd90a28872e79

<li>PAL: dev 1e61f9ab6eb2041c73d7e7bd8f2021fa05a4bdd8</li>
<li>SPVGEN: dev 7c1cf34dd7eb4d3a6e6652c26292eef1bfcbcaf0</li>
<li>XGL: dev 3237b046add2809634e7f5f3fcdea1af4ad6613d</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 294f18dfb21b21ca76f4b742bffa3177babe46ff</li>
</ul></details>

<details><summary>CTS tests (Failed: 2/168360)</summary><ul><li>Built with version 1.2.2.2</li>

<li>Passed: 26814/56120 (47.8%)</li>
<li>Failed: 0/56120 (0.0%)</li>
<li>Skipped: 29306/56120 (52.2%)</li>
</ul></details>

<li>Passed: 26812/56120 (47.8%)</li>
<li><details><summary>Failed: 2/56120 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.texture.explicit_lod.2d.sizes.128x128_linear_linear_mipmap_nearest_clamp
Stack trace: Flake

FAILURE: dEQP-VK.texture.filtering.3d.combinations.nearest_mipmap_linear.nearest.mirrored_repeat.mirror_clamp_to_edge.mirrored_repeat
Stack trace: Flake
```
</p></details></li>

<li>Skipped: 29306/56120 (52.2%)</li>
</ul></details>

<li>Timeout or crash</li>

</ul></details>

<li>Passed: 26829/56120 (47.8%)</li>
<li>Failed: 0/56120 (0.0%)</li>
<li>Skipped: 29291/56120 (52.2%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchCheckShaderCache.cpp
```diff
@@ -102,7 +102,8 @@ bool PatchCheckShaderCache::runOnModule(Module &module) {
               vals.push_back(user);
               continue;
             }
-            return false;
+            if (getShaderStage(cast<Instruction>(user)->getFunction()) != ShaderStageFragment)
+              return false;
```
jayfoad, line 106: My understanding of this is: if a Vs (for example) uses global constants, then it might put them at the end of the text section, so we wouldn't be able to merge it with a Fs to make a VsFs pipeline, because it wouldn't be able to just "fall through" into the Fs.

But is this still true? Don't we have a mechanism for forcing the Vs to put constants in a separate section and use relocs to access them?

s-perron: It is possible for LLVM to generate the data in a separate section.  There is a test that check for that https://github.com/GPUOpen-Drivers/llpc/blob/e1e567748c5b80ae06308cecfcb440f0b4663e49/llpc/test/shaderdb/relocatable_shaders/PipelineVsFs_MultipleConstData.pipe#L5.

s-perron (commented):
Would it be possible to have some tests written so we can see how this works.

Also, I'd like to try to understand the assumptions you are making.  It seems like you are assuming that the relocations are only in the fragment shader, and that there is no prologue or epilogue for the fragment shader.

lgc/patch/PatchCheckShaderCache.cpp
jayfoad: I know that, but I don't know how LLVM makes that decision.

I am suggesting that if a Vs has global constants, then rather than disabling partial pipeline compilation here, it would be better to do whatever it is you need to do to tell LLVM to put them in a separate section.

xuechen417: > Would it be possible to have some tests written so we can see how this works.
> 
> Also, I'd like to try to understand the assumptions you are making. It seems like you are assuming that the relocations are only in the fragment shader, and that there is no prologue or epilogue for the fragment shader.

Sorry for missing the tests description.
The issue is produced when runing the following two tests with ShaderCacheMode 1 on Navi21
dEQP-VK.robustness.robustness2.bind.notemplate.r32ui.dontunroll.nonvolatile.uniform_buffer.no_fmt_qual.null_descriptor.samples_1.1d.frag (pass)
dEQP-VK.robustness.robustness2.bind.notemplate.r32ui.dontunroll.nonvolatile.uniform_buffer.no_fmt_qual.len_4.samples_1.1d.frag (fail, it has reloc section in FS but the offset and corresponding data at that offset is wrong)

xuechen417: @jayfoad The condition `if (getShaderStage(cast<Instruction>(user)->getFunction()) != ShaderStageFragment)` was removed at early time as a workaround to make the above tests pass, because ELF with reloc in FS wasn't implemented in `mergeElfBinary`. Here, I restore the condition after refining `mergeElfBinary`. 
You mean we shouldn't disable partial pipeline compilation here. Could you explain more on " tell LLVM to put them in a separate section"?

jayfoad: I don't know how to "tell LLVM to put them in a separate section". I was hoping someone else could explain that part, if it is possible.

s-perron: @xuechen417 The test that I linked to above tests that when relocatable shaders are enabled, the constant data is handled correctly when there is constant data in the VS.  There should be no fundamental reason to not handle a relocation for constant data.

The way this is done with relocatable shaders is to have LLVM place the data in its own section, and then the linking code makes sure the relocations are still valid.  This is the PR that enabled it: https://github.com/GPUOpen-Drivers/llpc/pull/900.

I cannot say exactly what tells the LLVM to place the data in the .rodata section.  That part was done by others.  I agree with Jay that this should be handled by placing the the constants in a different section.  This will handle even more cases.

You can try to trace what happens when --enable-relocatable-shader-elf is used and when it is not to find the difference.

s-perron:
> The issue is produced when runing the following two tests with ShaderCacheMode 1 on Navi21
dEQP-VK.robustness.robustness2.bind.notemplate.r32ui.dontunroll.nonvolatile.uniform_buffer.no_fmt_qual.null_descriptor.samples_1.1d.frag (pass)
dEQP-VK.robustness.robustness2.bind.notemplate.r32ui.dontunroll.nonvolatile.uniform_buffer.no_fmt_qual.len_4.samples_1.1d.frag (fail, it has reloc section in FS but the offset and corresponding data at that offset is wrong)

Having small shaderdb tests for stuff like this is good.  I someone makes a change that breaks it then they will know right away, and will be easier to debug than having to run all of CTS.  It also makes it easier to review the code because I can see what cases you are thinking of and maybe notice cases you did not think of.

xuechen417:
Thanks @jayfoad @s-perron for your review and good suggestions. I will try --enable-relocatable-shader-elf and close the PR temporarily.

PR #1123, opened 2021-01-11
Fix buffer descriptor indexing with NonUniform

xuechen417:
The argument `isNonUniform` of `CreateLoadBufferDesc` is determined by
checking the users of the block index for any NonUniform calls. If there is any
NonUniform call, the flag is true, otherwise it is false. This way will
give a wrong answer for the case that `NonUniform` is decorated on a GEP
instruction not the index in SPIR-V shader. To fix it, we need check the
users of GEP itself for any NonUniform calls before checking the users
of the index to avoid creating readfirstlane on the index.

The SPIR-V shader can be written as below:
OpDecorate %21 NonUniform
...
%17 = OpLoad %int %rIndex
%18 = OpCopyObject %int %17
%21 = OpAccessChain %_ptr_StorageBuffer_v4float %data %18 %int_1

### DIFF
--- a/llpc/lower/llpcSpirvLowerGlobal.cpp
+++ b/llpc/lower/llpcSpirvLowerGlobal.cpp
@@ -1737,18 +1737,32 @@ void SpirvLowerGlobal::lowerBufferBlock() {
 
           bool isNonUniform = false;
 
-          // Run the users of the block index to check for any nonuniform calls.
-          for (User *const user : blockIndex->users()) {
+          // Run the users of the GEP to check for any nonuniform calls.
+          for (User *const user : getElemPtr->users()) {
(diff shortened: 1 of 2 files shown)


dstutt (approved):
LGTM

Has this become an issue since the changes were made to the way the decoration for NonUniform now happens?

xuechen417:
> LGTM
> 
> Has this become an issue since the changes were made to the way the decoration for NonUniform now happens?

No, current tests has not cover this case. This is a fix for some furture tests.

JaxLinAMD:
@amdrexu please  review

amdrexu:
@JaxLinAMD I'm OK with this PR and David has approved it. Please merge it.

PR #1129, opened 2021-01-18
Support the gl_ViewIndex in TCS and Fix the issue of no ouput

amd-wangd:


amdrexu (approved):
In the future, if two issues are irrelevant, you can make two pull requests.

amd-wangd:
> In the future, if two issues are irrelevant, you can make two pull requests.

I see, thank you.

PR #894, opened 2020-08-12
Rework implementation for OpCopyObject

LLJJDD:
Previously OpCopyObject is handled by a store + load, it will lead to problem when copying pointer type object, for example, mentioned in #834. LLPC fails to handle the pointer-to-pointer type in some lower passes so invalid IR is generated / assertion is triggered.

Directly mapping the llvm value of the Operand to the Result should be enough for OpCopyObject, since the Operand and Result should be exactly the same for OpCopyObject.

This fixes #834.

PR #1130, opened 2021-01-18
Fix OpCopyObject with NonUniform decoration

LLJJDD:
#894 makes OpCopyObject to use the llvm value of the operand directly instead of inserting store + load. When the OpCopyObject is in a different block to its operand, and is decorated with NonUniform, invalid IR is generated as the call to spirv.NonUniform is inserted in the operand's after the block terminator.

In such case, we generate the call before the block terminator instead.

### DIFF
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -6860,7 +6860,14 @@ bool SPIRVToLLVM::transShaderDecoration(SPIRVValue *bv, Value *v) {
       std::string mangledFuncName(gSPIRVMD::NonUniform);
       appendTypeMangling(nullptr, args, mangledFuncName);
       auto f = getOrCreateFunction(m_m, voidTy, types, mangledFuncName);
-      CallInst::Create(f, args, "", bb);
+      if (bb->getTerminator()) {
+        // NOTE: For OpCopyObject, we use the operand value directly, which may be in another block that already has a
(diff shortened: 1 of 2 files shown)


JaxLinAMD:
retest this please

dstutt (approved):
LGTM

Minor niggle - might it be possible to make the test do something other than just test for success? But I don't think it is essential.

LLJJDD:
This test is a simplified one of a shader which crashes the compiler - so I think a success should be enough.

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6860,7 +6860,13 @@ bool SPIRVToLLVM::transShaderDecoration(SPIRVValue *bv, Value *v) {
...
       appendTypeMangling(nullptr, args, mangledFuncName);
       auto f = getOrCreateFunction(m_m, voidTy, types, mangledFuncName);
-      CallInst::Create(f, args, "", bb);
+      if (bb->getTerminator()) {
+        // NOTE: For OpCopyObject, we use the operand value directly, which may be in another block that already has a
+        // terminator. In this case, insert the call before the terminator.
+        assert(bv->getOpCode() == OpCopyObject);
+        CallInst::Create(f, args, "", bb->getTerminator());
+      } else
+        CallInst::Create(f, args, "", bb);
```
kuhar, line 6869: nit: missing braces around else's body

JaxLinAMD:
retest this please

JaxLinAMD:
retest this please

PR #1153, opened 2021-02-03
Add thread group swizzling feature for compute shaders.

ruanjm:
This feature generates new thread group id based on system thread group id for user. The algorithm is generating morton code from the flat system thread group id. So that the order of new id would look like z-order Lebesgue curve.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -47,7 +47,7 @@
 #define LLPC_INTERFACE_MAJOR_VERSION 45
 
 /// LLPC minor interface version.
-#define LLPC_INTERFACE_MINOR_VERSION 4
+#define LLPC_INTERFACE_MINOR_VERSION 5
 
 #ifndef LLPC_CLIENT_INTERFACE_MAJOR_VERSION
 #if VFX_INSIDE_SPVGEN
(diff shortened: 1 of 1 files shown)


amdvlk-admin:

JaxLinAMD:
ok to test( <-- token to trigger jenkins)

JaxLinAMD:
add to whitelist (<- token to add ruanjm to whitelist)

include/vkgcDefs.h
```diff
@@ -328,6 +328,15 @@ struct ExtendedRobustness {
                            ///  valid to access and acts as if the descriptor is bounded to nothing.
 };
 
+/// Represents the tiling modes for compute shader thread group swizzling
```
amdrexu, line 331: Please bump the version to 45.5 and add a comment in the beginning of this file (mention that you add a new enum type for CS swizzling).

LLPC_INTERFACE_MINOR_VERSION 4 -> LLPC_INTERFACE_MINOR_VERSION 5

amdrexu (approved):
LGTM.
Next time, when you make any code update. You can use `git commit --amend -C head` to modify your changes and submit them, rather than using `git commit`. This will not generate new commit for this pull request.

PR #1156, opened 2021-02-05
Remove adjusting vertex offset for DynamicStride

ArkadiuszSarwaAMD:
If the vertex offset is greater than vertex stride, we have to adjust both vertex index and offset
accordingly. Otherwise, vertex fetch might behave unexpectedly. It cannot be applied to DynamicVertexStride.
Setting stride to 0 disallow for that adjustment.


amdvlk-admin:

PR #1160, opened 2021-02-09
lgc testing lit test check strings can no longer be unused

dstutt:
Changes to how lit testing works mean that you can't define check labels on the
RUN line and not use them in the test body.

Fix one malformed case and adjust another test to allow for different labels for
NGG


piotrAMD:
LGTM.

piotrAMD (approved):

PR #1168, opened 2021-02-15
Update to handle new size of TimeRecord in upstream LLVM.

perlfu:

PR #1169, opened 2021-02-16
Use updated api of CloneFunctionInto

Sisyph:
In upstream LLVM 22a52dfddcefad4f275eb8ad1cc0e200074c2d8a
they changed the visibility of changes from a bool to an enum.

### DIFF
--- a/lgc/patch/NggPrimShader.cpp
+++ b/lgc/patch/NggPrimShader.cpp
@@ -2885,7 +2885,11 @@ void NggPrimShader::splitEs(Module *module) {
     valueMap[&arg] = newArg++;
 
   SmallVector<ReturnInst *, 8> retInsts;
+#if LLVM_MAIN_REVISION && LLVM_MAIN_REVISION < 379998
   CloneFunctionInto(esCullDataFetchFunc, esEntryPoint, valueMap, false, retInsts);
+#else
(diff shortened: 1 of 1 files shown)


amdvlk-admin:

Sisyph:
Not sure about the CI involved, but this patch should be built against LLVM that is updated through 22a52dfddcefad4f275eb8ad1cc0e200074c2d8a

lgc/patch/NggPrimShader.cpp
```diff
@@ -2885,7 +2885,7 @@ void NggPrimShader::splitEs(Module *module) {
     valueMap[&arg] = newArg++;
 
   SmallVector<ReturnInst *, 8> retInsts;
-  CloneFunctionInto(esCullDataFetchFunc, esEntryPoint, valueMap, false, retInsts);
+  CloneFunctionInto(esCullDataFetchFunc, esEntryPoint, valueMap, CloneFunctionChangeType::LocalChangesOnly, retInsts);
```
trenouf, line 2888: As Carl said elsewhere, please guard this with the applicable LLVM revision number, like in #1168. Then it does not need to be synchronized with an update to the LLVM it is built with.

trenouf (changes requested):

JaxLinAMD:
ok to test

trenouf (approved):

PR #1178, opened 2021-02-26
Fix LLVM_MAIN_REVISION to a correct value for flag day change in PR#1177

dstutt:
The LLVM_MAIN_REVISION was calculated incorrectly in the previous change

Confirmed this builds with both versions of LLVM (pre and post change), as well as verifying that next merge has the correct revision number.


trenouf (approved):
LGTM. I will merge this.

PR #1186, opened 2021-03-08
Remove workaround in ArithBuilder::CreateSMod

jayfoad:
The comments mention a backend bug relating to i16 mod, but I think it
must have been fixed because Vulkan CTS testing shows no failures on a
variety of different hardware.

### DIFF
--- a/lgc/builder/ArithBuilder.cpp
+++ b/lgc/builder/ArithBuilder.cpp
@@ -217,27 +217,6 @@ Value *ArithBuilder::CreateQuantizeToFp16(Value *value, const Twine &instName) {
 // @param divisor : Divisor value
 // @param instName : Name to give instruction(s)
 Value *ArithBuilder::CreateSMod(Value *dividend, Value *divisor, const Twine &instName) {
-  if (divisor->getType()->getScalarType()->isIntegerTy(32) &&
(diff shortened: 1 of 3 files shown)


lgc/builder/ArithBuilder.cpp
```diff
@@ -217,27 +217,6 @@ Value *ArithBuilder::CreateQuantizeToFp16(Value *value, const Twine &instName) {
 // @param divisor : Divisor value
 // @param instName : Name to give instruction(s)
 Value *ArithBuilder::CreateSMod(Value *dividend, Value *divisor, const Twine &instName) {
-  if (divisor->getType()->getScalarType()->isIntegerTy(32) &&
-      getPipelineState()->getTargetInfo().getGpuWorkarounds().gfx10.disableI32ModToI16Mod) {
```
jayfoad, line 221: @trenouf should I also remove the definition of "disableI32ModToI16Mod", and are there any subtleties I need to be aware of, like replacing it with a dummy bit?

amdrexu: I think you can completely remove the workaround if backend behaves well now.

amdrexu (commented):
LGTM

JaxLinAMD:
retest this please

amdrexu (approved):

PR #1189, opened 2021-03-12
Amend Unroll loop hint handling

tcorringham:
Added a default value of 1800 for -unroll-hint-threshold.
This will be used as the default loop threshold when unrolling loops with the SPIR-V Unroll hint. This value may be amended by the AMDGPU specific heuristics.
Removed the default value for -pragma-unroll-threshold as that is no longer required.


tcorringham:
Whoops, wrong change!

PR #1195, opened 2021-03-23
Support input-output packing for non-zero-based vector components

kuhar:
Don't assume that the original vector elements start at 0. This fixes a null pointer dereference.

Print vector components in LLPC location input/output mapping results.

Add a `-v` flag to the lgc tool to test this.

### DIFF
--- a/lgc/patch/PatchResourceCollect.cpp
+++ b/lgc/patch/PatchResourceCollect.cpp
@@ -1354,11 +1354,13 @@ void PatchResourceCollect::matchGenericInOut() {
     assert(inOutUsage.inputMapLocCount == 0);
     for (const auto &locInfoPair : inLocInfoMap) {
       const unsigned origLoc = locInfoPair.first.getLocation();
+      const unsigned origComp = locInfoPair.first.getComponent();
       const unsigned newLoc = locInfoPair.second.getLocation();
+      const unsigned newComp = locInfoPair.second.getComponent();
       assert(newLoc != InvalidValue);
       inOutUsage.inputMapLocCount = std::max(inOutUsage.inputMapLocCount, newLoc + 1);
       LLPC_OUTS("(" << getShaderStageAbbreviation(m_shaderStage) << ") Input:  loc = " << origLoc
-                    << "  =>  Mapped = " << newLoc << "\n");
+                    << ", comp = " << origComp << " =>  Mapped = " << newLoc << ", " << newComp << "\n");
     }
     LLPC_OUTS("\n");
   }
@@ -1374,7 +1376,9 @@ void PatchResourceCollect::matchGenericInOut() {
     // Note: The value of outLocMap should be filled in packing code path
     for (const auto &locInfoPair : outLocInfoMap) {
       const unsigned origLoc = locInfoPair.first.getLocation();
+      const unsigned origComp = locInfoPair.first.getComponent();
       const unsigned newLoc = locInfoPair.second.getLocation();
+      const unsigned newComp = locInfoPair.second.getComponent();
 
       if (m_shaderStage == ShaderStageGeometry) {
         const unsigned streamId = locInfoPair.first.getStreamId();
@@ -1383,11 +1387,12 @@ void PatchResourceCollect::matchGenericInOut() {
 
         inOutUsage.outputMapLocCount = std::max(inOutUsage.outputMapLocCount, assignedLocCount);
         LLPC_OUTS("(" << getShaderStageAbbreviation(m_shaderStage) << ") Output: stream = " << streamId << ", "
-                      << " loc = " << origLoc << "  =>  Mapped = " << newLoc << "\n");
+                      << " loc = " << origLoc << ", comp = " << origComp << "  =>  Mapped = " << newLoc << ", "
+                      << newComp << "\n");
       } else {
         inOutUsage.outputMapLocCount = std::max(inOutUsage.outputMapLocCount, newLoc + 1);
         LLPC_OUTS("(" << getShaderStageAbbreviation(m_shaderStage) << ") Output: loc = " << origLoc
-                      << "  =>  Mapped = " << newLoc << "\n");
+                      << ", comp = " << origComp << "  =>  Mapped = " << newLoc << ", " << newComp << "\n");
 
         if (m_shaderStage == ShaderStageFragment)
           outOrigLocs[newLoc] = origLoc;
@@ -2548,21 +2553,27 @@ void PatchResourceCollect::reassembleOutputExportCalls() {
   BuilderBase builder(*m_context);
   builder.SetInsertPoint(m_outputCalls.back());
 
+  // The output vector can have at most 4 elements, but 2 16-bit values can be packed a single element
+  static constexpr size_t MaxVectorComponents = 4;
+  static constexpr size_t MaxNumElems = MaxVectorComponents * 2;
   // ElementsInfo represents the info of composing a vector in a location
   struct ElementsInfo {
     // Elements to be packed in one location, where 32-bit element is placed at the even index
-    Value *elements[8];
+    std::array<Value *, MaxNumElems> elements = {};
     // The corresponding call of each element
-    CallInst *outCalls[8];
+    std::array<CallInst *, MaxNumElems> outCalls = {};
+    static_assert(std::tuple_size<decltype(elements)>::value == std::tuple_size<decltype(outCalls)>::value,
+                  "The code assumes that 'elements' and 'outCalls' have the same number of elements");
     // Element number of 32-bit
-    unsigned elemCountOf32bit;
+    unsigned elemCountOf32bit = 0;
     // Element number of 16-bit
-    unsigned elemCountOf16bit;
+    unsigned elemCountOf16bit = 0;
+    // First component index in the mapped vector.
+    unsigned baseMappedComponentIdx = InvalidValue;
   };
 
   // Collect ElementsInfo in each packed location
-  ElementsInfo elemsInfo = {{nullptr}, {nullptr}, 0, 0};
-  std::vector<ElementsInfo> elementsInfoArray(m_outputCalls.size(), elemsInfo);
+  std::vector<ElementsInfo> elementsInfoArray(m_outputCalls.size());
 
   const auto &tcsInputLocInfoMap =
       m_pipelineState->getShaderResourceUsage(ShaderStageTessControl)->inOutUsage.inputLocInfoMap;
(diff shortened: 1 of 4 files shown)


lgc/patch/PatchResourceCollect.cpp
```diff
@@ -2583,13 +2594,18 @@ void PatchResourceCollect::reassembleOutputExportCalls() {
...
     }
 
     const unsigned newLoc = mapIter->second.getLocation();
     auto &elementsInfo = elementsInfoArray[newLoc];
-    unsigned elemIdx = mapIter->second.getComponent() * 2 + mapIter->second.isHighHalf();
+
+    const unsigned origComponentIdx = mapIter->second.getComponent();
+    // Update the first components used
+    elementsInfo.baseOrigComponentIdx = std::min(elementsInfo.baseOrigComponentIdx, origComponentIdx);
+
```
xuechen417, line 2607: `mapIter->first` represents the original `InOutLocationInfo` and `mapIter->second` represents the mapped `InOutLocationInfo`. I think it's easy to get confused by using the `orig` prefix on the mapped `InOutLocationInfo`.

xuechen417 (commented):
Thank you for adding lgc test of input-output packing. The refactor of `reassembleOutputExportCalls` looks good to me. 
But I don't think your commit message is accurate. There is no assumption that the original vector is non-zero-based vector components. But the mapped vector is zero-based vector components after packing.

kuhar:
> But I don't think your commit message is accurate. There is no assumption that the original vector is non-zero-based vector components. But the mapped vector is zero-based vector components after packing.

Before this PR, the code unconditionally copied components 0..compCount to a new vector 0..compCount. If there was no element at 0, the code crashed. This change makes it copy components k..k + compCount to a new vector 0..compCount.
I don't know if this was a different assumption or an oversight in the implementation.

lgc/patch/PatchResourceCollect.cpp
kuhar: I renamed it to `baseMappedComponentIdx`. Hope it matches the naming of the other code better now.

kuhar:
V1: Rename `baseOrigComponentIdx` to `baseMappedComponentIdx`, based on feedback from @xuechen417.

xuechen417:
> > But I don't think your commit message is accurate. There is no assumption that the original vector is non-zero-based vector components. But the mapped vector is zero-based vector components after packing.
> 
> Before this PR, the code unconditionally copied components 0..compCount to a new vector 0..compCount. If there was no element at 0, the code crashed. This change makes it copy components k..k + compCount to a new vector 0..compCount.
> I don't know if this was a different assumption or an oversight in the implementation.

I'm not clear what kind of cases have your mentioned behavior. If the fragment shader has the output like this
layout(location = 0, component = 2) in vec2 vsOut0;
layout(location = 1) in vec2 vsOut1;
The alorigthm of InOutLocationInfoMapManager will do a maping 
(0, 2) -> (0, 0)
(0, 3) -> (0, 1)
(1, 0) -> (0, 2)
(1, 1) -> (0, 3)

kuhar:
@xuechen417 I'm not very familiar with LGC and may be explaining this in a confusing way.
If you give the LGC testace from the PR to unmodified lgc/llpc, it will crash when trying to use the 0th component:
```
#0  llvm::Value::getType (this=0x0) at external/llvm/include/llvm/IR/Value.h:246
#1  0x00000000027af568 in llvm::IRBuilderBase::CreateCast (this=0x7ffffffe83d8, Op=llvm::Instruction::BitCast, V=0x0, DestTy=0x4ca05f8, Name=...)
    at external/llvm/include/llvm/IR/IRBuilder.h:2195
#2  0x000000000283ff41 in lgc::PatchResourceCollect::reassembleOutputExportCalls (this=<optimized out>) at external/llpc/lgc/patch/PatchResourceCollect.cpp:2741
#3  0x000000000283e132 in lgc::PatchResourceCollect::packInOutLocation (this=0x4d53a60, isInput=<optimized out>) at external/llpc/lgc/patch/PatchResourceCollect.cpp:2496
#4  0x000000000283a251 in lgc::PatchResourceCollect::matchGenericInOut (this=0x4d53a60) at external/llpc/lgc/patch/PatchResourceCollect.cpp:1424
#5  0x0000000002837177 in lgc::PatchResourceCollect::processShader (this=0x4d53a60) at external/llpc/lgc/patch/PatchResourceCollect.cpp:1096
```

in this for loop:
```
for (unsigned compIdx = 0; compIdx < compCount; ++compIdx) {
      const unsigned elemIdx = compIdx * 2;
        Value *elems[2] = {elementsInfo.elements[elemIdx], elementsInfo.elements[elemIdx + 1]};
```
because the first filled element is at index 2:

```
; CHECK:       (VS) Output: loc = 0, comp = 3  =>  Mapped = 0, 0
; CHECK-NEXT:  (VS) Output: loc = 3, comp = 0  =>  Mapped = 0, 1   <-- elements[2]
; CHECK-NEXT:  (VS) Output: loc = 3, comp = 1  =>  Mapped = 0, 2   <-- elements[4]
; CHECK-NEXT:  (VS) Output: loc = 3, comp = 2  =>  Mapped = 0, 3   <-- elements[6]
```
while `elements[0]` is a nullptr.

Does this look like a case that's not handled by the current implementation to you?

xuechen417 (commented):
I don't understand InOutPackingNonZeroBase.lgc quite well.
I see the FS has two generic input calls. In PatchResourceCollect pass, the two v4f32 calls will be scalarized into 8 f32 calls, where two calls have users so they will join to build the InOutLocationInfoMap. The map for this case will be 
(3, 1)->(0,0)
(3, 2)->(0, 1)
The VS should has two corresponding generic output calls. They will be scalareized into 8 f32 calls. Depending on the map, the `elementsInfoArray` has one item where `elementsInfoArray[0].element[0] = usedScalarizedCall0` and `elementsInfoArray[0].element[1]=usedScalarizedCall0`. 
After `reassembleOutputExportCalls`, VS should generate a packed call @lgc.output.export.generic.i32.i32.v2f32(i32 3, i32 0, <2 x float> %1). 
After PatchInOutImportExport pass, VS will generate @llvm.amdgcn.exp.f32(i32 32, i32 3, float, float, undef, undef...

lgc/test/InOutPackingNonZeroBase.lgc
```diff
@@ -0,0 +1,130 @@
...
+
+; CHECK: %[[e0:.+]] = extractelement <3 x float> %[[bitcast]], i32 0
+; CHECK: %[[e1:.+]] = extractelement <3 x float> %[[bitcast]], i32 1
+; CHECK: %[[e2:.+]] = extractelement <3 x float> %[[bitcast]], i32 2
+
+  call void (...) @lgc.create.write.generic.output(<4 x float> %a7, i32 3, i32 0, i32 0, i32 0, i32 0, i32 undef)
+  call void (...) @lgc.create.write.builtin.output(<4 x float> %a7, i32 0, i32 0, i32 undef, i32 undef)
+
+; CHECK: call void @llvm.amdgcn.exp.f32(i32 immarg {{.+}}, i32 immarg {{.+}}, float %[[e0]], float %[[e1]], float %[[e2]], float 1.000000e+00, i1 immarg true, i1 immarg false)
+; CHECK: call void @llvm.amdgcn.exp.f32(i32 {{.+}}, i32 {{.+}}, float undef, float %[[e0]], float %[[e1]], float %[[e2]], i1 false, i1 false)
```
xuechen417, line 54: This kind of pattern will not be generated in packing mode. The undef value in exp instruction can only be placed at the back of defined value.

kuhar: These are the instructions from the original vertex shader:
```
  %71 = insertelement <4 x float> %66, float %70, i32 3
  %72 = insertelement <4 x float> undef, float %32, i32 0
  %73 = insertelement <4 x float> %72, float %37, i32 1
  %74 = insertelement <4 x float> %73, float %42, i32 2
  call void (...) @lgc.create.write.generic.output(<4 x float> %74, i32 3, i32 0, i32 0, i32 0, i32 0, i32 undef)
  call void (...) @lgc.create.write.builtin.output(<4 x float> %71, i32 0, i32 0, i32 undef, i32 undef)
  call void (...) @lgc.create.write.builtin.output(float undef, i32 1, i32 0, i32 undef, i32 undef)
  call void (...) @lgc.create.write.builtin.output([1 x float] undef, i32 3, i32 4096, i32 undef, i32 undef)
```

kuhar:
@xuechen417 Thank you for the explanations, but I can't follow your reasoning. This test is a minimized reproducer based on a crash happens on a few pipelines from a real game. These pipelines compile fine with the proprietary compiler and llpc with `-pack-in-out=false`. Since it's no longer possible to disable input-output packing, I'm trying to get this fixed.

I can't share the original pipeline or the shaders, but are there any other logs or prints that I could do and pass to you so that you could make sense of it?

lgc/test/InOutPackingNonZeroBase.lgc
kuhar: and from the fragment shader:
```
  %0 = call i8 addrspace(7)* (...) @lgc.create.load.buffer.desc.p7i8(i32 6, i32 4, i32 0, i32 0)
  %1 = call {}* @llvm.invariant.start.p7i8(i64 -1, i8 addrspace(7)* %0)
  %2 = call <4 x float> (...) @lgc.create.read.generic.input.v4f32(i32 0, i32 0, i32 0, i32 0, i32 16, i32 undef)
  %__llpc_input_proxy_.12.vec.extract = extractelement <4 x float> %2, i32 3

...

  %6 = call <4 x float> (...) @lgc.create.read.generic.input.v4f32(i32 3, i32 0, i32 0, i32 0, i32 16, i32 undef)
  %7 = shufflevector <4 x float> %6, <4 x float> undef, <3 x i32> <i32 0, i32 1, i32 2>
  ... ; %9 is based on %7
  %10 = extractelement <3 x float> %9, i32 0
  %11 = extractelement <3 x float> %9, i32 1
  %12 = extractelement <3 x float> %9, i32 2
  %13 = insertelement <4 x float> undef, float %10, i32 0
  %14 = insertelement <4 x float> %13, float %11, i32 1
  %15 = insertelement <4 x float> %14, float %12, i32 2
  ...
    %__llpc_output_proxy_.12.vec.insert = insertelement <4 x float> %15, float %18, i32 3
 ...
     call void (...) @lgc.create.write.generic.output(<4 x float> %__llpc_output_proxy_.0, i32 3, i32 0, i32 0, i32 0, i32 0, i32 undef)
```

kuhar: The mapping in the original pipeline is identical to the mapping from this test.

xuechen417: @kuhar Thank you for investigating and fixing it. I got your point. 
Yes, it is a case missing from the original implementation. I had thought a non-vertex input must match an output from a previous shader stage, but it is not true. 
`%2 = call <4 x float> (...) @lgc.create.read.generic.input.v4f32(i32 0, i32 0, i32 0, i32 0, i32 16, i32 undef)` has no matching output in the vertex shader. Although its value is no sense, it is allowed.

xuechen417 (approved):

Issue #1049, opened 2020-11-20
Relocatable shader corruption


Flakebi:
When using relocatable shaders (`EnableRelocatableShaders,1` in `.config/amdVulkanSettings.cfg`), the following fragment shader leads to corrupted output.

This could be related to wave64/wave32. The only difference in the pipeline I could spot was the value of `PS_W32_EN` in the `PS_IN_CONTROL` register. Forcing this value to `false` fixed the corruption.

This amber script can be used to reproduce the issue:
```glsl
#!amber
SHADER vertex vert_shader GLSL
#version 430

layout(location = 0) in vec4 position;
layout(location = 1) in vec4 vert_color;
layout(location = 0) out vec4 frag_color;

void main() {
  gl_Position = position;
  frag_color = vert_color;
}
END

SHADER fragment frag_shader GLSL
#version 450
#extension GL_ARB_separate_shader_objects : enable

layout(location = 0) in vec4 fragColor;

layout(location = 0) out vec4 outColor;

void main() {
  // If the size of this array gets less than 16 or the loop gets run less than 13 times, optimizations make the corruption go away
  float[20] fs;
  for (int i = 0; i < 20; i++)
    fs[i] = i;
  float fac = fs[min(abs(int(fragColor.x)), 20)] / 20;
  // The value of fac will often be different for relocatable shaders
  outColor = vec4(fac, 1.0, 1.0, 1.0);
}
END

BUFFER position_buf DATA_TYPE R8G8_SNORM DATA
# Full frame
-128 -128
 127  127
-128  127
-128 -128
 127  127
 127 -128
END

BUFFER vert_color DATA_TYPE R8G8B8A8_UNORM DATA
255   0   0 0
255   0   0 0
255   0   0 0
255   0   0 0
255   0   0 0
255   0   0 0
END


BUFFER img_buf FORMAT B8G8R8A8_UNORM
PIPELINE graphics my_pipeline
  ATTACH vert_shader
  ATTACH frag_shader

  VERTEX_DATA position_buf LOCATION 0
  VERTEX_DATA vert_color LOCATION 1

  BIND BUFFER img_buf AS color LOCATION 0
END

CLEAR my_pipeline

RUN my_pipeline DRAW_ARRAY AS TRIANGLE_LIST START_IDX 0 COUNT 6
EXPECT img_buf IDX 0 0 SIZE 250 250 EQ_RGBA 13 255 255 255
```

[...]

PR #1198, opened 2021-03-25
Don't use the PS_IN_CONTROL from glue shaders when linking

s-perron:
The PS_IN_CONTROL for the glue shaders does not have all of the data
needed to geneerate the value correctly, so it needs to be ignored.

Fixes https://github.com/GPUOpen-Drivers/llpc/issues/1049


amdrexu (approved):
LGTM

PR #1200, opened 2021-03-31
Optimize transform feedback exports

amdrexu:
Previously, we always split a floating-point or integer array to
separate elements. This will result in unnecessary tbuffer stores.
See this example:

  layout(xfb_buffer = 0, xfb_offset = 0, xfb_stride = 128, location = 0)
  out float f[26];

We will finally get 26 tbuffer_x stores. Rather, we ought to get 6
tbuffer_xyzw stores and 1 tbuffer_xy store.

Change-Id: I1030fc924032bcb2dae4032c5fb79cd437134e48


amdrexu:
Close this PR. Will redesign the processing. The XFB export is mixed with normal output export. The lower pass is not a good place to handle this.

Issue #1213, opened 2021-04-13
Build failure with -Werror after #1201


dnovillo:
Our builds are configured with `-Werror`.  They are now failing with

```
llpc/llpc/translator/lib/SPIRV/SPIRVToLLVMDbgTran.cpp:907:16: error: using the result of an assignment as a condition without parentheses [-Werror,-Wparentheses]
  if (LineInfo = SpirvInst->getLine()) {
      ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

If think the assignment inside the `if()` is intended, given the rest of the logic.  Should this also add `!= nullptr`?

nhaehnle:
This is outdated.

PR #1214, opened 2021-04-14
Fix issue: #1213

jiaolu:
Some c++ compiler would treat -Wparentheses as error, maybe we can add
-Wno-parentheses in the compiling options


amdrexu (approved):

ruiling (commented):
Please change the title of the commit into something like "Fix build failure under -Werror". Add a note "Fixes: #1213" in a new line at the end of the commit message.

trenouf:
Is this superseded by #1216 ?

kuhar:
> Is this superseded by #1216 ?

Even if it is, I'd say that the code after this PR looks less suspicious, so I'd still like it to land.

PR #1215, opened 2021-04-14
Enable -Werror for llpc

JaxLinAMD:


amdrexu (approved):

Flakebi:
Can we have `-Werror` off by default and enable it everywhere we want it? (CI, developer builds, etc.)
Having it on by default usually creates problems when other people with different compiler versions or different compilers start to build a project. Even if we do not have any warnings in our CI, using a newer compiler can add warnings and fail compilation if we enable `-Werror` by default. This makes it harder for packagers and for users that just want to compile the driver or compiler.

JaxLinAMD:
got it. close it.

Issue #1203, opened 2021-04-08
New CI build failures after the latest LLPC and XGL merges


kuhar:
Seems like the latest imports broke LLPC and XGL CI:
- https://github.com/GPUOpen-Drivers/llpc/actions/runs/725183732
- https://github.com/GPUOpen-Drivers/llpc/actions/runs/725930478

I opened a PR for the XGL cache-info warning: https://github.com/GPUOpen-Drivers/xgl/pull/109 and a PAL issue for unrecognized compiler flags: https://github.com/GPUOpen-Drivers/pal/issues/62

kuhar:
cc @JaxLinAMD @Flakebi @amdrexu

amdrexu:
Thank you. Jax will take a look and resolve it.

JaxLinAMD:
@kuhar  what version of gcc is used? I didn't observe compile failure on ubuntu 18.04

Flakebi:
The GitHub runners use Ubuntu 20.04 and gcc 9.3.0.

PR #1216, opened 2021-04-14
Do not run CI with warnings-as-errors

kuhar:
Werror causes too much havoc when llpc/xgl cmake change.

Also remove a build flag set to its default value.

Fixes: https://github.com/GPUOpen-Drivers/llpc/issues/1203


dstutt (approved):
LGTM - less havoc is always good :)

PR #1246, opened 2021-05-18
[CI] Install openssh-client in docker

kuhar:
This fixes recent CI failures, e.g.,
https://github.com/GPUOpen-Drivers/llpc/runs/2605831264?check_suite_focus=true.

### DIFF
--- a/docker/amdvlk.Dockerfile
+++ b/docker/amdvlk.Dockerfile
@@ -45,7 +45,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
        libssl-dev libx11-dev libxcb1-dev x11proto-dri2-dev libxcb-dri3-dev \
        libxcb-dri2-0-dev libxcb-present-dev libxshmfence-dev libxrandr-dev \
        libwayland-dev \
-       git curl wget \
+       git curl wget openssh-client \
     && rm -rf /var/lib/apt/lists/* \
     && python3 -m pip install --no-cache-dir --upgrade pip \
     && python3 -m pip install --no-cache-dir --upgrade cmake \


kuhar:
The CI failures under this PR are not caused by the PR itself. I think they should go away once this PR gets merged and a new nightly image is produced.

Flakebi (approved):
Thanks!

Flakebi:
I wonder how this happens, openssh-client is a dependency of git: https://packages.ubuntu.com/focal/git

kuhar:
> I wonder how this happens, openssh-client is a dependency of git: https://packages.ubuntu.com/focal/git

I don't know much about `apt`, but ssh-listed is listed as a `recommended` dependency (green) instead of depends (red). So it may not be a hard dependency installed by default with git?

Flakebi:
> I don't know much about apt, but ssh-listed is listed as a recommended dependency (green) instead of depends (red). So it may not be a hard dependency installed by default with git?

Ah, right, I missed that. The script also uses `--no-install-recommends`.

trenouf:
I will merge this so we can get CI working again.

trenouf:
I tried rerunning CI here: https://github.com/GPUOpen-Drivers/llpc/pull/1237/checks?check_run_id=2617274642
and it still says

warning: repo is not tracking a remote branch, so it will not receive updates
repo reset: error: Entry 'docs/internal-fs-layout.md' not uptodate. Cannot merge.
fatal: Could not reset index file to revision '339f2df1ddd741070e340ec01d6882dd1eee617c'.

Is that unrelated to what you fixed here? Or do I need to wait until a new nightly image is produced?

PR #1276, opened 2021-06-09
Fix an issue of exporting GS outputs without packing

amdrexu:
We don't set stream ID when filling the location info structure
'InOutLocationInfo' if I/O packing is disabled for GS. This is
a mistake and will cause transform feedback failures because GS
output exporting in non-zero streams are always skipped.

Change-Id: If98d8b2ee289729cb146d720987aa6e9c30b09db


trenouf (approved):
LGTM

linqun (approved):

PR #1288, opened 2021-06-16
Remove the resource collect pass from lowering

bsaleil:
It seems that this pass is used to collect information about modules and to
remove dead global variables. In LegacySpirvLower::addPasses, the pass is almost
directly followed by LLVM's global DCE pass that can also remove these globals.
Also, it seems the collected information is never used for this instance, so the
pass can probably be removed from there. I tried compiling 10K+ shaders with and
without the pass and the generated code is exactly the same in all cases.

### DIFF
--- a/llpc/lower/llpcSpirvLowerResourceCollect.h
+++ b/llpc/lower/llpcSpirvLowerResourceCollect.h
@@ -68,14 +68,15 @@ public:
   bool detailUsageValid() { return m_detailUsageValid; }
 
   virtual bool runOnModule(llvm::Module &module);
-  void visitCalls(llvm::Module &module);
-  llvm::Value *findCallAndGetIndexValue(llvm::Module &module, llvm::CallInst *const targetCall);
 
   static char ID; // ID of this pass
 
 private:
   SpirvLowerResourceCollect(const SpirvLowerResourceCollect &) = delete;
   SpirvLowerResourceCollect &operator=(const SpirvLowerResourceCollect &) = delete;
+
+  void visitCalls(llvm::Module &module);
+  llvm::Value *findCallAndGetIndexValue(llvm::Module &module, llvm::CallInst *const targetCall);
 
   unsigned getFlattenArrayElementCount(const llvm::Type *ty) const;
   const llvm::Type *getFlattenArrayElementType(const llvm::Type *ty) const;
--- a/llpc/lower/llpcSpirvLower.cpp
+++ b/llpc/lower/llpcSpirvLower.cpp
@@ -154,9 +154,6 @@ void LegacySpirvLower::addPasses(Context *context, ShaderStage stage, legacy::Pa
   if (lowerTimer)
     passMgr.add(LgcContext::createStartStopTimer(lowerTimer, true));
 
-  // Lower SPIR-V resource collecting
-  passMgr.add(createSpirvLowerResourceCollect(false));
-
   // Function inlining. Use the "always inline" pass, since we want to inline all functions, and
   // we marked (non-entrypoint) functions as "always inline" just after SPIR-V reading.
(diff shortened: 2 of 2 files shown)


llpc/lower/llpcSpirvLowerResourceCollect.h
```diff
@@ -68,15 +68,16 @@ class SpirvLowerResourceCollect : public LegacySpirvLower, public llvm::InstVisi
...
-  llvm::Value *findCallAndGetIndexValue(llvm::Module &module, llvm::CallInst *const targetCall);
 
   static char ID; // ID of this pass
 
 private:
   SpirvLowerResourceCollect(const SpirvLowerResourceCollect &) = delete;
   SpirvLowerResourceCollect &operator=(const SpirvLowerResourceCollect &) = delete;
 
+  void visitCalls(llvm::Module &module);
+  llvm::Value *findCallAndGetIndexValue(llvm::Module &module, llvm::CallInst *const targetCall);
```
kuhar, line 79: nit: the `const` can be dropped here. It only applies to the pointer, not to the pointee. This won't even change the function signature.

amdvlk-admin:
### Test summary for commit 6ac514e66a103ab43ae43c161ed5a760e54fe272

<li>PAL: dev 7f6e731aa5193a518356bdfddc75571d032a3928</li>
<li>SPVGEN: dev 79f8e32dd7242d7d34fa0074f761c9d794f11811</li>
<li>XGL: dev 290c9fc19eab37d76cbbffa718782210f6760fed</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 5f0d44ad8891d9d3b202f7175970ba305ac37506</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/228128)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 31146/57032 (54.6%)</li>
<li>Failed: 0/57032 (0.0%)</li>
<li>Not Supported: 25886/57032 (45.4%)</li>
</ul></details>

<li>Passed: 31145/57032 (54.6%)</li>
<li>Failed: 0/57032 (0.0%)</li>
<li>Not Supported: 25887/57032 (45.4%)</li>
</ul></details>

<li>Passed: 31285/57032 (54.9%)</li>
<li>Failed: 0/57032 (0.0%)</li>
<li>Not Supported: 25747/57032 (45.1%)</li>
</ul></details>

<li>Passed: 34853/57032 (61.1%)</li>
<li>Failed: 0/57032 (0.0%)</li>
<li>Not Supported: 22179/57032 (38.9%)</li>
</ul></details>


</ul></details>

amdrexu:
Do you plan to remove the whole class from LLPC? Resource collecting was originally designed to make pipeline usage analysis. But later, some functionalities are moved to LGC builder and other are gone into LGC resource collecting because at this phase the analysis is more reliable and doable. If you didn't encounter any issue,  I think we can eliminate the class but this is better to be done after your change is merged and there is no regression found by further QA testing.

llpc/lower/llpcSpirvLowerResourceCollect.h
piotrAMD: Good point. The function always returns nullptr anyway (and has been for ~1 year) so I suggest to remove it entirely - now or in a pr where the class will be removed.

piotrAMD (approved):
LGTM. Good idea to wait a week or so and remove the rest, as Rex suggested.

amdrexu (approved):

bsaleil:
I guess we could remove the whole class from LLPC. My only concern is that this pass is also used when the `-enable-shader-module-opt` option is provided. However, that option seems to be broken, so I'll not be able to run tests to ensure removing the pass doesn't break anything. I think we should fix the option before removing that class.

PR #1292, opened 2021-06-21
Revert "Remove output export call if the output value is undefined"

amdrexu:
This reverts commit 409ba7456efd70cf59aca7738d791c4efcb53ec2.


piotrAMD (approved):

amdrexu:
> @amdrexu why was this reverted?

The pre check-in testing is not a full list. And we encounter some transform feedback failures in the group transform_feedback.fuzz.* after code submission (post check-in runs much more tests). Most failures are GS cases. This change triggers a assertion and the root cause is found. I will resubmit a review with new updates.

PR #1316, opened 2021-07-08
Don't update output loc info for undef outputs.

s-perron:
When there is an explict store to a vertex shader output that is an
undef value, then it will be removed.  That is fair.  However, the
output location map is updated to remove this output dropping the
channel that is used for other output.  This does not work for unlinked
shaders because the pixel shader will then read its inputs from the
wrong channel.

### DIFF
--- a/lgc/patch/PatchResourceCollect.cpp
+++ b/lgc/patch/PatchResourceCollect.cpp
@@ -1175,20 +1175,22 @@ void PatchResourceCollect::visitCallInst(CallInst &callInst) {
       // export call.
       m_deadCalls.push_back(&callInst);
 
-      InOutLocationInfo outLocInfo;
-      outLocInfo.setLocation(cast<ConstantInt>(callInst.getArgOperand(0))->getZExtValue());
-      outLocInfo.setComponent(cast<ConstantInt>(callInst.getArgOperand(1))->getZExtValue());
-      if (m_shaderStage == ShaderStageGeometry)
-        outLocInfo.setStreamId(cast<ConstantInt>(callInst.getArgOperand(2))->getZExtValue());
-      // Also, we remove the output location info from the map if it exists
-      auto &outLocInfoMap = m_resUsage->inOutUsage.outputLocInfoMap;
-      if (outLocInfoMap.count(outLocInfo) > 0)
-        outLocInfoMap.erase(outLocInfo);
-      // For GS, we remove transform feedback location info as well if it exists
-      if (m_shaderStage == ShaderStageGeometry) {
-        auto &locInfoXfbOutInfoMap = m_resUsage->inOutUsage.gs.locInfoXfbOutInfoMap;
-        if (locInfoXfbOutInfoMap.count(outLocInfo) > 0)
-          locInfoXfbOutInfoMap.erase(outLocInfo);
+      if (m_pipelineState->canPackOutput(m_shaderStage)) {
+        InOutLocationInfo outLocInfo;
+        outLocInfo.setLocation(cast<ConstantInt>(callInst.getArgOperand(0))->getZExtValue());
+        outLocInfo.setComponent(cast<ConstantInt>(callInst.getArgOperand(1))->getZExtValue());
+        if (m_shaderStage == ShaderStageGeometry)
+          outLocInfo.setStreamId(cast<ConstantInt>(callInst.getArgOperand(2))->getZExtValue());
+        // Also, we remove the output location info from the map if it exists
+        auto &outLocInfoMap = m_resUsage->inOutUsage.outputLocInfoMap;
+        if (outLocInfoMap.count(outLocInfo) > 0)
+          outLocInfoMap.erase(outLocInfo);
+        // For GS, we remove transform feedback location info as well if it exists
+        if (m_shaderStage == ShaderStageGeometry) {
+          auto &locInfoXfbOutInfoMap = m_resUsage->inOutUsage.gs.locInfoXfbOutInfoMap;
+          if (locInfoXfbOutInfoMap.count(outLocInfo) > 0)
+            locInfoXfbOutInfoMap.erase(outLocInfo);
+        }
       }
     } else {
       m_outputCalls.push_back(&callInst);
--- a/llpc/test/shaderdb/UndefVertexOutput.spvasm
+++ b/llpc/test/shaderdb/UndefVertexOutput.spvasm
@@ -1,0 +1,55 @@
+; Make sure that the export channels match the location in the spir-v.
+; This is how the PS knows which channel from which to read the value.
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: _amdgpu_vs_main_fetchless:
+; SHADERTEST: v_mov_b32_e32 [[zero:v[0-9]*]], 0
+; SHADERTEST: v_mov_b32_e32 [[one:v[0-9]*]], 1.0
+; SHADERTEST-DAG: exp param8 [[one]], [[zero]], [[zero]], [[one]]
+; SHADERTEST-DAG: exp param7 [[zero]], [[zero]], [[one]], [[one]]
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
+
+               OpCapability Shader
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Vertex %2 "main" %3 %gl_Position %5 %6 %7
+          %8 = OpString ""
+               OpDecorate %3 Location 1
+               OpDecorate %3 Binding 0
+               OpDecorate %gl_Position Binding 0
+               OpDecorate %gl_Position BuiltIn Position
+               OpDecorate %5 Location 5
+               OpDecorate %5 Binding 0
+               OpDecorate %6 Location 7
+               OpDecorate %6 Binding 0
+               OpDecorate %7 Location 8
+               OpDecorate %7 Binding 0
+       %void = OpTypeVoid
+         %10 = OpTypeFunction %void
+      %float = OpTypeFloat 32
+    %v3float = OpTypeVector %float 3
+    %v4float = OpTypeVector %float 4
+%_ptr_Input_v4float = OpTypePointer Input %v4float
+          %3 = OpVariable %_ptr_Input_v4float Input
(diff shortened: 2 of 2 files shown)


amdvlk-admin:
### Test summary for commit 1d6b8873621e7e89d6139241c8a562eddaa77eac

<li>PAL: dev fa6b0b7adbe0cac60f9a2c1853783eb03ec4b7a5</li>
<li>SPVGEN: dev bb3da8bc00bcf4373d563bf9b3c74a80a8b28fe6</li>
<li>XGL: dev 57f8643a621a3f4d2db663539b2eeeb9d16650db</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 0a3fa7fcc965c8e6d37ad16469bbf4ea8caf90ee</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/229120)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 31315/57280 (54.7%)</li>
<li>Failed: 0/57280 (0.0%)</li>
<li>Not Supported: 25965/57280 (45.3%)</li>
</ul></details>

<li>Passed: 31314/57280 (54.7%)</li>
<li>Failed: 0/57280 (0.0%)</li>
<li>Not Supported: 25966/57280 (45.3%)</li>
</ul></details>

<li>Passed: 31628/57280 (55.2%)</li>
<li>Failed: 0/57280 (0.0%)</li>
<li>Not Supported: 25652/57280 (44.8%)</li>
</ul></details>

<li>Passed: 35022/57280 (61.1%)</li>
<li>Failed: 0/57280 (0.0%)</li>
<li>Not Supported: 22258/57280 (38.9%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchResourceCollect.cpp
```diff
@@ -1175,20 +1175,22 @@ void PatchResourceCollect::visitCallInst(CallInst &callInst) {
...
-      // Also, we remove the output location info from the map if it exists
-      auto &outLocInfoMap = m_resUsage->inOutUsage.outputLocInfoMap;
-      if (outLocInfoMap.count(outLocInfo) > 0)
-        outLocInfoMap.erase(outLocInfo);
-      // For GS, we remove transform feedback location info as well if it exists
-      if (m_shaderStage == ShaderStageGeometry) {
-        auto &locInfoXfbOutInfoMap = m_resUsage->inOutUsage.gs.locInfoXfbOutInfoMap;
-        if (locInfoXfbOutInfoMap.count(outLocInfo) > 0)
-          locInfoXfbOutInfoMap.erase(outLocInfo);
+      if (!m_pipelineState->isUnlinked()) {
```
xuechen417, line 1178: How about change the condition to m_pipelineState->canPackInOut(m_shaderStage)? The modification of locInfoMap/locInfoXfboutInfoMap is required for packing use.

s-perron: Good call.  Thanks.  I have made the change.

s-perron:
V2: Made the change suggested in https://github.com/GPUOpen-Drivers/llpc/pull/1316#discussion_r666687003.

amdvlk-admin:
### Test summary for commit dc6230b961cc1b2608cf5c1bac3dd6ac4a16551f

<li>PAL: dev fa6b0b7adbe0cac60f9a2c1853783eb03ec4b7a5</li>
<li>SPVGEN: dev bb3da8bc00bcf4373d563bf9b3c74a80a8b28fe6</li>
<li>XGL: dev 57f8643a621a3f4d2db663539b2eeeb9d16650db</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 01f326c1e88cbf80e50a631ea0ce61a57152a7c0</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/227984)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 31680/56996 (55.6%)</li>
<li>Failed: 0/56996 (0.0%)</li>
<li>Not Supported: 25316/56996 (44.4%)</li>
</ul></details>

<li>Passed: 31677/56996 (55.6%)</li>
<li>Failed: 0/56996 (0.0%)</li>
<li>Not Supported: 25319/56996 (44.4%)</li>
</ul></details>

<li>Passed: 32051/56996 (56.2%)</li>
<li>Failed: 0/56996 (0.0%)</li>
<li>Not Supported: 24945/56996 (43.8%)</li>
</ul></details>

<li>Passed: 34932/56996 (61.3%)</li>
<li>Failed: 0/56996 (0.0%)</li>
<li>Not Supported: 22064/56996 (38.7%)</li>
</ul></details>


</ul></details>

xuechen417:
LGTM.

JaxLinAMD:
@amdrexu @trenouf PING for code review

lgc/patch/PatchResourceCollect.cpp
```diff
@@ -1175,20 +1175,22 @@ void PatchResourceCollect::visitCallInst(CallInst &callInst) {
...
-      // Also, we remove the output location info from the map if it exists
-      auto &outLocInfoMap = m_resUsage->inOutUsage.outputLocInfoMap;
-      if (outLocInfoMap.count(outLocInfo) > 0)
-        outLocInfoMap.erase(outLocInfo);
-      // For GS, we remove transform feedback location info as well if it exists
-      if (m_shaderStage == ShaderStageGeometry) {
-        auto &locInfoXfbOutInfoMap = m_resUsage->inOutUsage.gs.locInfoXfbOutInfoMap;
-        if (locInfoXfbOutInfoMap.count(outLocInfo) > 0)
-          locInfoXfbOutInfoMap.erase(outLocInfo);
+      if (m_pipelineState->canPackOutput(m_shaderStage)) {
```
amdrexu, line 1178: Sorry. I didn't notice your discussion with Xue previously. This piece of code is added by me. I want to remove unused output components to save buffer/LDS store. And also reduce transform feedback outputs. This is for both packed pipeline and unpacked pipeline. But I didn't realize this will cause problems for unlinked pipeline. So please change the condition to `if (!unlinked)`. That is my real purpose.

trenouf: I think Xue's suggestion (that Steven has implemented) of using canPackOutput is better, because it allows for the case of compiling a part-pipeline, where you cannot pack the final-vertex-stage shader's outputs but you can pack other outputs (e.g. VS -> TCS).

trenouf (approved):
LGTM

lgc/patch/PatchResourceCollect.cpp
s-perron: I agree with Tim.  Using "canPackOutput" seems to make the most sense.  Removing the location from the map is packing the output.  So this code should only be executed if you can pack the input.

amdrexu: Thanks all for your reply. Let explain this issue:
In the function `visitCallInst`, we have a branch to remove undefined transform feedback values. Please see this:

```
    auto outputValue = callInst.getArgOperand(callInst.getNumArgOperands() - 1);
    if (isa<UndefValue>(outputValue)) {
      // NOTE: If an output value is undefined, we can safely drop it and remove the transform feedback output export
      // call.
      m_deadCalls.push_back(&callInst);
```

But we have already set up the map `locInfoXfbOutInfoMap` when builder is called. See `CreateWriteXfbOutput`. Therefore, we have to remove the in/out pack info from the maps `outputLocInfoMap` and `locInfoXfbOutInfoMap`.

Note that this is irrelevant to pack status. Even though the output is not packed, we still have to remove such undefined transform feedback values and their in/out location map items. Hence, if you use the condition `canPackOutput`, this piece of code will not be triggered. And in `PatchCopyShader::exportOutput`, when we traverse the map, the assert `assert(outputLocInfoMap.count(origLocInfo))` will be hit. This is why I suggest we use `unlinked` as the condition.

For future transform feedback implementation, we actually disable GS output packing. Therefore, all this piece of codes will be skipped entirely. And finally the assert in copy shader is hit.

s-perron: This sounds like something where should be a test for.  Can you create one?  That way I'll be able to see what is going on for myself.

amdrexu: Please see this pipeline from the CTS: dEQP-VK.transform_feedback.fuzz.various_buffers.buffers110_instance_array_geometry

Some outputs are undefined values because they are not written by GS. To reproduce this issue, you must add those code to `InOutBuilder::CreateWriteXfbOutput`:

```
if (m_shaderStage == ShaderStageGeometry) {
    getPipelineState()->setPackOutput(ShaderStageGeometry, false);
    getPipelineState()->setPackInput(ShaderStageFragment, false);
   ...
}

This will cause the copy shader generation to go to unpacked path. If you use `canPackOutput` rather than `!isUnlinked`, the removal of map items will not be executed. And this will finally cause an assert in copy shader.
[xfb_fuzz.txt](https://github.com/GPUOpen-Drivers/llpc/files/6809782/xfb_fuzz.txt)

```

s-perron: I think I see the issue now.  Correct me if I am wrong.  "Packing" output is taking two or more output and merging them into a single output.  For example taking two vec2 outputs and outputting at a single vec4 output.  That is not what this code is doing.

I'll change this to check the next shader stage.  If the next stage is invalid, then we will not remove the entry in the map.  This should the partial-pipeline scenario that Tim mentioned to still work, and fix the problem that I am seeing.  Let me know if this is reasonable.

Also, we should have lit tests as much as possible.  They are much easier to run and debug than CTS.  The lit test can also be very specific with comments that explain what is suppose to happen if I were to do something that breaks it.  If something breaks in CTS, which in this case it will not because we need more changes to the repo, I do not have something guiding me on what is suppose to happen except the Vulkan SPEC.

amdrexu:
@JaxLinAMD Hi Jax, please revert this change. Steven will rework it a little. Otherwise, this will break our internal testing for further transform feedback implementation. Thanks.

lgc/patch/PatchResourceCollect.cpp
amdrexu: > Correct me if I am wrong. "Packing" output is taking two or more output and merging them into a single output. For example taking two vec2 outputs and outputting at a single vec4 output. That is not what this code is doing.

Yes, correct. This code is to clean up undefined outputs whether it is packed or unpacked. Because we remove the output call for those undefined values, we must remove them from the map items as well.

> I'll change this to check the next shader stage. If the next stage is invalid, then we will not remove the entry in the map. This should the partial-pipeline scenario that Tim mentioned to still work, and fix the problem that I am seeing. Let me know if this is reasonable.

This could be a solution.

JaxLinAMD:
@amdrexu sorry. I reverted it.

PR #1326, opened 2021-07-21
Fail to link reloc shader because of builtin fs input

s-perron:
When a builtin is used that LLPC must assign a HW channel to pass it
from the vertex processing stage to the fragment stage, relocatable
shaders could fail.  This will happen when the vertex outputs do not
match the fragment inputs exactly.  In this case, LLPC has no way to
consistently assign a channel to the builtin.

For a quick solution to this problem, we will fail when trying to link
relocatable shaders that have builtin FS input, and then fall back to
whole pipeline compilation.


trenouf (approved):
LGTM

amdrexu (approved):

s-perron:
@JaxLinAMD The two failing checks are not because of my change.  See https://github.com/GPUOpen-Drivers/llpc/pull/1327

PR #1367, opened 2021-08-23
Handle upstream changes for hasFnAttrs

dstutt:
Guard for a previous fix was using a version that was too late (so missed some
cases).


dstutt:
Issue was actually elsewhere - this change is not required

Issue #725, opened 2020-05-27
SpirvLowerGlobal shouldn't visit the whole module


jayfoad:
The SpirvLowerGlobal pass visits the whole module to find loads and stores to global variables, and call and return instructions. I think it would be more efficient to find the load and store instructions by iterating over the users of the global variable (and recursively the users of any getelementptr instructions), and to find the call instructions by iterating over users of the callee, and to find the return instructions by looking at the terminator of each basic block. This should be faster when visiting very large functions with very few load/store/call/return instructions.

amdrexu:
The original visitors were always based on module not function uses. At the beginning of LLPC, we don't have enough time to evaluate the performance. Feel free to make changes towards the right direction.

bsaleil:
Fixed by https://github.com/GPUOpen-Drivers/llpc/pull/1537

PR #1433, opened 2021-10-08
Remove parts of the SpirvLowerGlobal pass

bsaleil:
This patch remove parts of the `SpirvLowerGlobal` pass. This pass handles chained GEP instructions, but these chained instructions are already coalesced by the SpirvLowerAccessChain pass. The patch also cleans a bit the code.
This is preliminary work to fix https://github.com/GPUOpen-Drivers/llpc/issues/725.


llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -348,6 +348,26 @@ void SpirvLowerGlobal::visitCallInst(CallInst &callInst) {
   }
 }
 
+// =====================================================================================================================
+// Check if the given metadata value has a vertex index.
+//
+// @param metaVal : Metadata
+static bool hasVertexIdx(const llvm::Constant &metaVal) {
```
amdrexu, line 355: Could remove the llvm:: prefix in the cpp files.

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -348,6 +348,26 @@ void SpirvLowerGlobal::visitCallInst(CallInst &callInst) {
...
+  assert(metaVal.getNumOperands() == 4);
+  ShaderInOutMetadata inOutMeta = {};
+  inOutMeta.U64All[0] = cast<ConstantInt>(metaVal.getOperand(2))->getZExtValue();
+  inOutMeta.U64All[1] = cast<ConstantInt>(metaVal.getOperand(3))->getZExtValue();
+
+  if (inOutMeta.IsBuiltIn) {
```
amdrexu, line 361: According to the suggestion of Jakub, the style of this could be:
```
if (...) {
  return ...
}

return ...
```
Move the return in else branch outside. Make sure we have a return unconditionally in the end the function.

PR #1442, opened 2021-10-15
Fix namespace name CodeObjectMetadataKey to PalCodeObjectMetadataKey.

vettoreldaniele:
The class `CodeObjectMetadataKey` does not exist. The correct name is `PalCodeObjectMetadataKey`.


amdvlk-admin:

s-perron (approved):

PR #1443, opened 2021-10-18
Reorganize multi-input shaderdb tests

kuhar:
Change the file extension from `.ll` to `.mutli-input` and add a
README clarifying this convention.

We intend to add more tests in this folder.

PR #1444, opened 2021-10-18
Fix input error reporting format in amdllpc

kuhar:
Add a negative LIT test for mixing .pipe and shader inputs.

Depends on https://github.com/GPUOpen-Drivers/llpc/pull/1443 which needs to be merged first.


dstutt (approved):
LGTM

amdrexu (approved):

PR #1479, opened 2021-11-01
Fix a mistach issue of LLPC/LGC shader stage mask

amdrexu:
The method GraphicsShaderCacheChecker::check() is called by the
middle-end pass PatchCheckShaderCache. However, the input stage
mask and the stage hashes are organized according to LGC shader stage
enumerant order. The order is not necessarily identical to that of
LLPC. We have to do conversion.

This issue is latent because we always have the same order. This
assumption is not ture if either LLPC or LGC introduces new stages
while the counterpart is not.

Change-Id: Id21a2053d752eea6ec4c97dc8ddb1c080ed4e0ca

### DIFF
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -1120,7 +1120,7 @@ Result Compiler::buildPipelineInternal(Context *context, ArrayRef<const Pipeline
         }
       }
     }
-    SmallVector<Module *, 5> modulesToLink;
+    SmallVector<Module *, ShaderStageGfxCount> modulesToLink;
     for (unsigned shaderIndex = 0; shaderIndex < shaderInfo.size() && result == Result::Success; ++shaderIndex) {
       // Per-shader SPIR-V lowering passes.
       const PipelineShaderInfo *shaderInfoEntry = shaderInfo[shaderIndex];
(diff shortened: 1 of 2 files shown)


kuhar (commented):
While the fix looks fine to me, these seems very fragile. Would it be possible to use strongly typed enums instead of `unsigned` for lgc stages?

JaxLinAMD:
retest this please

amdrexu:
> Would it be possible to use strongly typed enums instead of `unsigned` for lgc stages?

Yes, I plan to do a future change to add ShaderStageBit enum in LGC and the shader stage mask should be expressed as the combination of these enumerants. This is similar to `VkShaderStageFlagBits`.

llpc/context/llpcCompiler.cpp
```diff
@@ -1251,32 +1251,32 @@ unsigned GraphicsShaderCacheChecker::check(const Module *module, const unsigned
   Compiler::buildShaderCacheHash(m_context, stageMask, stageHashes, &fragmentHash, &nonFragmentHash);
   unsigned stagesLeftToCompile = stageMask;
 
-  if (stageMask & shaderStageToMask(ShaderStageFragment)) {
+  if (stageMask & (1 << lgc::ShaderStageFragment)) {
```
kuhar, line 1254: Could we have this as a new function, e.g., `lgcStageToMask()`?

amdrexu: Sure. Will modify.

kuhar (approved):

linqun (approved):

PR #1498, opened 2021-11-05
Fix LGC tests after changes to the AMDGPU_Gfx calling convention.

tsymalla:
This patch aims to fix the LGC tests which are failing because of a
recent LLVM change 76cbe62262a3ccd0604599a1103ce5d38ac0164b.
Here, the AMDGPU_Gfx calling convention was changed so that some
SGPRs are now callee-save, the resulting ISA is different now.
The LGC tests are changed in a way so they work both with and
without the LLVM change.


amdvlk-admin:

Flakebi (approved):

trenouf (approved):
LGTM. I will merge this.

PR #1499, opened 2021-11-05
Remove unused output for some unlinked shaders.

s-perron:
The TES inputs and TCS outputs are not matching up when compiling to
unlinked shaders.  This is because the compiler is inconsistent on when
it will try to pack the inputs and outputs.  It is also inconsistent on
how unused TES inputs are mapped.  This change tries to make that this
consistent.

The idea behind this change is to always remove unused output of the
next stage is available, even if we are doing unlinked shaders.

### DIFF
--- a/lgc/builder/InOutBuilder.cpp
+++ b/lgc/builder/InOutBuilder.cpp
@@ -313,7 +313,7 @@ void InOutBuilder::markGenericInputOutputUsage(bool isOutput, unsigned location,
   if (!isOutput || m_shaderStage != ShaderStageGeometry) {
     bool keepAllLocations = false;
     if (getPipelineState()->isUnlinked()) {
-      if (m_shaderStage == ShaderStageVertex && isOutput)
+      if (isOutput && m_pipelineState->getNextShaderStage(m_shaderStage, true) == ShaderStageFragment)
         keepAllLocations = true;
       if (m_shaderStage == ShaderStageFragment && !isOutput)
         keepAllLocations = true;
--- a/llpc/test/shaderdb/relocatable_shaders/PipelineTess_RelocRemoveUnusedTcsOutputs.pipe
+++ b/llpc/test/shaderdb/relocatable_shaders/PipelineTess_RelocRemoveUnusedTcsOutputs.pipe
@@ -1,0 +1,80 @@
+; Test that the TCS output at location 0 is removed, and that the mapping of the TES inputs and TCS outputs match.
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% \
+; RUN:         -enable-relocatable-shader-elf \
+; RUN:         -o %t.elf %gfxip %s -v | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: // LLPC location input/output mapping results (TES shader)
+; SHADERTEST: (TES) Input:  loc = 2, comp = 0 =>  Mapped = 0, 0
+; SHADERTEST-LABEL: // LLPC location input/output mapping results (TCS shader)
+; SHADERTEST: (TCS) Output: loc = 2, comp = 0  =>  Mapped = 0, 0
+; END_SHADERTEST
+
+[Version]
+version = 50
+
+[VsGlsl]
+#version 450
+
+void main()
+{
+    gl_Position = vec4(0.0);
+}
+
+[VsInfo]
+entryPoint = main
+
+[TcsGlsl]
+#version 450
+layout(vertices = 3) out;
+
+layout(location = 0) out vec2 _6[3];
+layout(location = 2) out float _7[3];
+
+void main()
+{
+    gl_out[gl_InvocationID].gl_Position = vec4(0.0);
+    _6[gl_InvocationID] = vec2(0.0);
+    _7[gl_InvocationID] = 0.0;
+}
+
+[TcsInfo]
+entryPoint = main
+
+[TesGlsl]
+#version 450
+layout(triangles) in;
+
+layout(location = 2) in float _7[];
+
+void main()
+{
+    float _30_unrolled[3];
+    for (int i = 0; i < int(3); i++)
+    {
+        _30_unrolled[i] = _7[i];
+    }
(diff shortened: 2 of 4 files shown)


llpc/test/shaderdb/relocatable_shaders/PipelineTess_RelocRemoveUnusedTcsOutputs.pipe
```diff
@@ -0,0 +1,80 @@
+; Test that the TCS output at location 0 is removed, and that the mapping of the TES inputs and TCS outputs match.
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% \
+; RUN:         -enable-relocatable-shader-elf \
+; RUN:         -o %t.elf %gfxip %s -v | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: // LLPC location input/output mapping results (TES shader)
```
kuhar, line 6: This needs `{{^}}` to match the beginning of the line, otherwise this can match the test file itself.

s-perron: The test file is not part of the data being checked.  The "-v" option give the disassembled SPIR-V, which will include the glsl source for the shaders.  However the RUN and SHADERTEST* lines are not part of that.

Also because I used "-LABEL", it would fail if it found a second instance, so I'm not too worried about it unknowingly matching something else.

kuhar: I got somewhat paranoid after wasting a couple hours on debugging a group of test that matched fine but did the wrong thing. For example, this test does not check for `LLPC SUCCESS` so I can imagine a scenario were it did not compile anything and merely printed the input and some warning, even if this does not happen with the code today.

I'd like to have at least one beginning of line marker to be able to tell that the test is checking the actual output, without having to run the test and see what the output is to confirm that.

JaxLinAMD:
retest this please

amdvlk-admin:
### Test summary for commit b9ecc3cfc5ebafc59aacb0012824f7c6222a2cb6

<li>PAL: dev 6cc0a035259663610287d8d99262505e78e01db2</li>
<li>SPVGEN: dev efa3cec20c6ebc66254ff9cca9b3ea124811d7a3</li>
<li>XGL: dev f1bf5feb06f0d3c3e64e960797a35cee3eae4aa4</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev 413996e376673364c2c03e5fef2744753c867171</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/225388)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 31425/56347 (55.8%)</li>
<li>Failed: 0/56347 (0.0%)</li>
<li>Not Supported: 24922/56347 (44.2%)</li>
</ul></details>

<li>Passed: 31413/56347 (55.7%)</li>
<li>Failed: 0/56347 (0.0%)</li>
<li>Not Supported: 24934/56347 (44.3%)</li>
</ul></details>

<li>Passed: 31748/56347 (56.3%)</li>
<li>Failed: 0/56347 (0.0%)</li>
<li>Not Supported: 24599/56347 (43.7%)</li>
</ul></details>

<li>Passed: 34582/56347 (61.4%)</li>
<li>Failed: 0/56347 (0.0%)</li>
<li>Not Supported: 21765/56347 (38.6%)</li>
</ul></details>


</ul></details>

lgc/builder/InOutBuilder.cpp
```diff
@@ -313,7 +313,7 @@ void InOutBuilder::markGenericInputOutputUsage(bool isOutput, unsigned location,
   if (!isOutput || m_shaderStage != ShaderStageGeometry) {
     bool keepAllLocations = false;
     if (getPipelineState()->isUnlinked()) {
-      if (m_shaderStage == ShaderStageVertex && isOutput)
+      if (isOutput && m_pipelineState->getNextShaderStage(m_shaderStage, true) == ShaderStageFragment)
```
xuechen417, line 316: The original condition makes `keepAllLocations` true for VS output with TCS/GS/FS as the next shader stage. Current condition works for TES and VS output with FS as the next shader stage, is it expected?

xuechen417:
> The TES inputs and TCS outputs are not matching up when compiling to unlinked shaders. This is because the compiler is inconsistent on when it will try to pack the inputs and outputs. 
> 
For a tessellation pipeline, we only pack VS outputs and TCS inputs, TES outputs and FS inputs. TCS outputs and TES inputs will keep the original behavior.  The framework of input/output packing is designed for linked shaders, that is unlinked shaders should not go througth the code path of input/output packing.

lgc/builder/InOutBuilder.cpp
s-perron: Yes, that is what I wanted.  The unlinked shaders assume that if any vertex processing stage (VS, TCS, GS, TES) is provided, then no new vertex processing stage will be provided while linking.  If we have both the VS and say GS, then we can change the outputs of the VS to match the inputs of the GS, since we have both of them.  The same for TCS.

Is that correct? Can the VS output be used by the FS when other stages are used?

xuechen417: I think FS recieves the output of last vertex processting stage instead of any vertex processing stage, is it right? @amdrexu

amdrexu: Yes, only the outputs of last vertex processing stage will be used to match the inputs of fragment shader.

s-perron: Cool, so I think this is fine.  I'll resolve this converstaion.

s-perron:
V2: rebase

@xuechen417 @amdrexu I think this is fine for a review.  I have not changed anything because I believe the code already does what @xuechen417 suggested it should do, or I explained why I think it is acceptable as is.  Let me know if there is anything else to do.

xuechen417:
LGTM.

amdrexu (approved):

PR #1538, opened 2021-11-24
[CI] Add nightly coverage build configurations

vettoreldaniele:
This PR adds two code coverage configurations to the nightly CI. A follow-up PR will integrate it with the per-PR CI.

Changes:
1) To keep the current maximum of 8 build configurations, the following configurations have been modified:

    - `+clang, assertions OFF` &rarr;  `+clang+coverage, assertions OFF`
    - `+clang+shadercache, assertions ON`  &rarr; `+clang+shadercache+coverage, assertions ON`.

    This makes sure we run code coverage for both configurations with and without shadercache and with and without assertions.

    The above change will also be made to XGL (https://github.com/GPUOpen-Drivers/xgl/blob/dev/.github/workflows/check-xgl-docker.yml) in a follow-up PR to allow it to pick up the new `+clang+coverage, assertions OFF` configuration that replaces the old `+clang, assertions OFF`.

2) For configurations that specify `+coverage`, the driver is built with Clang/LLVM's [source-based code coverage instrumentation](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) by using the compiler and linker flags `-fprofile-instr-generate -fcoverage-mapping`.

   I've changed how flags are handled in the docker script that generates the CMake command. Flags are now stored in an 
 Bash array, which should make it easier to handle flags with spaces in it (such as a collection of compiler flags).

3) When tests are run under a configuration that specifies `+coverage`, a new script `generate-coverage-report.sh` merges the raw profiles generated by running the tests and generates an HTML report limited to LLPC source files.

4) The report is uploaded to a public Google Cloud Storage bucket (`amdvlk-llpc-github-ci-artifacts-public`), storing it into a directory that uses the configuration string and the commit SHA as an identifier. The bucket is maintained by Google, and is read-only to the general public while write access is only granted to the CI bot and Google maintainers.

### DIFF
--- a/docker/amdvlk.Dockerfile
+++ b/docker/amdvlk.Dockerfile
@@ -46,7 +46,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
     && rm -rf /var/lib/apt/lists/* \
     && python3 -m pip install --no-cache-dir --upgrade pip \
     && python3 -m pip install --no-cache-dir --upgrade cmake \
-    && for tool in clang clang++ clang-tidy llvm-symbolizer lld ld.lld ; do \
+    && for tool in clang clang++ clang-tidy llvm-symbolizer llvm-profdata llvm-cov lld ld.lld ; do \
          update-alternatives --install /usr/bin/"$tool" "$tool" /usr/bin/"$tool"-11 10 ; \
         done \
     && update-alternatives --install /usr/bin/ld ld /usr/bin/ld.gold 10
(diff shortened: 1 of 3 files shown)


amdvlk-admin:

vettoreldaniele:
@kuhar can you please take a look?

docker/amdvlk.Dockerfile
```diff
@@ -62,34 +62,42 @@ RUN wget -P /usr/bin/ https://storage.googleapis.com/git-repo-downloads/repo \
...
+         EXTRA_COMPILER_FLAGS+=("-fprofile-instr-generate=/vulkandriver/profile%4m.profraw" "-fcoverage-mapping"); \
+         EXTRA_LINKER_FLAGS+=("-fprofile-instr-generate=/vulkandriver/profile%4m.profraw" "-fcoverage-mapping"); \
        fi \
     && if echo "$FEATURES" | grep -q "+sanitizers" ; then \
-         EXTRA_FLAGS="$EXTRA_FLAGS -DXGL_USE_SANITIZER=Address;Undefined"; \
-         echo "export ASAN_OPTIONS=detect_leaks=0" >> /vulkandriver/env.sh; \
-         echo "export LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so)" >> /vulkandriver/env.sh; \
+         EXTRA_FLAGS+=("-DXGL_USE_SANITIZER=Address;Undefined"); \
+         echo "export ASAN_OPTIONS=detect_leaks=0" >> env.sh; \
+         echo "export LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so)" >> env.sh; \
```
Flakebi, line 94: I think this writes to `/vulkandriver/builds/ci-build/env.sh` instead of `/vulkandriver/env.sh`, which is still read below and in `llpc.Dockerfile`?

vettoreldaniele: Thanks, fixed.

vettoreldaniele:
V1: fixed `env.sh` being written to the wrong location.

vettoreldaniele:
Making this a draft since we're planning to slightly rework the CI configuration as a whole.

PR #1543, opened 2021-11-26
[CI] Refactor nightly CI configuration matrix and build code.

vettoreldaniele:
Initial changes to nightly CI:
1) Remove explicit assertions dimensions and move it into a `+assertions` feature.
2) Always enable the step that frees up space.
3) Change +sanitizers feature name to +ubsan+asan.
4) Change CMake flag handling code to use Bash arrays. This will make it easier to deal with flags with spaces or different combinations of sanitizers.

Changes to be committed after this PR:
XGL change (https://github.com/GPUOpen-Drivers/xgl/pull/139)
LLPC CI change (https://github.com/GPUOpen-Drivers/llpc/pull/1544)

Example successfull run (on my fork): https://github.com/vettoreldaniele/llpc/actions/runs/1505195415

### DIFF
--- a/.github/workflows/build-amdvlk-docker.yml
+++ b/.github/workflows/build-amdvlk-docker.yml
@@ -7,22 +7,21 @@ on:
 
 jobs:
   build-and-push-amdvlk:
-    name: "Features: ${{ matrix.feature-set }}, assertions: ${{ matrix.assertions }}"
+    name: "Features: ${{ matrix.feature-set }}"
     if: github.repository == 'GPUOpen-Drivers/llpc'
     runs-on: ${{ matrix.host-os }}
     strategy:
       fail-fast: false
       matrix:
         host-os:        ["ubuntu-20.04"]
-        image-template: ["gcr.io/stadia-open-source/amdvlk_%s%s_%s:nightly"]
+        image-template: ["gcr.io/stadia-open-source/amdvlk_%s%s:nightly"]
         branch:         [dev]
         config:         [Release]
-        assertions:     ["OFF", "ON"]
-        feature-set:    ["+gcc", "+clang", "+clang+shadercache", "+clang+sanitizers"]
+        feature-set:    ["+gcc", "+clang", "+clang+shadercache", "+clang+ubsan+asan",
+                         "+gcc+assertions", "+clang+assertions", "+clang+shadercache+assertions", "+clang+ubsan+asan+assertions"]
         generator:      [Ninja]
     steps:
       - name: Free up disk space
-        if: contains(matrix.feature-set, '+sanitizers')
         run: |
           echo 'Before:' && df -h
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/hostedtoolcache/boost /opt/ghc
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

Flakebi (commented):
Looks good to me (I’ll leave it for others to approve)

dstutt (commented):
LGTM too (not that I'm that familiar with this)

kuhar (commented):
The previous way of setting CMake flags by directly appending new ones seemed simpler to me. I guess that separating out sanitizers made it more difficult which made you also clean up the rest of the flag processing?

If so, could we build the other flags directly (in a single pass) and only add an extra array-to-cmake-flag case for assertions?

docker/amdvlk.Dockerfile
```diff
@@ -67,29 +64,49 @@ COPY docker/update-llpc.sh /vulkandriver/
...
+    && if echo "$FEATURES" | grep -q "+ubsan" ; then \
+         SANITIZERS+=("Undefined"); \
+       fi \
+    && if [ ${#SANITIZERS[@]} -ne 0 ]; then  \
+         SANITIZERS_SEPARATED_LIST="${SANITIZERS[@]}"; \
+         SANITIZERS_SEPARATED_LIST="${SANITIZERS_SEPARATED_LIST// /;}"; \
+         EXTRA_FLAGS+=("-DXGL_USE_SANITIZER='${SANITIZERS_SEPARATED_LIST}'"); \ 
        else \
-         EXTRA_FLAGS="$EXTRA_FLAGS -DXGL_BUILD_CACHE_CREATOR=ON"; \
+         EXTRA_FLAGS+=("-DXGL_BUILD_CACHE_CREATOR=ON"); \
```
kuhar, line 97: I think we should always enable cache creator now. In the pass, this was disabled because of running into the disk space quota. But since we added a cleanup step, this stopped being an issue.

vettoreldaniele:
> The previous way of setting CMake flags by directly appending new ones seemed simpler to me. I guess that separating out sanitizers made it more difficult which made you also clean up the rest of the flag processing?
> 
> If so, could we build the other flags directly (in a single pass) and only add an extra array-to-cmake-flag case for assertions?

That system breaks when we add flags such as C and C++ and linker flags that have spaces in them (like `-flag1 -flag2`, not happening now but will happen when we add coverage). And sanitizers as well as you pointed out. Considering this, I think this is already in the most simplified state as only compiler,linker and sanitizer flags have two passes (which are required).

vettoreldaniele:
V1: always enable cache creator.

docker/amdvlk.Dockerfile
```diff
@@ -67,29 +64,47 @@ COPY docker/update-llpc.sh /vulkandriver/
...
+         SANITIZERS+=("Undefined"); \
+       fi \
+    && if [ ${#SANITIZERS[@]} -ne 0 ]; then  \
+         SANITIZERS_SEPARATED_LIST="${SANITIZERS[@]}"; \
+         SANITIZERS_SEPARATED_LIST="${SANITIZERS_SEPARATED_LIST// /;}"; \
+         EXTRA_FLAGS+=("-DXGL_USE_SANITIZER='${SANITIZERS_SEPARATED_LIST}'"); \
+       fi \
+    && ASSERTIONS="OFF" \
+    && if echo "$FEATURES" | grep -q "+assertions" ; then \
+         ASSERTIONS="ON"; \
```
kuhar, line 99: Can we update extra flags directly to enable assertions?

kuhar (commented):
> That system breaks when we add flags such as C and C++ and linker flags that have spaces in them (like -flag1 -flag2, not happening now but will happen when we add coverage). And sanitizers as well as you pointed out. Considering this, I think this is already in the most simplified state as only compiler,linker and sanitizer flags have two passes (which are required).

OK, makes sense to me. Thanks for the extra context.

vettoreldaniele:
V2: enable assertions through EXTRA_FLAGS directly.

kuhar (approved):
LGTM, thanks for keeping this infrastructure maintainable.

.github/workflows/build-amdvlk-docker.yml
```diff
@@ -7,22 +7,21 @@ on:
...
         branch:         [dev]
         config:         [Release]
-        assertions:     ["OFF", "ON"]
-        feature-set:    ["+gcc", "+clang", "+clang+shadercache", "+clang+sanitizers"]
+        feature-set:    ["+gcc", "+clang", "+clang+shadercache", "+clang+ubsan+asan",
+                         "+gcc+assertions", "+clang+assertions", "+clang+shadercache+assertions", "+clang+ubsan+asan+assertions"]
         generator:      [Ninja]
     steps:
       - name: Free up disk space
-        if: contains(matrix.feature-set, '+sanitizers')
```
kuhar, line 25: I noticed one more thing: this should be only required when we have any of the sanitizers enabled.

kuhar: It's not a big  deal because the cleanup step takes a couple of minutes, compared to O(hours) of the whole workflow. But we probably want not to always clean up disk space in presubmit workflows.

vettoreldaniele: It'll also be required when we add coverage. During my test runs it never went over the space limit but it came close so I will enable it there as well just for peace of mind.

If we don't want this to always be triggered, I can change this back to a series of ORs (any of the sanitizers for now, + coverage when I add it in the future). Does that sounds good?

kuhar: Ah, good point about coverage. I'd leave it as-is then (at least in the nightly workflow).

vettoreldaniele: SG. Curiously, the step of freeing up space on my fork consistently takes <= 5 seconds.

amdrexu (approved):

PR #1544, opened 2021-11-26
[CI] Reflect nightly CI changes in per-PR CI configuration.

vettoreldaniele:
Reflect changes made in https://github.com/GPUOpen-Drivers/llpc/pull/1543 to apply the same modifications for the per-PR CI.

### DIFF
--- a/.github/workflows/check-amdllpc-docker.yml
+++ b/.github/workflows/check-amdllpc-docker.yml
@@ -9,19 +9,19 @@ on:
 
 jobs:
   build-and-test:
-    name: "Features: ${{ matrix.feature-set }}, assertions: ${{ matrix.assertions }}"
+    name: "Features: ${{ matrix.feature-set }}"
     runs-on: ${{ matrix.host-os }}
     strategy:
       fail-fast: false
       matrix:
         host-os:             ["ubuntu-20.04"]
-        base-image-template: ["gcr.io/stadia-open-source/amdvlk_%s%s_%s:nightly"]
+        image-template:      ["gcr.io/stadia-open-source/amdvlk_%s%s:nightly"]
         config:              [Release]
-        assertions:          ["OFF", "ON"]
-        feature-set:         ["+gcc", "+clang", "+clang+shadercache", "+clang+sanitizers"]
+        feature-set:         ["+gcc", "+clang", "+clang+shadercache", "+clang+ubsan+asan",
+                              "+gcc+assertions", "+clang+assertions", "+clang+shadercache+assertions", "+clang+ubsan+asan+assertions"]
     steps:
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

Flakebi (approved):

kuhar (commented):
LGTM. Once #1543 lands, this needs to be force-pushed to confirm the CI works.

kuhar (changes requested):
The clang tidy check also needs to be updated: https://github.com/GPUOpen-Drivers/llpc/blob/dev/.github/workflows/check-clang-tidy-llpc.yml

vettoreldaniele:
V1: also apply changes to the clang tidy check workflow.

kuhar (approved):
LGTM. Once #1543 lands, this needs to be force-pushed to confirm the CI works.

vettoreldaniele:
@kuhar regarding freeing up space in this workflow, the following options are available:
1) Always free up space, but this step can take up to 2min.
2) Only free up space for sanitizers and coverage. For this, since there's no regex support for `contains`, we can use either:
    * `if: contains(matrix.feature-set, 'san') || contains(matrix.feature-set, '+coverage')`. This is short but may be confusing.
    * `if: contains(matrix.feature-set, '+ubsan') || if: contains(matrix.feature-set, '+asan') || if: contains(matrix.feature-set, '+tsan') || contains(matrix.feature-set, '+coverage')`. This is very verbose.
    
I personally would go with 2), short-hand version with a comment explaining that `san` matches any sanitizer. What do you think?

kuhar:
> @kuhar regarding freeing up space in this workflow, the following options are available:
> 
> 1. Always free up space, but this step can take up to 2min.
> 2. Only free up space for sanitizers and coverage. For this, since there's no regex support for `contains`, we can use either:
>    
>    * `if: contains(matrix.feature-set, 'san') || contains(matrix.feature-set, '+coverage')`. This is short but may be confusing.
>    * `if: contains(matrix.feature-set, '+ubsan') || if: contains(matrix.feature-set, '+asan') || if: contains(matrix.feature-set, '+tsan') || contains(matrix.feature-set, '+coverage')`. This is very verbose.
> 
> I personally would go with 2), short-hand version with a comment explaining that `san` matches any sanitizer. What do you think?

@vettoreldaniele options 2) sounds good for me. This is the only logic in this workflow file so I don't think being explicit hurts. This code is unlikely to grow in complexity because of it.

kuhar (dismissed):
LGTM. Make sure to mark this as ready to review -- we can't merge draft PRs.

trenouf:
This can't possibly make the Jenkins run fail as that does not use the .github directory. :-)

PR #1546, opened 2021-11-29
Extends to multiple BinaryDatas

jiaolu:
Future new pipelines beside graphics/compute may return multiple binary datas

### DIFF
--- a/llpc/tool/llpcPipelineBuilder.cpp
+++ b/llpc/tool/llpcPipelineBuilder.cpp
@@ -84,19 +84,24 @@ bool PipelineBuilder::isGraphicsPipeline() const {
 //
 // @returns : Result::Success on success, other status on failure.
 Result PipelineBuilder::build() {
-  BinaryData pipeline = {};
+  llvm::SmallVector<BinaryData, 1> pipelines;
+  pipelines.push_back(BinaryData());
+  pipelines[0] = {};
   const bool isGraphics = isGraphicsPipeline();
   if (isGraphics) {
-    Result result = buildGraphicsPipeline(pipeline);
+    Result result = buildGraphicsPipeline(pipelines[0]);
     if (result != Result::Success)
       return result;
   } else {
-    Result result = buildComputePipeline(pipeline);
+    Result result = buildComputePipeline(pipelines[0]);
     if (result != Result::Success)
       return result;
   }
 
-  return decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
+  for (const auto &pipeline : pipelines) {
+    decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
+  }
+  return Result::Success;
 }
 
 // =====================================================================================================================
--- a/llpc/tool/llpcPipelineBuilder.cpp
+++ b/llpc/tool/llpcPipelineBuilder.cpp
@@ -84,9 +84,9 @@ bool PipelineBuilder::isGraphicsPipeline() const {
 //
 // @returns : Result::Success on success, other status on failure.
 Result PipelineBuilder::build() {
-  llvm::SmallVector<BinaryData, 1> pipelines;
-  pipelines.push_back(BinaryData());
-  pipelines[0] = {};
+  // Use vector of BinaryData for future pipeline type
+  SmallVector<BinaryData, 1> pipelines;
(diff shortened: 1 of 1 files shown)


amdvlk-admin:

llpc/tool/llpcPipelineBuilder.cpp
```diff
@@ -84,19 +84,24 @@ bool PipelineBuilder::isGraphicsPipeline() const {
 //
 // @returns : Result::Success on success, other status on failure.
 Result PipelineBuilder::build() {
-  BinaryData pipeline = {};
+  llvm::SmallVector<BinaryData, 1> pipelines;
+  pipelines.push_back(BinaryData());
+  pipelines[0] = {};
```
kuhar, line 89: The assignment redundant, the first push back will already initialize the element. I'd simplify this to:
```
pipelines.push_back({});
```

llpc/tool/llpcPipelineBuilder.cpp
```diff
@@ -84,19 +84,24 @@ bool PipelineBuilder::isGraphicsPipeline() const {
 //
 // @returns : Result::Success on success, other status on failure.
 Result PipelineBuilder::build() {
-  BinaryData pipeline = {};
+  llvm::SmallVector<BinaryData, 1> pipelines;
```
kuhar, line 87: Could you add a comment explaining why we need more than one BinaryData? This is not clear from just the code.

llpc/tool/llpcPipelineBuilder.cpp
```diff
@@ -84,19 +84,24 @@ bool PipelineBuilder::isGraphicsPipeline() const {
...
   } else {
-    Result result = buildComputePipeline(pipeline);
+    Result result = buildComputePipeline(pipelines[0]);
     if (result != Result::Success)
       return result;
   }
 
-  return decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
+  for (const auto &pipeline : pipelines) {
+    decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
```
kuhar, line 102: This discards the returned result. It should be something like:
```
for (const auto &pipeline : pipelines) {
    Result result = decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
    if (result != Result::Success)
      return result;
}
return Result::Success;
```

kuhar (changes requested):

llpc/tool/llpcPipelineBuilder.cpp
kuhar: nit: no need to spell out the llvm namespace (`llvm::`) in `.cpp` files

llpc/tool/llpcPipelineBuilder.cpp
```diff
@@ -84,19 +84,26 @@ bool PipelineBuilder::isGraphicsPipeline() const {
 //
 // @returns : Result::Success on success, other status on failure.
 Result PipelineBuilder::build() {
-  BinaryData pipeline = {};
+  // Use vector of BinaryData for future pipeline type
```
kuhar, line 87: 1. nit: this sentence should end in a `.`
2. I'd be more explicit and say something like:
```c++
`// Even though `build*Pipeline` produces a single `BinaryData` today, future pipeline types (e.g., raytracing) will result multiple outputs. We use a vector instead of a single variable to prepare for that.
```

llpc/tool/llpcPipelineBuilder.cpp
```diff
@@ -84,19 +84,26 @@ bool PipelineBuilder::isGraphicsPipeline() const {
...
   } else {
-    Result result = buildComputePipeline(pipeline);
+    Result result = buildComputePipeline(pipelines[0]);
     if (result != Result::Success)
       return result;
   }
 
-  return decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
+  for (const auto &pipeline : pipelines) {
+    auto result = decodePipelineBinary(&pipeline, &m_compileInfo, isGraphics);
```
kuhar, line 102: nit: could you change this to `Result result` to keep consistent with the surrounding code?

kuhar (approved):
LGTM. Thanks for the fixes.

amdrexu (approved):

PR #1554, opened 2021-12-01
Refine the per-vertex input variable

AMD-dwang:
Add OpCode::ReadPerVertexInput and separate per-vertex input variable from
Opcode::ReadGenericInput, when Replaying we can handle it alone.
Subdivision primitive type, For different type, we have different order
for per-vertex input.


amdvlk-admin:

PR #1555, opened 2021-12-01
Refine the per-vertex input variable

AMD-dwang:
Add OpCode::ReadPerVertexInput and separate per-vertex input variable from
Opcode::ReadGenericInput, when Replaying we can handle it alone.
Subdivision primitive type, For different type, we have different order
for per-vertex input.

### DIFF
--- a/lgc/builder/BuilderRecorder.cpp
+++ b/lgc/builder/BuilderRecorder.cpp
@@ -160,6 +160,8 @@ StringRef BuilderRecorder::getCallName(Opcode opcode) {
     return "buffer.ptrdiff";
   case Opcode::ReadGenericInput:
     return "read.generic.input";
+  case Opcode::ReadPerVertexInput:
+    return "read.pervertex.input";
   case Opcode::ReadGenericOutput:
     return "read.generic.output";
   case Opcode::WriteGenericOutput:
@@ -1384,6 +1386,35 @@ Value *BuilderRecorder::CreateReadGenericInput(Type *resultTy, unsigned location
 }
 
 // =====================================================================================================================
+// Create a read of (part of) a perVertex input value, passed from the previous shader stage.
+//
+// @param resultTy : Type of value to read
+// @param location : Base location (row) of input
+// @param locationOffset : Variable location offset; must be within locationCount
+// @param elemIdx : Element index in vector. (This is the SPIR-V "component", except that it is half the component for
+// 64-bit elements.)
+// @param locationCount : Count of locations taken by the input
+// @param inputInfo : Extra input info (FS interp info)
+// @param vertexIndex : Vertex index, for each vertex, it is unique. the range is 0-2, up to three vertices per
+// primitive; for FS custom interpolated input: auxiliary interpolation value;
+// @param instName : Name to give instruction(s)
+// @returns Value of input
+Value *BuilderRecorder::CreateReadPerVertexInput(Type *resultTy, unsigned location, Value *locationOffset,
+                                                 Value *elemIdx, unsigned locationCount, InOutInfo inputInfo,
+                                                 Value *vertexIndex, const Twine &instName) {
+  return record(Opcode::ReadPerVertexInput, resultTy,
+                {
+                    getInt32(location),
+                    locationOffset,
+                    elemIdx,
+                    getInt32(locationCount),
+                    getInt32(inputInfo.getData()),
+                    vertexIndex,
+                },
+                instName);
+}
+
+// =====================================================================================================================
 // Create a read of (part of) a user output value, the last written value in the same shader stage.
 //
 // @param resultTy : Type of value to read
@@ -2079,6 +2110,7 @@ Instruction *BuilderRecorder::record(BuilderRecorder::Opcode opcode, Type *resul
     case Opcode::ReadBuiltInOutput:
     case Opcode::ReadGenericInput:
     case Opcode::ReadGenericOutput:
+    case Opcode::ReadPerVertexInput:
       // Functions that only read memory.
       func->addFnAttr(Attribute::ReadOnly);
       // Must be marked as returning for DCE.
--- a/lgc/builder/BuilderReplayer.cpp
+++ b/lgc/builder/BuilderReplayer.cpp
@@ -613,6 +613,17 @@ Value *BuilderReplayer::processCall(unsigned opcode, CallInst *call) {
                                              isa<UndefValue>(args[5]) ? nullptr : &*args[5]); // Vertex index
   }
 
+  case BuilderRecorder::Opcode::ReadPerVertexInput: {
+    InOutInfo inputInfo(cast<ConstantInt>(args[4])->getZExtValue());
+    return m_builder->CreateReadPerVertexInput(call->getType(),                                // Result type
+                                               cast<ConstantInt>(args[0])->getZExtValue(),     // Location
+                                               args[1],                                        // Location offset
+                                               isa<UndefValue>(args[2]) ? nullptr : &*args[2], // Element index
+                                               cast<ConstantInt>(args[3])->getZExtValue(),     // Location count
+                                               inputInfo,                                      // Input info
+                                               args[5]);                                       // Vertex index
+  }
+
   case BuilderRecorder::Opcode::ReadGenericOutput: {
     InOutInfo outputInfo(cast<ConstantInt>(args[4])->getZExtValue());
     return m_builder->CreateReadGenericOutput(call->getType(),                                 // Result type
--- a/lgc/builder/InOutBuilder.cpp
+++ b/lgc/builder/InOutBuilder.cpp
@@ -66,6 +66,150 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
 }
 
 // =====================================================================================================================
+// Create a read of (part of) a perVertex input value, passed from the previous shader stage.
+// The result type is as specified by pResultTy, a scalar or vector type with no more than four elements.
+// A "location" contains four "components", each of which can contain a 16- or 32-bit scalar value. A
+// 64-bit scalar value takes two components.
+//
+// @param resultTy : Type of value to read
+// @param location : Base location (row) of input
+// @param locationOffset : Variable location offset; must be within locationCount
+// @param elemIdx : Element index in vector. (This is the SPIR-V "component", except that it is half the component for
+// 64-bit elements.)
+// @param locationCount : Count of locations taken by the input
+// @param inputInfo : Extra input info (FS interp info)
+// @param vertexIndex : Vertex index, for each vertex, it is unique. the range is 0-2, up to three vertices per
+// primitive; for FS custom interpolated input: auxiliary interpolation value;
+// @param instName : Name to give instruction(s)
+// @returns Value of input
+Value *InOutBuilder::CreateReadPerVertexInput(Type *resultTy, unsigned location, Value *locationOffset, Value *elemIdx,
+                                              unsigned locationCount, InOutInfo inputInfo, Value *vertexIndex,
+                                              const Twine &instName) {
+  assert(!resultTy->isAggregateType());
+  assert(inputInfo.getInterpMode() == InOutInfo::InterpModeCustom);
+  assert(m_shaderStage == ShaderStageFragment);
+
+  // Fold constant pLocationOffset into location.
+  if (auto constLocOffset = dyn_cast<ConstantInt>(locationOffset)) {
+    location += constLocOffset->getZExtValue();
+    locationOffset = getInt32(0);
+    locationCount = divideCeil(resultTy->getPrimitiveSizeInBits(), 128);
+  }
+
+  // Mark the usage of the input/output.
+  markGenericInputOutputUsage(false, location, locationCount, inputInfo, vertexIndex);
+
+  // Fixing the order
+  bool parity = false;
+  Value *odd = nullptr;
+  Value *even = nullptr;
+
+  auto primType = getPipelineState()->getPrimitiveType();
+  switch (primType) {
+  case PrimitiveType::Point:
+    break;
+  case PrimitiveType::Line_List:
+  case PrimitiveType::Line_Strip:
+    break;
+  case PrimitiveType::Triangle_List: {
+    switch (cast<ConstantInt>(vertexIndex)->getZExtValue()) {
+    case 0: {
+      vertexIndex = getInt32(2);
+      break;
+    }
+    case 1: {
+      vertexIndex = getInt32(0);
+      break;
+    }
+    case 2: {
+      vertexIndex = getInt32(1);
+      break;
+    }
+    }
+    break;
+  }
+  case PrimitiveType::Triangle_Strip:
+  case PrimitiveType::Triangle_Fan:
+  case PrimitiveType::Triangle_Strip_Adjacency: {
+    switch (cast<ConstantInt>(vertexIndex)->getZExtValue()) {
+    case 0: {
+      odd = getInt32(2);
+      even = vertexIndex;
+      break;
+    }
+    case 1: {
+      odd = getInt32(0);
+      even = vertexIndex;
+      break;
+    }
+    case 2: {
+      odd = getInt32(1);
+      even = vertexIndex;
+      break;
+    }
+    }
+    parity = true;
+    break;
+  }
+  case PrimitiveType::Triangle_List_Adjacency: {
+    switch (cast<ConstantInt>(vertexIndex)->getZExtValue()) {
+    case 0: {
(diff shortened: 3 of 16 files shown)


amdvlk-admin:

lgc/builder/BuilderReplayer.cpp
```diff
@@ -613,6 +613,17 @@ Value *BuilderReplayer::processCall(unsigned opcode, CallInst *call) {
...
 
+  case BuilderRecorder::Opcode::ReadPerVertexInput: {
+    InOutInfo inputInfo(cast<ConstantInt>(args[4])->getZExtValue());
+    return m_builder->CreateReadPerVertexInput(call->getType(),                                 // Result type
+                                               cast<ConstantInt>(args[0])->getZExtValue(),      // Location
+                                               args[1],                                         // Location offset
+                                               isa<UndefValue>(args[2]) ? nullptr : &*args[2],  // Element index
+                                               cast<ConstantInt>(args[3])->getZExtValue(),      // Location count
+                                               inputInfo,                                       // Input info
+                                               isa<UndefValue>(args[5]) ? nullptr : &*args[5]); // Vertex index
```
kuhar, line 624: Could you extract the parameters into local variables? This will make it easier to debug, as we will be able to dump values without having to retype everything. And we won't need the comments.

lgc/builder/InOutBuilder.cpp
```diff
@@ -65,6 +65,161 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
...
+// @param resultTy : Type of value to read
+// @param location : Base location (row) of input
+// @param locationOffset : Variable location offset; must be within locationCount
+// @param elemIdx : Element index in vector. (This is the SPIR-V "component", except that it is half the component for
+// 64-bit elements.)
+// @param locationCount : Count of locations taken by the input
+// @param inputInfo : Extra input info (FS interp info)
+// @param vertexIndex : Vertex index; for FS custom interpolated input: auxiliary interpolation value;
+// @param instName : Name to give instruction(s)
+Value *InOutBuilder::CreateReadPerVertexInput(Type *resultTy, unsigned location, Value *locationOffset, Value *elemIdx,
```
kuhar, line 83: `@returns` is missing

lgc/builder/InOutBuilder.cpp
```diff
@@ -65,6 +65,161 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
...
+                                              unsigned locationCount, InOutInfo inputInfo, Value *vertexIndex,
+                                              const Twine &instName) {
+  assert(resultTy->isAggregateType() == false);
```
kuhar, line 86: nit: `!resultTy->isAggregateType()`

lgc/state/PipelineState.cpp
```diff
@@ -1531,18 +1531,58 @@ unsigned PipelineState::getVerticesPerPrimitive() {
...
+    }
+  } else if (hasShaderStage(ShaderStageTessControl) || hasShaderStage(ShaderStageTessEval)) {
+    assert(getInputAssemblyState().primitiveType == PrimitiveType::Patch);
+    const auto &tessMode = getShaderModes()->getTessellationMode();
+    if (tessMode.pointMode)
+      return PrimitiveType::Point;
+    else if (tessMode.primitiveMode == PrimitiveMode::Isolines)
+      return PrimitiveType::Line_Strip;
+    else if (tessMode.primitiveMode == PrimitiveMode::Triangles)
+      return PrimitiveType::Triangle_Strip;
```
kuhar, line 1578: nit: no need for `else` after `return`

kuhar (commented):
This seems like a big functional change but there are no new tests. Would it be possible to add LIT tests for this, is it sufficiently covered by existing lit tests, or do you rely on some CTS tests?

AMD-dwang:
> This seems like a big functional change but there are no new tests. Would it be possible to add LIT tests for this, is it sufficiently covered by existing lit tests, or do you rely on some CTS tests?

Yes, there are no tests for this. This is a new feature, and can't be tested now.

lgc/builder/BuilderRecorder.cpp
```diff
@@ -1383,6 +1385,32 @@ Value *BuilderRecorder::CreateReadGenericInput(Type *resultTy, unsigned location
...
+//
+// @param resultTy : Type of value to read
+// @param location : Base location (row) of input
+// @param locationOffset : Variable location offset; must be within locationCount
+// @param elemIdx : Vector index
+// @param locationCount : Count of locations taken by the input
+// @param inputInfo : Extra input info (FS interp info)
+// @param vertexIndex : Vertex index
+// @param instName : Name to give instruction(s)
+Value *BuilderRecorder::CreateReadPerVertexInput(Type *resultTy, unsigned location, Value *locationOffset,
```
kuhar, line 1399: `@returns` is missing

lgc/builder/BuilderRecorder.cpp
```diff
@@ -1383,6 +1385,32 @@ Value *BuilderRecorder::CreateReadGenericInput(Type *resultTy, unsigned location
...
+// =====================================================================================================================
+// Create a read of (part of) a perVertex input value, passed from the previous shader stage.
+//
+// @param resultTy : Type of value to read
+// @param location : Base location (row) of input
+// @param locationOffset : Variable location offset; must be within locationCount
+// @param elemIdx : Vector index
+// @param locationCount : Count of locations taken by the input
+// @param inputInfo : Extra input info (FS interp info)
+// @param vertexIndex : Vertex index
```
kuhar, line 1397: 'Vertex index' is used as the description for both `elemIdx` and `vertexIndex`. What's the difference between the two?

lgc/interface/lgc/Builder.h
```diff
@@ -1092,6 +1088,29 @@ class Builder : public BuilderCommon {
...
+  // @param resultTy : Type of value to read
+  // @param location : Base location (row) of input
+  // @param locationOffset : Location offset; must be within locationCount if variable
+  // @param elemIdx : Element index in vector. (This is the SPIR-V "component", except that it is half the component for
+  // 64-bit elements.)
+  // @param locationCount : Count of locations taken by the input. Ignored if pLocationOffset is const
+  // @param inputInfo : Extra input info (FS interp info)
+  // @param vertexIndex : Vertex index (For FS custom interpolated input: auxiliary interpolation value)
+  // @param instName : Name to give instruction(s)
+  virtual llvm::Value *CreateReadPerVertexInput(llvm::Type *resultTy, unsigned location, llvm::Value *locationOffset,
```
kuhar, line 1110: `@returns` is missing

lgc/builder/BuilderRecorder.cpp
AMD-dwang: "VertexIndex" is for vertex index, For each vertex, it is unique. the range is 0~2, up to three vertices per primitive.
"elemIdx" is for vector component index, the range is 0~3.

linqun (commented):
looks good to me for others.

lgc/builder/InOutBuilder.cpp
```diff
@@ -65,6 +65,162 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
...
+    // Odd
+    SmallVector<Value *, 6> args;
+    args.push_back(getInt32(location));
+    args.push_back(locationOffset);
+    args.push_back(elemIdx);
+    args.push_back(getInt32(inputInfo.getInterpMode()));
+    args.push_back(odd);
+    std::string callName(lgcName::InputImportInterpolant);
+    addTypeMangling(resultTy, args, callName);
+    auto oddRes = emitCall(callName, resultTy, args, {Attribute::ReadOnly, Attribute::WillReturn}, &*GetInsertPoint());
```
Flakebi, line 190: Do you think it makes sense to move this code into a lambda instead of duplicating it?
I.e.
```c++

auto readInput = [&](Value *vertexIndex) {
  std::string callName = lgcName::InputImportInterpolant;
  SmallVector<Value *, 5> args({
    getInt32(location),
    locationOffset,
    elemIdx,
    getInt32(inputInfo.getInterpMode()),
    vertexIndex,
  });
  addTypeMangling(resultTy, args, callName);
  return emitCall(callName, resultTy, args, {Attribute::ReadOnly, Attribute::WillReturn}, &*GetInsertPoint());
}

if (parity) {
  auto oddRes = readInput(odd);
  auto evenRes = readInput(even);
  …
} else {
  result = readInput(vertexIndex);
}
```

lgc/patch/PatchInOutImportExport.cpp
```diff
@@ -2615,6 +2617,48 @@ Value *PatchInOutImportExport::patchFsBuiltInInputImport(Type *inputTy, unsigned
...
       break;
     }
+    case PrimitiveType::Triangle_Strip:
+    case PrimitiveType::Triangle_Fan:
+    case PrimitiveType::Triangle_Strip_Adjacency: {
+      // Odd
+      auto kCoord = builder.CreateFSub(ConstantFP::get(Type::getFloatTy(*m_context), 1.0), iCoord);
+      kCoord = builder.CreateFSub(kCoord, jCoord);
+      Value *odd = UndefValue::get(inputTy);
+      odd = builder.CreateInsertElement(odd, iCoord, uint64_t(2));
```
Flakebi, line 2627: I think the `uint64_t` C-style cast is unneeded here and below.

lgc/patch/PatchInOutImportExport.cpp
```diff
@@ -2615,6 +2617,48 @@ Value *PatchInOutImportExport::patchFsBuiltInInputImport(Type *inputTy, unsigned
...
+
+      auto primitiveId = patchFsBuiltInInputImport(builder.getInt32Ty(), BuiltInPrimitiveId, nullptr, insertPos);
+      auto evenV = builder.CreateSRem(primitiveId, builder.getInt32(2));
+      Value *con = builder.CreateICmpEQ(evenV, builder.getInt32(0));
+      input = builder.CreateSelect(con, even, odd);
+      break;
+    }
+    case PrimitiveType::Triangle_List_Adjacency: {
+      // Odd
+      auto kCoord = builder.CreateFSub(ConstantFP::get(Type::getFloatTy(*m_context), 1.0), iCoord);
```
Flakebi, line 2644: `Type::getFloatTy(*m_context)` could be `builder.getFloatTy()` to be consistent with the `getInt32Ty`s.

lgc/builder/InOutBuilder.cpp
```diff
@@ -65,6 +65,162 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
...
+      vertexIndex = getInt32(1);
+      break;
+    }
+    }
+    break;
+  }
+  case PrimitiveType::Triangle_Strip:
+  case PrimitiveType::Triangle_Fan:
+  case PrimitiveType::Triangle_Strip_Adjacency: {
+    switch (dyn_cast<ConstantInt>(vertexIndex)->getZExtValue()) {
```
Flakebi, line 133: This can just be a `cast`. If the `dyn_cast` returned `null`, it would crash anyway.
(The same goes for the cast above and below.)

Flakebi:
Thanks for the helpful comments in the code.

I think a lit test for this change would be nice to prevent future regressions.

lgc/builder/BuilderRecorder.cpp
kuhar: Thanks for the explanation. Could you add it as a code comment?

lgc/patch/PatchInOutImportExport.cpp
AMD-dwang: It will report an error "ambiguous call to overloaded function" without C-style cast.
PS:
it is right like this:
  input = builder.CreateInsertElement(input, iCoord, uint64_t(0));
  input = builder.CreateInsertElement(input, jCoord, 1);
  input = builder.CreateInsertElement(input, kCoord, 2);

but it is wrong like this:
  input = builder.CreateInsertElement(input, iCoord, uint64_t(2));
  input = builder.CreateInsertElement(input, jCoord, 0);
  input = builder.CreateInsertElement(input, kCoord, 1);

lgc/builder/InOutBuilder.cpp
AMD-dwang: Yes, it is clearer. Thanks.

lgc/patch/PatchInOutImportExport.cpp
Flakebi: Oh, I see. `0` is ambiguous between an `uint64_t` and a `Value*` because it can be interpreted as `null`. Thanks for checking.

(Seems like numbers other than 0 are not ambiguous and the ambiguity of 0 can also be solved by writing `0ul`, but that’s better done as a separate change.)

AMD-dwang:
> Thanks for the helpful comments in the code.
> 
> I think a lit test for this change would be nice to prevent future regressions.

There are no tests for this. This is a new feature, and can't be tested now.

lgc/state/PipelineState.cpp
```diff
@@ -1531,18 +1531,58 @@ unsigned PipelineState::getVerticesPerPrimitive() {
...
+      llvm_unreachable("Unexpected output primitive type!");
+    }
+  } else if (hasShaderStage(ShaderStageTessControl) || hasShaderStage(ShaderStageTessEval)) {
+    assert(getInputAssemblyState().primitiveType == PrimitiveType::Patch);
+    const auto &tessMode = getShaderModes()->getTessellationMode();
+    if (tessMode.pointMode)
+      return PrimitiveType::Point;
+    if (tessMode.primitiveMode == PrimitiveMode::Isolines)
+      return PrimitiveType::Line_Strip;
+    if (tessMode.primitiveMode == PrimitiveMode::Triangles)
```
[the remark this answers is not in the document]
AMD-dwang: I will add it. It should return Triangle_Strip

lgc/builder/InOutBuilder.cpp
```diff
@@ -65,6 +65,162 @@ Value *InOutBuilder::CreateReadGenericInput(Type *resultTy, unsigned location, V
...
+    args.push_back(getInt32(location));
+    args.push_back(locationOffset);
+    args.push_back(elemIdx);
+    args.push_back(getInt32(inputInfo.getInterpMode()));
+    args.push_back(even);
+    callName = std::string(lgcName::InputImportInterpolant);
+    addTypeMangling(resultTy, args, callName);
+    auto evenRes = emitCall(callName, resultTy, args, {Attribute::ReadOnly, Attribute::WillReturn}, &*GetInsertPoint());
+
+    auto primtiveId = readBuiltIn(false, BuiltInPrimitiveId, {}, nullptr, nullptr, "");
```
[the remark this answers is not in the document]
AMD-dwang: This question is interesting, but i am not sure what will happen. currently there seems to be no way to check this. It needs to be tested. Thanks.

lgc/state/PipelineState.cpp
```diff
@@ -1531,18 +1531,58 @@ unsigned PipelineState::getVerticesPerPrimitive() {
...
+    }
   }
 
   llvm_unreachable("Unable to get vertices per primitive!");
   return 0;
 }
 
+// =====================================================================================================================
+// Get the primitive type. For GS, the type is for output primitive.
+PrimitiveType PipelineState::getPrimitiveType() {
```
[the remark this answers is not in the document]
AMD-dwang: There is no difference between wireframe and fill.

kuhar (commented):
Thanks for the fixes, @AMD-dwang.

> There are no tests for this. This is a new feature, and can't be tested now.

Can you open an issue on GitHub reminding us to add tests in the future? You could even reference it in the code, e.g.: `TODO(https://github.com/GPUOpen-Drivers/llpc/issues/9999): Add tests once <some feature> is done.`

amdrexu (approved):

amdrexu:
@AMD-dwang Please check the clang-tidy issues. I think they might be historical problems, not relevant to your changes. Also, please check internal Jenkins results. If you believe it is OK to merge, please ping me. Also, file a issue ticket of adding LIT tests of this change once the feature is released. Thanks.

AMD-dwang:
> @AMD-dwang Please check the clang-tidy issues. I think they might be historical problems, not relevant to your changes. Also, please check internal Jenkins results. If you believe it is OK to merge, please ping me. Also, file a issue ticket of adding LIT tests of this change once the feature is released. Thanks.

@amdrexu the clang tidy issues is irrelevant to my changes. Internal Jenkins test are passed, I think you can help to merge it. I created a issue of adding LIT tests: https://github.com/GPUOpen-Drivers/llpc/issues/1576. Thanks.

PR #1573, opened 2021-12-07
Workaround "ImageQueryLod"

AMD-dwang:
When UV width equals 0, the result will return 0, but we expect the value is
-22.0 to -infinity. we perform derivative operations on the XY of UV coordinate
 respectively, if both of result are zero, we will return -FLT_MAX for LOD.


amdvlk-admin:

kuhar (commented):
Can we have a LIT test for this?

AMD-dwang:
> 

This workaround is only related to the input data, and has nothing to do with the shader itself.
 only when we use same texture coordinates (fwidth(UV) == 0) we can find the difference. 
and cts tests will be added. Thanks.

kuhar:
@AMD-dwang 
> This workaround is only related to the input data, and has nothing to do with the shader itself.
only when we use same texture coordinates (fwidth(UV) == 0) we can find the difference.
and cts tests will be added. Thanks.

Sorry, I don't follow. This pass takes SPIR-V as input and outputs LLVM IR. Why can't we have a LIT test that makes sure that the code that comes out of the translator has the expected LLVM instructions?

With CTS the problem is that you need a GPU to execute the compiled code, which we don't have on generic CI bots and most development workstations. There's also no clear connection between the code and CTS tests. This makes it difficult to tell if the code is doing what is was designed to do or jus happens to work, or if our in-development changes cause breakage.

AMD-dwang:
> @AMD-dwang
> 
> > This workaround is only related to the input data, and has nothing to do with the shader itself.
> > only when we use same texture coordinates (fwidth(UV) == 0) we can find the difference.
> > and cts tests will be added. Thanks.
> 
> Sorry, I don't follow. This pass takes SPIR-V as input and outputs LLVM IR. Why can't we have a LIT test that makes sure that the code that comes out of the translator has the expected LLVM instructions?
> 
> With CTS the problem is that you need a GPU to execute the compiled code, which we don't have on generic CI bots and most development workstations. There's also no clear connection between the code and CTS tests. This makes it difficult to tell if the code is doing what is was designed to do or jus happens to work, or if our in-development changes cause breakage.

Okay, I will add a LIT test, thanks.

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6336,6 +6336,32 @@ Value *SPIRVToLLVM::transSPIRVImageQueryLodFromInst(SPIRVInstruction *bi, BasicB
     result = ConvertingSamplerSelectLadderHelper(result, convertingSamplerIdx, createImageGetLod);
   }
 
+  // NOTE: This is a workaround. When UV width equals 0, the result return 0, but we expect the value is
+  // -22.0 to -infinity.
+  // Derivative operations are performed on the XY of UV coordinate respectively, if both of result
+  // are zero, we will return -FLT_MAX for LOD.
+  Value *dpdx = getBuilder()->CreateDerivative(coord, false, false);
+  Value *dpdy = getBuilder()->CreateDerivative(coord, true, false);
+  Value *absDpdx = getBuilder()->CreateUnaryIntrinsic(Intrinsic::fabs, dpdx);
```
amdrexu, line 6345: Let's try to reduce some local variables
Value *absDpdx = getBuilder()->CreateUnaryIntrinsic(Intrinsic::fabs, getBuilder()->CreateDerivative(coord, false, false));
Value *absDpdy = ...
Remove dpdx and dpdy

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6336,6 +6336,32 @@ Value *SPIRVToLLVM::transSPIRVImageQueryLodFromInst(SPIRVInstruction *bi, BasicB
...
+  Value *absDpdy = getBuilder()->CreateUnaryIntrinsic(Intrinsic::fabs, dpdy);
+  Value *add = getBuilder()->CreateFAdd(absDpdx, absDpdy);
```
amdrexu, line 6347: add -> absDdpxPlusDpdy

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6336,6 +6336,32 @@ Value *SPIRVToLLVM::transSPIRVImageQueryLodFromInst(SPIRVInstruction *bi, BasicB
...
+  Value *cmp = getBuilder()->CreateFCmpOEQ(add, ConstantFP::get(coord->getType(), 0.0));
```
amdrexu, line 6348: cmp -> maybeZero

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6336,6 +6336,32 @@ Value *SPIRVToLLVM::transSPIRVImageQueryLodFromInst(SPIRVInstruction *bi, BasicB
...
+
+  Value *con = getBuilder()->getInt1(true);
```
amdrexu, line 6350: con -> isZero
getInt1(true) -> getTrue()

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -6336,6 +6336,32 @@ Value *SPIRVToLLVM::transSPIRVImageQueryLodFromInst(SPIRVInstruction *bi, BasicB
...
+  if (cmp->getType()->isVectorTy()) {
+    auto vec = cast<FixedVectorType>(cmp->getType());
```
amdrexu, line 6352: Could merge these two statements
for (unsigned elemIdx = 0; elemIdx < cast<FixedVectorType>(maybeZero->getType())->getNumElements(); elemIdx++) {

AMD-dwang: Fixed.

amdrexu (approved):

PR #1582, opened 2021-12-10
Fix the non-root pushconst

AMD-dwang:
When a "PushConst" is in a non-root table, we will not get pushConst
args, the tests crash.


amdvlk-admin:

PR #1587, opened 2021-12-13
Add resource layout scheme

AMD-dwang:
Indirect scheme is designed for the case that the pipeline layout only
contains part of information of the final executable pipeline. So that
user data is used in conservative way to make sure that this pipeline
layout is alaways compatiable with other layouts which may contain the
descriptor which is not known currently.

In indirect mode, the user data registers for various resources are
allocated in the following fashion:
 1. one user data entry for vertex buffer
 2. one user data entry for the push constant buffer
 3. descriptor set index for each set.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -47,7 +47,7 @@
 #define LLPC_INTERFACE_MAJOR_VERSION 53
 
 /// LLPC minor interface version.
-#define LLPC_INTERFACE_MINOR_VERSION 1
+#define LLPC_INTERFACE_MINOR_VERSION 2
 
 #ifndef LLPC_CLIENT_INTERFACE_MAJOR_VERSION
 #if VFX_INSIDE_SPVGEN
@@ -86,6 +86,7 @@
 //  %Version History
 //  | %Version | Change Description                                                                                    |
 //  | -------- | ----------------------------------------------------------------------------------------------------- |
+//  |     53.2 | Add resourceLayoutScheme to PipelineOptions                                                           |
 //  |     53.1 | Add PartPipelineStage enum for part-pipeline mode                                                     |
 //  |     53.0 | Add optimizationLevel to PipelineOptions                                                              |
 //  |     52.3 | Add fastMathFlags to PipelineShaderOptions                                                            |
@@ -393,6 +394,16 @@ enum class ThreadGroupSwizzleMode : unsigned {
   Count
 };
 
+/// Represents mapping layout of the resources used in shaders
+enum class ResourceLayoutScheme : unsigned {
+  Compact = 0, ///< Compact scheme make full use of all the user data registers.
+  Indirect     ///< Fixed layout, push constant will be the sub node of DescriptorTableVaPtr
+               ///  In indirect scheme, The order of resources is like this:
+               ///  1. one user data entry for vertex buffer
+               ///  2. one user data entry for the push constant buffer
+               ///  3. descriptor set index for each set
+};
+
 /// Represents per pipeline options.
 struct PipelineOptions {
   bool includeDisassembly;         ///< If set, the disassembly for all compiled shaders will be included in
@@ -420,6 +431,7 @@ struct PipelineOptions {
   uint32_t optimizationLevel; ///< The higher the number the more optimizations will be performed.  Valid values are
                               ///< between 0 and 3.
 #endif
+  ResourceLayoutScheme resourceLayoutScheme; ///< Resource layout scheme
 };
 
 /// Prototype of allocator for output data buffer, used in shader-specific operations.
--- a/lgc/builder/DescBuilder.cpp
+++ b/lgc/builder/DescBuilder.cpp
@@ -221,10 +221,31 @@ Value *DescBuilder::CreateGetDescPtr(ResourceNodeType descType, unsigned descSet
 // @param pushConstantsTy : Type of the push constants table that the returned pointer will point to
 // @param instName : Name to give instruction(s)
 Value *DescBuilder::CreateLoadPushConstantsPtr(Type *pushConstantsTy, const Twine &instName) {
+  Type *returnTy = pushConstantsTy->getPointerTo(ADDR_SPACE_CONST);
+  const bool isIndirect = getPipelineState()->getOptions().resourceLayoutScheme == ResourceLayoutScheme::Indirect;
+  if (isIndirect) {
+    // Push const is the sub node of DescriptorTableVaPtr.
+    if (m_pipelineState->getUserDataNodes().empty()) {
+      Value *highHalf = getInt32(HighAddrPc);
+      Value *descPtr = CreateNamedCall(
+          lgcName::DescriptorTableAddr, getInt8Ty()->getPointerTo(ADDR_SPACE_CONST),
+          {getInt32(unsigned(ResourceNodeType::PushConst)), getInt32(-1), getInt32(0), highHalf}, Attribute::ReadNone);
+      return CreateBitCast(descPtr, returnTy);
+    }
+
+    const ResourceNode *topNode = m_pipelineState->findPushConstantResourceNode();
+    assert(topNode);
+    const ResourceNode subNode = topNode->innerTable[0];
+    Value *highHalf = getInt32(HighAddrPc);
+    Value *descPtr = CreateNamedCall(
+        lgcName::DescriptorTableAddr, getInt8Ty()->getPointerTo(ADDR_SPACE_CONST),
+        {getInt32(unsigned(ResourceNodeType::PushConst)), getInt32(subNode.set), getInt32(subNode.binding), highHalf},
+        Attribute::ReadNone);
+    return CreateBitCast(descPtr, returnTy);
+  }
   // Get the push const pointer. If subsequent code only uses this with constant GEPs and loads,
   // then PatchEntryPointMutate might be able to "unspill" it so the code uses shader entry SGPRs
   // directly instead of loading from the spill table.
-  Type *returnTy = pushConstantsTy->getPointerTo(ADDR_SPACE_CONST);
   std::string callName = lgcName::PushConst;
   addTypeMangling(returnTy, {}, callName);
   return CreateNamedCall(callName, returnTy, {}, Attribute::ReadOnly, instName);
--- a/lgc/interface/lgc/CommonDefs.h
+++ b/lgc/interface/lgc/CommonDefs.h
@@ -89,6 +89,12 @@ enum class ResourceNodeType : unsigned {
   DescriptorReserved16,
   Count, ///< Count of resource mapping node types.
 };
+
+// Represents mapping layout of the resources used in shaders
+enum class ResourceLayoutScheme : unsigned {
+  Compact = 0, ///< Compact scheme make full use of all the user data registers.
+  Indirect     ///< Fixed layout, push constant will be the sub node of DescriptorTableVaPtr
+};
 } // namespace lgc
 
 namespace llvm {
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -360,10 +360,20 @@ void PatchEntryPointMutate::gatherUserDataUsage(Module *module) {
         auto &descriptorTable = getUserDataUsage(stage)->descriptorTables;
 
         if (m_pipelineState->isUnlinked() && m_pipelineState->getUserDataNodes().empty()) {
-          // The user data nodes are not available, so we use the set as the
-          // index.
-          descriptorTable.resize(std::max(descriptorTable.size(), size_t(set + 1)));
-          descriptorTable[set].users.push_back(call);
+          if (m_pipelineState->getOptions().resourceLayoutScheme == ResourceLayoutScheme::Indirect) {
+            // If the type is pushconst, the index is set 0, others are set + 1.
+            if (resType == ResourceNodeType::PushConst) {
+              descriptorTable.resize(std::max(descriptorTable.size(), size_t(1)));
+              descriptorTable[0].users.push_back(call);
+            } else {
+              descriptorTable.resize(std::max(descriptorTable.size(), size_t(set + 2)));
+              descriptorTable[set + 1].users.push_back(call);
+            }
+          } else {
+            // The user data nodes are not available, so we use the set as the index.
+            descriptorTable.resize(std::max(descriptorTable.size(), size_t(set + 1)));
+            descriptorTable[set].users.push_back(call);
+          }
         } else {
           // The user data nodes are available, so we use the offset of the node as the
           // index.
--- a/llpc/context/llpcPipelineContext.cpp
+++ b/llpc/context/llpcPipelineContext.cpp
@@ -334,6 +334,7 @@ void PipelineContext::setOptionsInPipeline(Pipeline *pipeline, Util::MetroHash64
   options.disableImageResourceCheck = getPipelineOptions()->disableImageResourceCheck;
   options.enableInterpModePatch = getPipelineOptions()->enableInterpModePatch;
   options.pageMigrationEnabled = getPipelineOptions()->pageMigrationEnabled;
+  options.resourceLayoutScheme = static_cast<lgc::ResourceLayoutScheme>(getPipelineOptions()->resourceLayoutScheme);
 
   // Driver report full subgroup lanes for compute shader, here we just set fullSubgroups as default options
   options.fullSubgroups = true;
--- a/llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
+++ b/llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
@@ -1,0 +1,97 @@
+; Check the resource layout mode. In "Indirect" mode, "PushConst" resource is not in root node.
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} SPIR-V lowering results
+; SHADERTEST: [[Desc:%[0-9]*]] = call [16 x i8] addrspace(4)* (...) @lgc.create.load.push.constants.ptr.p4a16i8()
+; SHADERTEST: [[Addr:%[0-9]*]] = bitcast [16 x i8] addrspace(4)* [[Desc]] to <4 x float> addrspace(4)*
+; SHADERTEST: [[Value:%[0-9]*]] = load <4 x float>, <4 x float> addrspace(4)* [[Addr]], align 16
+; SHADERTEST: call void (...) @lgc.create.write.generic.output(<4 x float> [[Value]], i32 0, i32 0, i32 0, i32 0, i32 0, i32 undef)
+
+; SHADERTEST-LABEL: {{^// LLPC}} pipeline before-patching results
+; SHADERTEST: [[Desc:%[0-9]*]] = call i8 addrspace(4)* @lgc.descriptor.table.addr(i32 9, i32 -1, i32 0, i32 -1)
+; SHADERTEST: [[Addr1:%[0-9]*]] = bitcast i8 addrspace(4)* [[Desc]] to [16 x i8] addrspace(4)*
+; SHADERTEST: [[Addr2:%[0-9]*]] = bitcast [16 x i8] addrspace(4)* [[Addr1]] to <4 x float> addrspace(4)*
+; SHADERTEST: [[Value:%[0-9]*]] = load <4 x float>, <4 x float> addrspace(4)* [[Addr2]], align 16
+; SHADERTEST: call void @lgc.output.export.generic.i32.i32.v4f32(i32 0, i32 0, <4 x float> [[Value]])
+
+; SHADERTEST-LABEL: {{^// LLPC}} pipeline patching results
+; SHADERTEST: define dllexport amdgpu_ps void @_amdgpu_ps_main(i32 inreg %{{[^,]*}}, i32 inreg %{{[^,]*}}, i32 inreg %descTable1,
+; SHADERTEST: [[Addr0:%[0-9]*]] = zext i32 %descTable1 to i64
+; SHADERTEST: [[Addr1:%[0-9]*]] = or i64 %{{[0-9]*}}, [[Addr0]]
+; SHADERTEST: [[Addr2:%[0-9]*]] = inttoptr i64 [[Addr1]] to <4 x float> addrspace(4)*
+; SHADERTEST: [[Value:%[0-9]*]] = load <4 x float>, <4 x float> addrspace(4)* [[Addr2]], align 16
+; SHADERTEST: [[Value3:%[.a-zA-Z0-9]+]] = extractelement <4 x float> [[Value]], i64 3
+; SHADERTEST: [[Value2:%[.a-zA-Z0-9]+]] = extractelement <4 x float> [[Value]], i64 2
+; SHADERTEST: [[Value1:%[.a-zA-Z0-9]+]] = extractelement <4 x float> [[Value]], i64 1
+; SHADERTEST: [[Value0:%[.a-zA-Z0-9]+]] = extractelement <4 x float> [[Value]], i64 0
+; SHADERTEST: [[Color0:%[0-9]*]] = call <2 x half> @llvm.amdgcn.cvt.pkrtz(float [[Value0]], float [[Value1]])
+; SHADERTEST: [[Color1:%[0-9]*]] = call <2 x half> @llvm.amdgcn.cvt.pkrtz(float [[Value2]], float [[Value3]])
+; SHADERTEST: call void @llvm.amdgcn.exp.compr.v2f16(i32 immarg 0, i32 immarg 15, <2 x half> [[Color0]], <2 x half> [[Color1]], i1 immarg true, i1 immarg true)
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
+
+
+
+[Version]
+version = 52
+
+[VsGlsl]
+#version 450
+
+layout( location = 0 ) in vec4 app_position;
+
+void main() {
+  gl_Position = app_position;
+}
+
+[VsInfo]
+entryPoint = main
+
+[FsGlsl]
+#version 450
+
+layout( location = 0 ) out vec4 frag_color;
+
+layout( push_constant ) uniform ColorBlock {
+  vec4 Color;
+} PushConstant;
+
+void main() {
+   frag_color = PushConstant.Color;
+}
+
+[FsInfo]
+entryPoint = main
+
+[ResourceMapping]
+userDataNode[0].visibility = 2
+userDataNode[0].type = IndirectUserDataVaPtr
+userDataNode[0].offsetInDwords = 0
+userDataNode[0].sizeInDwords = 1
+userDataNode[0].indirectUserDataCount = 4
+userDataNode[1].visibility = 66
+userDataNode[1].type = DescriptorTableVaPtr
+userDataNode[1].offsetInDwords = 1
+userDataNode[1].sizeInDwords = 1
+userDataNode[1].next[0].type = PushConst
+userDataNode[1].next[0].offsetInDwords = 0
+userDataNode[1].next[0].sizeInDwords = 4
+userDataNode[1].next[0].set = 0xFFFFFFFF
+userDataNode[1].next[0].binding = 0
+
+[GraphicsPipelineState]
+colorBuffer[0].format = VK_FORMAT_R8G8B8A8_UNORM
+colorBuffer[0].channelWriteMask = 15
+colorBuffer[0].blendEnable = 0
+colorBuffer[0].blendSrcAlphaToColor = 0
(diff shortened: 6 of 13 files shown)


amdvlk-admin:

AMD-dwang:
Can some one take a look?

include/vkgcDefs.h
```diff
@@ -378,6 +379,14 @@ enum class ThreadGroupSwizzleMode : unsigned {
   Count
 };
 
+/// Represents mapping layout of the resources used in shaders
+enum class ResourceMappingLayout : unsigned {
+  // Compact scheme make full use of all the user data registers.
+  Compact = 0,
+  // The searching path of resource belongs to a specific binding is fixed in indirect scheme.
```
kuhar, line 386: nit: The search path?

include/vkgcDefs.h
```diff
@@ -378,6 +379,14 @@ enum class ThreadGroupSwizzleMode : unsigned {
   Count
 };
 
+/// Represents mapping layout of the resources used in shaders
+enum class ResourceMappingLayout : unsigned {
+  // Compact scheme make full use of all the user data registers.
```
kuhar, line 384: nit: can you put these comments on the right of values, just like in the lgc enum?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1190,6 +1190,62 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
     return;
   }
 
+  // Define the function
```
kuhar, line 1193: This comment is not very helpful. It'd be better to comment what this lambda is used for, why do we need it, etc.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1190,6 +1190,62 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
...
+  auto addPushConstArgs = [&](const ResourceNode &node) {
```
kuhar, line 1194: This lambda is very long. How about making a freestanding or a member function?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1190,6 +1190,62 @@ void PatchEntryPointMutate::addUserDataArgs(SmallVectorImpl<UserDataArg> &userDa
...
+          continue;
+        }
+
+        // Add the arg (part of the push const) that we can potentially unspill.
+        addUserDataArg(userDataArgs, node.offsetInDwords + dwordOffset, pushConstOffset.dwordSize,
+                       "pushConst" + Twine(dwordOffset), &pushConstOffset.entryArgIdx, builder);
+      }
+    } else {
+      // Mark push constant for spill for compute library.
+      userDataUsage->pushConstSpill = true;
```
kuhar, line 1241: nit: Can you flip the if condition and put this `else` case first? The current `then` case is very long and it's not immediately obvious which `if` this `else` belongs to.

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
```diff
@@ -0,0 +1,171 @@
+
+
```
kuhar, line 2: Can you describe what the test checks?

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
```diff
@@ -0,0 +1,171 @@
...
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} SPIRV-to-LLVM translation results
+; SHADERTEST: AMDLLPC SUCCESS
```
kuhar, line 6: This does not match any compilation results (LLVM IR, GCN, Pal metadata, etc.). How can we tell if the code does what its supposed to?

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
```diff
@@ -0,0 +1,171 @@
...
+binding[0].stride = 32
+binding[0].inputRate = VK_VERTEX_INPUT_RATE_VERTEX
+attribute[0].location = 0
+attribute[0].binding = 0
+attribute[0].format = VK_FORMAT_R32G32B32A32_SFLOAT
+attribute[0].offset = 0
+attribute[1].location = 1
+attribute[1].binding = 0
+attribute[1].format = VK_FORMAT_R32G32B32A32_SFLOAT
+attribute[1].offset = 16
```
kuhar, line 171: missing newline

kuhar (commented):
I only commented on some cosmetics. Steven would be a better person to check the logic.

include/vkgcDefs.h
AMD-dwang: Yes, I will change it . Thanks.

include/vkgcDefs.h
AMD-dwang: This comment is copied from driver, but it is really hard to understand.

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
AMD-dwang: It will crash when compiling this pipeline without this change.  This pipeline is aim to verify this change, not to special instructions. It is okay if it can be compiled successfully. Thanks.

kuhar: Can you add a comment with this explanation?

AMD-dwang: I added it. Thanks.

lgc/patch/PatchEntryPointMutate.cpp
AMD-dwang: I make it a member function. Thanks.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1325,6 +1281,69 @@ void PatchEntryPointMutate::addUserDataArg(SmallVectorImpl<UserDataArg> &userDat
...
 }
 
+// =====================================================================================================================
+// Add a PushConst UserDataArg
+//
+// @param userDataArgs : Vector to add UserDataArg to
+// @param node : PushConst resource node
+// @param builder : IRBuilder (just for getting types)
+void PatchEntryPointMutate::addPushConstArgs(SmallVectorImpl<UserDataArg> &userDataArgs, const ResourceNode &node,
+                                             llvm::IRBuilder<> &builder) {
```
s-perron, line 1291: It looks a little suspicious to me that you can just use the same code here regardless of whether the node is in the root table or a descriptor table.  I expect that you will need to different function that will share some similar code.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1325,6 +1281,69 @@ void PatchEntryPointMutate::addUserDataArg(SmallVectorImpl<UserDataArg> &userDat
...
+
+      // Add the arg (part of the push const) that we can potentially unspill.
+      addUserDataArg(userDataArgs, node.offsetInDwords + dwordOffset, pushConstOffset.dwordSize,
+                     "pushConst" + Twine(dwordOffset), &pushConstOffset.entryArgIdx, builder);
+    }
+  }
+
+  // Ensure we mark the push constant's part of the spill table as used.
+  if (userDataUsage->pushConstSpill)
+    userDataUsage->spillUsage = std::min(userDataUsage->spillUsage, node.offsetInDwords);
```
s-perron, line 1344: Is this right?  Should we be spilling the root table even if `node` is not in the root table?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -1325,6 +1281,69 @@ void PatchEntryPointMutate::addUserDataArg(SmallVectorImpl<UserDataArg> &userDat
...
+        }
+      }
+      if (haveOverlap) {
+        userDataUsage->pushConstSpill = true;
+        continue;
+      }
+
+      // Add the arg (part of the push const) that we can potentially unspill.
+      addUserDataArg(userDataArgs, node.offsetInDwords + dwordOffset, pushConstOffset.dwordSize,
+                     "pushConst" + Twine(dwordOffset), &pushConstOffset.entryArgIdx, builder);
```
s-perron, line 1338: I would expect that in the indirect case, these values will always have to be unspilled.  Can you point out the code that will do the unspilling?

s-perron (changes requested):
I have concerns about the correctness in all cases.  If I saw tests that called out what instructions are expected and what the PAL metadata is suppose to be, I could be convinced that this is correct.

Flakebi:
To make sure that the implementation actually works, you could also add an amber testcase like the one for a16 (https://github.com/GPUOpen-Drivers/llpc/blob/dev/test/amber/a16.amber).
It may be a bit hidden in the a16 example, but they support lit CHECK lines to check for lines in the pipeline dumps.

(These tests are not yet run in CI unfortunately, but you can run them locally with `ninja check-amber` in an xgl build.)

lgc/patch/PatchEntryPointMutate.cpp
s-perron: Sorry, I think I meant that it should never be unspilled.  Are we sure that will not happen?

lgc/interface/lgc/CommonDefs.h
```diff
@@ -89,6 +89,12 @@ enum class ResourceNodeType : unsigned {
   DescriptorReserved16,
   Count, ///< Count of resource mapping node types.
 };
+
+// Represents mapping layout of the resources used in shaders
+enum class ResourceMappingLayout : unsigned {
+  Compact = 0, /// Compact scheme make full use of all the user data registers.
+  Indirect     /// The searching path of resource belongs to a specific binding is fixed in indirect scheme.
```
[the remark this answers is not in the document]
AMD-dwang: The same enumeration was added in driver, which changes the resource mapping layout. I think it's best to keep their names and comments consistent. Also this enumeration is not only for push Constant, the whole resource mapping is affected. Thanks.

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
```diff
@@ -0,0 +1,171 @@
...
+options.enableLoadScalarizer = 0
+options.disableLicm = 0
+options.unrollThreshold = 0
+options.scalarThreshold = 0
+options.disableLoopUnroll = 0
+options.fp32DenormalMode = Auto
+options.adjustDepthImportVrs = 0
+options.disableLicmThreshold = 0
+options.unrollHintThreshold = 0
+options.dontUnrollHintThreshold = 0
```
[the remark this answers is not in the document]
AMD-dwang: Thanks, I will remove unnecessary options.

lgc/patch/PatchEntryPointMutate.cpp
AMD-dwang: I changed "node.offsetInDwords" to "tableNode.offsetInDwords". Thanks.

lgc/patch/PatchEntryPointMutate.cpp
AMD-dwang: The code is the same as before. I only move it. I think it should be correct, after all it has been stable. Thanks.

lgc/patch/PatchEntryPointMutate.cpp
AMD-dwang: > To make sure that the implementation actually works, you could also add an amber testcase like the one for a16 (https://github.com/GPUOpen-Drivers/llpc/blob/dev/test/amber/a16.amber). It may be a bit hidden in the a16 example, but they support lit CHECK lines to check for lines in the pipeline dumps.
> 
> (These tests are not yet run in CI unfortunately, but you can run them locally with `ninja check-amber` in an xgl build.)

Thanks, I try to add it.

AMD-dwang:
The change of driver has been submitted, an amber test is not be added.

AMD-dwang:
@s-perron I have addressed the comments. this change only adds a resource mapping mode,  no other feature is changed. Please help to take a look again. Thank you.

kuhar:
> @s-perron I have addressed the comments. this change only adds a resource mapping mode, no other feature is changed. Please help to take a look again. Thank you.

@AMD-dwang Steven is on vacation and will be back in January.

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
kuhar: I don't see the checks that Steven suggested in the test, so I'm marking this as not resolved yet.

kuhar (changes requested):
There's one outstanding change request from Steven's previous review: https://github.com/GPUOpen-Drivers/llpc/pull/1587/files#r769648173

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
AMD-dwang: Currently, are there such three test cases with compact mode?

lgc/interface/lgc/CommonDefs.h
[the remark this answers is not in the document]
s-perron: At the very least, could up update the comment above, even if you can't update the name.  If I am reading code that checks if the layout is indirect, I have no way of knowing what that means because the comment references itself: what is search path for the?  The comment for indirect does not read well.  I'm guessing "belongs" is a typo and should be "belonging".

Can we ask the driver team to update the comment in the driver code?  Then you can keep them the same.

lgc/patch/PatchEntryPointMutate.cpp
s-perron: Even if this is the same code that does not mean it is correct because you are executing it in a different context.  For example, should you be setting `userDataUsage->spillUsage` below if the push constant node is not in the root table?

I still wonder what will happen with the pushConst argument in `PatchEntryPointMutate::determineUnspilledUserDataArgs`.  Again, that code was written assuming that push constants are in the root table.  What will that code do if it decides pass the push constant in SGPRs instead of loading it from memory?  I assume that should never happen, but no code was added to make sure it does not.

llpc/test/shaderdb/PipelineVsFs_TestNonRootPushConst.pipe
s-perron: I don't know if they currently exist or not.  I believe the test that you have here would normally not spill the push constant, but you can look into that.  If it does spill you can probably change the type of mvp to float, and it will unspill it.  I don't know if there is a case for the compute shaders, but it should be easy enough to write.  I wrote a test for relocatable shaders that checks that it works when the push const is spilled: 

https://github.com/GPUOpen-Drivers/llpc/blob/b780577b54bcccb85c191d022b63ea5ce9cc72a4/llpc/test/shaderdb/relocatable_shaders/PipelineCs_PushConst.pipe.

lgc/interface/lgc/CommonDefs.h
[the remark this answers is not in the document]
AMD-dwang: "The fixed user data registers. When the pipeline layout only contains part of layout information of the final executable pipeline, we don't know the offset of user data register, so we must use fixed offset." 

Is it feasible to change the comment like above?

amdvlk-admin:
### Test summary for commit 5c010407a433f6aea2f06f5ec1a3086ddcf160bd

</ul></details>

</ul></details>

</ul></details>


</ul></details>

s-perron (dismissed):
The tests needs to be expanded.  I'd also like some better documentation on the "compact" and "indirect" schemes.  We should also have an separate discussion on why we need this.  I don't see the reason for this change in the discussion.  I'll send an email about it.

tool/dumper/vkgcPipelineDumper.cpp
```diff
@@ -684,6 +685,7 @@ void PipelineDumper::dumpComputeStateInfo(const ComputePipelineBuildInfo *pipeli
 void PipelineDumper::dumpPipelineOptions(const PipelineOptions *options, std::ostream &dumpFile) {
   dumpFile << "options.includeDisassembly = " << options->includeDisassembly << "\n";
   dumpFile << "options.scalarBlockLayout = " << options->scalarBlockLayout << "\n";
+  dumpFile << "options.resourceLayoutScheme = " << options->resourceLayoutScheme << "\n";
```
s-perron, line 688: I think this file should also include the resource layout scheme in the hash.

lgc/builder/DescBuilder.cpp
```diff
@@ -221,13 +221,27 @@ Value *DescBuilder::CreateGetDescPtr(ResourceNodeType descType, unsigned descSet
...
+  bool isIndirect = getPipelineState()->getOptions().resourceLayoutScheme == ResourceLayoutScheme::Indirect;
+  if (isIndirect) {
+    // Push const is the sub node of DescriptorTableVaPtr.
+    const ResourceNode *topNode = m_pipelineState->findPushConstantResourceNode();
+    assert(topNode);
+    const ResourceNode subNode = topNode->innerTable[0];
+    Value *highHalf = getInt32(HighAddrPc);
+    Value *descPtr = CreateNamedCall(
+        lgcName::DescriptorTableAddr, getInt8Ty()->getPointerTo(ADDR_SPACE_CONST),
+        {getInt32(unsigned(ResourceNodeType::PushConst)), getInt32(subNode.set), getInt32(subNode.binding), highHalf},
```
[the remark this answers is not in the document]
AMD-dwang: Here set and binding are a fixed value, set = -1,binding  = 0, they are not meaningful values,  but we can find matched resource node according to them.

llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
```diff
@@ -0,0 +1,69 @@
+; Check the resource layout mode. In "Indirect" mode, "PushConst" resource is not in root node.
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} SPIR-V lowering results
+; SHADERTEST: call [16 x i8] addrspace(4)* (...) @lgc.create.load.push.constants.ptr.p4a16i8()
+; SHADERTEST-LABEL: define dllexport spir_func void @lgc.shader.FS.main()
+; SHADERTEST: call i8 addrspace(4)* @lgc.descriptor.table.addr(i32 9, i32 -1, i32 0, i32 -1)
+; SHADERTEST: AMDLLPC SUCCESS
```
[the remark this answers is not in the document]
AMD-dwang: This test is not to verify the use of push constant variable, I just want to show that we will call `@lgc.descriptor.table.addr` to load descriptor in indirect scheme.  The second call is part of the 'before-patching results', I will fix here.
`@lgc.descriptor.table.addr(i32 9, i32 -1, i32 0, i32 -1)` can represent we are using `@lgc.descriptor.table.addr` to load push constant descriptor, other instructions are not important.
Thanks.

llpc/context/llpcPipelineContext.cpp
```diff
@@ -334,6 +334,7 @@ void PipelineContext::setOptionsInPipeline(Pipeline *pipeline, Util::MetroHash64
   options.disableImageResourceCheck = getPipelineOptions()->disableImageResourceCheck;
   options.enableInterpModePatch = getPipelineOptions()->enableInterpModePatch;
   options.pageMigrationEnabled = getPipelineOptions()->pageMigrationEnabled;
+  options.resourceLayoutScheme = static_cast<lgc::ResourceLayoutScheme>(getPipelineOptions()->resourceLayoutScheme);
```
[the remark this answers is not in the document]
AMD-dwang: Yes, I will add an option. Thanks.

tool/dumper/vkgcPipelineDumper.cpp
AMD-dwang: Oh, I think hash includes resource mapping, so I didn't add it.  After checking again, it only is added when relocatable is false. Thanks, I will add it.

include/vkgcDefs.h
```diff
@@ -393,6 +394,12 @@ enum class ThreadGroupSwizzleMode : unsigned {
   Count
 };
 
+/// Represents mapping layout of the resources used in shaders
+enum class ResourceLayoutScheme : unsigned {
+  Compact = 0, ///< Compact scheme make full use of all the user data registers.
+  Indirect     ///< Fixed layout, push constant will be the sub node of DescriptorTableVaPtr
```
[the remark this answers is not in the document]
AMD-dwang: When VK_PIPELINE_LAYOUT_CREATE_INDEPENDENT_SETS_BIT_EXT is used, we will using indirect mode.

AMD-dwang:
Add an option "resource-layout-scheme" to support 'offline' compilation

amdvlk-admin:
### Test summary for commit 20deb2dea6c352cfeeef04903d2e5d758a26b98a
<details><summary>CTS tests (Failed: 0/187823)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36645/65226 (56.2%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 28581/65226 (43.8%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>

<li>Passed: 31111/57371 (54.2%)</li>
<li>Failed: 0/57371 (0.0%)</li>
<li>Not Supported: 26260/57371 (45.8%)</li><li>Warnings: 0/57371 (0.0%)</li>
</ul></details>

<li>Passed: 37804/65226 (58.0%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 27422/65226 (42.0%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit 66ad702cc9a0bb97a55ceafcc8e84d4ff8f6d567
<details><summary>CTS tests (Failed: 0/187823)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36645/65226 (56.2%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 28581/65226 (43.8%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>

<li>Passed: 31111/57371 (54.2%)</li>
<li>Failed: 0/57371 (0.0%)</li>
<li>Not Supported: 26260/57371 (45.8%)</li><li>Warnings: 0/57371 (0.0%)</li>
</ul></details>

<li>Passed: 37804/65226 (58.0%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 27422/65226 (42.0%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>


</ul></details>

include/vkgcDefs.h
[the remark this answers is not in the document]
s-perron: Maybe a full document is not needed, but at least move the comment you have the EntryPointMutate.cpp here.  This will make it easier to find.  If someone wants to find out what the indirect scheme is, they look at the definition of the enum.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -364,6 +364,24 @@ void PatchEntryPointMutate::gatherUserDataUsage(Module *module) {
           // index.
           descriptorTable.resize(std::max(descriptorTable.size(), size_t(set + 1)));
           descriptorTable[set].users.push_back(call);
+
+          if (m_pipelineState->getOptions().resourceLayoutScheme == ResourceLayoutScheme::Indirect) {
+            // In indirect scheme, The order of resources is like this:
+            // 1. one user data entry for vertex buffer
+            // 2. one user data entry for the push constant buffer
+            // 3. descriptor set index for each set
+            // If the type is pushconst, the index is set 0, others are set + 1.
```
s-perron, line 373: This type of comment should be somewhere more discoverable.  If I am reading `DescBuilder.cpp` I might want this information.  Could you place this with the definition of `ResourceLayoutScheme` in include/vkgcDefs.h.

llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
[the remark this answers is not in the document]
s-perron: I still think the test should be checking for more, but ,if other reviewers do not think it is necessary, I won't stand in the way.

llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
```diff
@@ -0,0 +1,69 @@
+; Check the resource layout mode. In "Indirect" mode, "PushConst" resource is not in root node.
+
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
```
kuhar, line 4: I'm a bit confused. I don't see `--resource-layout-scheme` specified here, doesn't  this test default to the compact mode then?

lgc/builder/DescBuilder.cpp
```diff
@@ -221,13 +221,35 @@ Value *DescBuilder::CreateGetDescPtr(ResourceNodeType descType, unsigned descSet
...
 // @param instName : Name to give instruction(s)
 Value *DescBuilder::CreateLoadPushConstantsPtr(Type *pushConstantsTy, const Twine &instName) {
-  // Get the push const pointer. If subsequent code only uses this with constant GEPs and loads,
-  // then PatchEntryPointMutate might be able to "unspill" it so the code uses shader entry SGPRs
-  // directly instead of loading from the spill table.
   Type *returnTy = pushConstantsTy->getPointerTo(ADDR_SPACE_CONST);
-  std::string callName = lgcName::PushConst;
-  addTypeMangling(returnTy, {}, callName);
-  return CreateNamedCall(callName, returnTy, {}, Attribute::ReadOnly, instName);
+  bool isIndirect = getPipelineState()->getOptions().resourceLayoutScheme == ResourceLayoutScheme::Indirect;
```
kuhar, line 225: nit: add `const` if this won't be modified later in the function

lgc/builder/DescBuilder.cpp
```diff
@@ -221,13 +221,35 @@ Value *DescBuilder::CreateGetDescPtr(ResourceNodeType descType, unsigned descSet
...
+    const ResourceNode *topNode = m_pipelineState->findPushConstantResourceNode();
+    assert(topNode);
+    const ResourceNode subNode = topNode->innerTable[0];
+    Value *highHalf = getInt32(HighAddrPc);
+    Value *descPtr = CreateNamedCall(
+        lgcName::DescriptorTableAddr, getInt8Ty()->getPointerTo(ADDR_SPACE_CONST),
+        {getInt32(unsigned(ResourceNodeType::PushConst)), getInt32(subNode.set), getInt32(subNode.binding), highHalf},
+        Attribute::ReadNone);
+    return CreateBitCast(descPtr, returnTy);
+  } else {
```
kuhar, line 245: nit: else after return: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

llpc/test/shaderdb/general/PipelineVsFs_TestIndirectResourceLayout.pipe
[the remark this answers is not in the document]
AMD-dwang: Okay, I will add more instructions. Thanks

amdvlk-admin:
### Test summary for commit f38083652b54d78e7b6e3513fea7917db0fc05da
<details><summary>CTS tests (Failed: 0/186422)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36645/65226 (56.2%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 28581/65226 (43.8%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>

<li>Passed: 30002/55970 (53.6%)</li>
<li>Failed: 0/55970 (0.0%)</li>
<li>Not Supported: 25968/55970 (46.4%)</li><li>Warnings: 0/55970 (0.0%)</li>
</ul></details>

<li>Passed: 37804/65226 (58.0%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 27422/65226 (42.0%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>


</ul></details>

AMD-dwang:
Do someone have any comments for this change? It is blocking some tests. Thanks.

amdrexu:
We have some internal test failures that depend on this fix. If the PR could be reviewed and gets merged timely, that will be fine for us.

s-perron (commented):
LGTM

kuhar (approved):

amdrexu (approved):

PR #1606, opened 2021-12-28
Add unlinked shader stage build function

AMD-dwang:
When shaders are as a library, we sometimes need to quickly link them. In
this case, we must to compile early and store ElfPackage, then link them
to a whole pipeline.

### DIFF
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -54,6 +54,7 @@
 #include "lgc/ElfLinker.h"
 #include "lgc/EnumIterator.h"
 #include "lgc/PassManager.h"
+#include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/BinaryFormat/MsgPackDocument.h"
 #include "llvm/Bitcode/BitcodeReader.h"
@@ -574,6 +575,261 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T> 
 }
 
 // =====================================================================================================================
+// Helper function for dumping compiler options
+//
+// @param pipelineDumpFile : Handle of pipeline dump file
+void Compiler::dumpCompilerOptions(void *pipelineDumpFile) {
+  if (!pipelineDumpFile)
+    return;
+  std::string extraInfo;
+  raw_string_ostream os(extraInfo);
+  os << ";Compiler Options: " << join(m_options, " ");
+  os.flush();
+  PipelineDumper::DumpPipelineExtraInfo(reinterpret_cast<PipelineDumpFile *>(pipelineDumpFile), &extraInfo);
+}
+
+// =====================================================================================================================
+// Build unlinked shader to ElfPackage with pipeline info.
+//
+// @param pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param stage : Shader stage of needing to compile
+// @param pipelineDumpFile : Handle of pipeline dump file; It is optional, the default value is nullptr
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, UnlinkedShaderStage stage,
+                                          void *pipelineDumpFile) {
+  if (!pipelineInfo->pfnOutputAlloc)
+    return Result::ErrorInvalidPointer;
+  // clang-format off
+  SmallVector<const PipelineShaderInfo *, ShaderStageGfxCount> shaderInfo = {
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr
+  };
+  // clang-format on
+
+  switch (stage) {
+  case UnlinkedStageVertexProcess: {
+    shaderInfo[ShaderStageVertex] = &pipelineInfo->vs;
+    shaderInfo[ShaderStageTessControl] = &pipelineInfo->tcs;
+    shaderInfo[ShaderStageTessEval] = &pipelineInfo->tes;
+    shaderInfo[ShaderStageGeometry] = &pipelineInfo->gs;
+    break;
+  }
+  case UnlinkedStageFragment: {
+    shaderInfo[ShaderStageFragment] = &pipelineInfo->fs;
+    break;
+  }
+  default:
+    llvm_unreachable("");
+    break;
+  }
+
+  dumpCompilerOptions(pipelineDumpFile);
+
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, false, true, stage);
+
+  // Compile
+  GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
+  Context *context = acquireContext();
+  context->attachPipelineContext(&graphicsContext);
+
+  ElfPackage candidateElf;
+  Result result =
+      buildUnlinkedShaderInternal(context, shaderInfo, stage, candidateElf, pipelineOut->stageCacheAccesses);
+
+  releaseContext(context);
+
+  if (result != Result::Success)
+    return result;
+
+  void *allocBuf = pipelineInfo->pfnOutputAlloc(pipelineInfo->pInstance, pipelineInfo->pUserData, candidateElf.size());
+  uint8_t *code = static_cast<uint8_t *>(allocBuf);
+  memcpy(code, candidateElf.data(), candidateElf.size());
+  pipelineOut->pipelineBin.codeSize = candidateElf.size();
+  pipelineOut->pipelineBin.pCode = code;
+  return result;
+}
+
+// =====================================================================================================================
+// Build the whole graphics pipeline. If missing elfPackage of a certain stage, we will build it first, and
+// link them to the full pipeline last.
+//
+// @param pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param elfPackage : Early compiled elfPackage; it is an array which size is UnlinkedStageCount.
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *elfPackage) {
+  if (!pipelineInfo->pfnOutputAlloc)
+    return Result::ErrorInvalidPointer;
+
+  // clang-format off
+  SmallVector<const PipelineShaderInfo *, ShaderStageGfxCount> shaderInfo = {
+    &pipelineInfo->vs,
+    &pipelineInfo->tcs,
+    &pipelineInfo->tes,
+    &pipelineInfo->gs,
+    &pipelineInfo->fs,
+  };
+  // clang-format on
+
+  if (!canUseRelocatableGraphicsShaderElf(shaderInfo, pipelineInfo)) {
+    LLPC_OUTS("Relocatable shader compilation requested but not possible.\n");
+    return Result::ErrorInvalidValue;
+  }
+
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, false);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, false, false);
+
+  Optional<CacheAccessor> cacheAccessor;
+  if (cl::CacheFullPipelines) {
+    cacheAccessor.emplace(pipelineInfo, cacheHash, getInternalCaches());
+  }
+
+  Result result = Result::Success;
+  BinaryData elfBin = {};
+  ElfPackage elf[UnlinkedStageCount] = {};
+  ElfPackage pipelineElf;
+  if (!cacheAccessor && cacheAccessor->isInCache()) {
+    LLPC_OUTS("Cache hit for graphics pipeline.\n");
+    elfBin = cacheAccessor->getElfFromCache();
+    pipelineOut->pipelineCacheAccess =
+        cacheAccessor->hitInternalCache() ? CacheAccessInfo::InternalCacheHit : CacheAccessInfo::CacheHit;
+  } else {
+    GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
+    Context *context = acquireContext();
+    context->attachPipelineContext(&graphicsContext);
+
+    // Release context.
+    auto onExit = make_scope_exit([&] { releaseContext(context); });
+
+    for (unsigned idx = 0; idx < UnlinkedStageCount; idx++) {
+      if (elfPackage[idx].pCode) {
+        auto data = reinterpret_cast<const char *>(elfPackage[idx].pCode);
+        elf[idx].assign(data, data + elfPackage[idx].codeSize);
+      } else {
+        graphicsContext.setUnlinked(true);
+        result = buildUnlinkedShaderInternal(context, shaderInfo, static_cast<UnlinkedShaderStage>(idx), elf[idx],
+                                             pipelineOut->stageCacheAccesses);
+        if (result != Result::Success)
+          return result;
+      }
+    }
+
+    // Link
+    graphicsContext.setUnlinked(false);
+    bool hasError = false;
+    context->setDiagnosticHandler(std::make_unique<LlpcDiagnosticHandler>(&hasError));
+    hasError |= !linkRelocatableShaderElf(elf, &pipelineElf, context);
+
+    context->setDiagnosticHandler(nullptr);
+    if (hasError)
+      return Result::ErrorInvalidShader;
+
+    elfBin.codeSize = pipelineElf.size();
+    elfBin.pCode = pipelineElf.data();
+  }
+
+  void *allocBuf = pipelineInfo->pfnOutputAlloc(pipelineInfo->pInstance, pipelineInfo->pUserData, elfBin.codeSize);
+  uint8_t *code = static_cast<uint8_t *>(allocBuf);
+  memcpy(code, elfBin.pCode, elfBin.codeSize);
+  pipelineOut->pipelineBin.codeSize = elfBin.codeSize;
+  pipelineOut->pipelineBin.pCode = code;
+
+  return result;
+}
+
+// =====================================================================================================================
+// Build unlinked shader from the specified info.
+//
+// @param context : Acquired context
+// @param shaderInfo : Shader info of this pipeline
+// @param stage :  Shader stage of needing to compile
+// @param [out] elfPackage : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   CacheAccessInfo::CacheNotChecked
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildUnlinkedShaderInternal(Context *context, ArrayRef<const PipelineShaderInfo *> shaderInfo,
+                                             UnlinkedShaderStage stage, ElfPackage &elfPackage,
+                                             MutableArrayRef<CacheAccessInfo> stageCacheAccesses) {
+  if (!hasDataForUnlinkedShaderType(stage, shaderInfo))
+    return Result::Success;
+
+  unsigned originalShaderStageMask = context->getPipelineContext()->getShaderStageMask();
+  const MetroHash::Hash originalCacheHash = context->getPipelineContext()->getCacheHashCodeWithoutCompact();
+  unsigned shaderStageMask = getShaderStageMaskForType(stage) & originalShaderStageMask;
+  context->getPipelineContext()->setShaderStageMask(shaderStageMask);
+  const auto shaderStages = maskToShaderStages(shaderStageMask);
+  assert(all_of(shaderStages, isNativeStage) && "Unexpected stage kind");
+
+  // Check the cache for the relocatable shader for this stage.
+  MetroHash::Hash cacheHash = {};
+  if (context->isGraphics()) {
+    auto pipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(context->getPipelineBuildInfo());
+    cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
+  } else {
+    auto pipelineInfo = reinterpret_cast<const ComputePipelineBuildInfo *>(context->getPipelineBuildInfo());
+    cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true, true);
+  }
+  // Note that this code updates m_pipelineHash of the pipeline context. It must be restored.
+  context->getPipelineContext()->setHashForCacheLookUp(cacheHash);
+  LLPC_OUTS("Finalized hash for " << getUnlinkedShaderStageName(stage) << " stage cache lookup: "
+                                  << format_hex(context->getPipelineContext()->get128BitCacheHashCode()[0], 18) << ' '
+                                  << format_hex(context->getPipelineContext()->get128BitCacheHashCode()[1], 18)
+                                  << '\n');
+
+  auto onExit = make_scope_exit([&] {
+    context->getPipelineContext()->setShaderStageMask(originalShaderStageMask);
+    context->getPipelineContext()->setHashForCacheLookUp(originalCacheHash);
+  });
+
+  Result result = Result::Success;
+  CacheAccessor cacheAccessor(context, cacheHash, getInternalCaches());
+  if (cacheAccessor.isInCache()) {
+    BinaryData elfBin = cacheAccessor.getElfFromCache();
+    auto data = reinterpret_cast<const char *>(elfBin.pCode);
+    elfPackage.assign(data, data + elfBin.codeSize);
+    LLPC_OUTS("Cache hit for shader stage " << getUnlinkedShaderStageName(stage) << "\n");
+    for (ShaderStage gfxStage : shaderStages)
+      stageCacheAccesses[gfxStage] =
+          cacheAccessor.hitInternalCache() ? CacheAccessInfo::InternalCacheHit : CacheAccessInfo::CacheHit;
+  } else {
+    LLPC_OUTS("Cache miss for shader stage " << getUnlinkedShaderStageName(stage) << "\n");
+    for (ShaderStage gfxStage : shaderStages)
+      stageCacheAccesses[gfxStage] = CacheAccessInfo::CacheMiss;
+
+    // There was a cache miss, so we need to build the relocatable shader for
+    // this stage.
+    const PipelineShaderInfo *singleStageShaderInfo[ShaderStageNativeStageCount] = {nullptr, nullptr, nullptr,
+                                                                                    nullptr, nullptr, nullptr};
+
+    for (ShaderStage nativeStage : shaderStages)
+      singleStageShaderInfo[nativeStage] = shaderInfo[nativeStage];
+
+    result = buildPipelineInternal(context, singleStageShaderInfo, PipelineLink::Unlinked, nullptr, &elfPackage,
+                                   stageCacheAccesses);
+    if (result == Result::Success) {
+      // Add the result to the cache.
+      BinaryData elfBin = {elfPackage.size(), elfPackage.data()};
+      cacheAccessor.setElfInCache(elfBin);
+      LLPC_OUTS("Updating the cache for unlinked shader stage " << getUnlinkedShaderStageName(stage) << "\n");
+    }
+  }
+
+  return result;
+}
+
+// =====================================================================================================================
 // Builds a pipeline by building relocatable elf files and linking them together.  The relocatable elf files will be
 // cached for future use.
 //
@@ -589,7 +845,6 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
   LLPC_OUTS("Building pipeline with relocatable shader elf.\n");
   Result result = Result::Success;
 
-  unsigned originalShaderStageMask = context->getPipelineContext()->getShaderStageMask();
   bool isUnlinkedPipeline = context->getPipelineContext()->isUnlinked();
   context->getPipelineContext()->setUnlinked(true);
 
@@ -604,66 +859,9 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
   for (UnlinkedShaderStage stage : lgc::enumRange<UnlinkedShaderStage>()) {
     if (!hasDataForUnlinkedShaderType(stage, shaderInfo))
       continue;
-
-    unsigned shaderStageMask = getShaderStageMaskForType(stage) & originalShaderStageMask;
-    context->getPipelineContext()->setShaderStageMask(shaderStageMask);
-    const auto shaderStages = maskToShaderStages(shaderStageMask);
-    assert(all_of(shaderStages, isNativeStage) && "Unexpected stage kind");
-
-    // Check the cache for the relocatable shader for this stage .
-    MetroHash::Hash cacheHash = {};
-    if (context->isGraphics()) {
-      auto pipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(context->getPipelineBuildInfo());
-      cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
-    } else {
-      auto pipelineInfo = reinterpret_cast<const ComputePipelineBuildInfo *>(context->getPipelineBuildInfo());
-      cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true, true);
-    }
-    // Note that this code updates m_pipelineHash of the pipeline context. It
-    // must be restored before we link the pipeline ELF at the end of this for-loop.
-    context->getPipelineContext()->setHashForCacheLookUp(cacheHash);
-    LLPC_OUTS("Finalized hash for " << getUnlinkedShaderStageName(stage) << " stage cache lookup: "
-                                    << format_hex(context->getPipelineContext()->get128BitCacheHashCode()[0], 18) << ' '
-                                    << format_hex(context->getPipelineContext()->get128BitCacheHashCode()[1], 18)
-                                    << '\n');
-
-    CacheAccessor cacheAccessor(context, cacheHash, getInternalCaches());
-    if (cacheAccessor.isInCache()) {
-      BinaryData elfBin = cacheAccessor.getElfFromCache();
-      auto data = reinterpret_cast<const char *>(elfBin.pCode);
-      elf[stage].assign(data, data + elfBin.codeSize);
-      LLPC_OUTS("Cache hit for shader stage " << getUnlinkedShaderStageName(stage) << "\n");
-      for (ShaderStage stage : shaderStages)
-        stageCacheAccesses[stage] =
-            cacheAccessor.hitInternalCache() ? CacheAccessInfo::InternalCacheHit : CacheAccessInfo::CacheHit;
-
-      continue;
-    }
-
-    LLPC_OUTS("Cache miss for shader stage " << getUnlinkedShaderStageName(stage) << "\n");
-    for (ShaderStage stage : shaderStages)
-      stageCacheAccesses[stage] = CacheAccessInfo::CacheMiss;
-
-    // There was a cache miss, so we need to build the relocatable shader for
-    // this stage.
-    const PipelineShaderInfo *singleStageShaderInfo[ShaderStageNativeStageCount] = {nullptr, nullptr, nullptr,
-                                                                                    nullptr, nullptr, nullptr};
-    for (ShaderStage stage : shaderStages)
-      singleStageShaderInfo[stage] = shaderInfo[stage];
-
-    Vkgc::ElfPackage &stageElf = elf[stage];
-    result = buildPipelineInternal(context, singleStageShaderInfo, PipelineLink::Unlinked, nullptr, &stageElf,
-                                   stageCacheAccesses);
-    if (result != Result::Success)
+    if (buildUnlinkedShaderInternal(context, shaderInfo, stage, elf[stage], stageCacheAccesses) != Result::Success)
       break;
-
-    // Add the result to the cache.
-    BinaryData elfBin = {stageElf.size(), stageElf.data()};
-    cacheAccessor.setElfInCache(elfBin);
-    LLPC_OUTS("Updating the cache for unlinked shader stage " << getUnlinkedShaderStageName(stage) << "\n");
   }
-  context->getPipelineContext()->setHashForCacheLookUp(originalCacheHash);
-  context->getPipelineContext()->setShaderStageMask(originalShaderStageMask);
   context->getPipelineContext()->setUnlinked(false);
 
   if (result == Result::Success) {
@@ -1454,14 +1652,8 @@ Result Compiler::BuildGraphicsPipeline(const GraphicsPipelineBuildInfo *pipeline
     LLPC_OUTS("\n");
   }
 
-  if (result == Result::Success && pipelineDumpFile) {
-    std::stringstream strStream;
-    strStream << ";Compiler Options: ";
-    for (auto &option : m_options)
-      strStream << option << " ";
-    std::string extraInfo = strStream.str();
-    PipelineDumper::DumpPipelineExtraInfo(reinterpret_cast<PipelineDumpFile *>(pipelineDumpFile), &extraInfo);
-  }
+  if (result == Result::Success)
+    dumpCompilerOptions(pipelineDumpFile);
 
   Optional<CacheAccessor> cacheAccessor;
   if (cl::CacheFullPipelines) {
@@ -1587,14 +1779,8 @@ Result Compiler::BuildComputePipeline(const ComputePipelineBuildInfo *pipelineIn
     LLPC_OUTS("\n");
   }
 
-  if (result == Result::Success && pipelineDumpFile) {
-    std::stringstream strStream;
-    strStream << ";Compiler Options: ";
-    for (auto &option : m_options)
-      strStream << option << " ";
-    std::string extraInfo = strStream.str();
-    PipelineDumper::DumpPipelineExtraInfo(reinterpret_cast<PipelineDumpFile *>(pipelineDumpFile), &extraInfo);
-  }
+  if (result == Result::Success)
+    dumpCompilerOptions(pipelineDumpFile);
 
   Optional<CacheAccessor> cacheAccessor;
   if (cl::CacheFullPipelines) {
--- a/llpc/context/llpcCompiler.h
+++ b/llpc/context/llpcCompiler.h
@@ -102,6 +102,13 @@ public:
 
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, ShaderModuleBuildOut *shaderOut);
 
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pipelineDumpFile = nullptr);
+
+  virtual Result buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *elfPackage);
+
   virtual unsigned ConvertColorBufferFormatToExportFormat(const ColorTarget *target,
                                                           const bool enableAlphaToCoverage) const;
 
@@ -166,6 +173,11 @@ private:
   bool canUseRelocatableGraphicsShaderElf(const llvm::ArrayRef<const PipelineShaderInfo *> &shaderInfo,
                                           const GraphicsPipelineBuildInfo *pipelineInfo);
   bool canUseRelocatableComputeShaderElf(const ComputePipelineBuildInfo *pipelineInfo);
+  Result buildUnlinkedShaderInternal(Context *context, llvm::ArrayRef<const PipelineShaderInfo *> shaderInfo,
+                                     Vkgc::UnlinkedShaderStage stage, ElfPackage &elfPackage,
+                                     llvm::MutableArrayRef<CacheAccessInfo> stageCacheAccesses);
+  void dumpCompilerOptions(void *pipelineDumpFile);
+
   std::vector<std::string> m_options;           // Compilation options
   MetroHash::Hash m_optionHash;                 // Hash code of compilation options
   GfxIpVersion m_gfxIp;                         // Graphics IP version info
--- a/llpc/include/llpc.h
+++ b/llpc/include/llpc.h
@@ -232,6 +232,29 @@ public:
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build unlinked shader to ElfPackage with part pipeline info.
+  ///
+  /// @param [in]  pipelineInfo     : Info to build this shader module
+  /// @param [out] pipelineOut      : Output of building this shader module
+  /// @param [in]  stage            : Shader stage of needing to compile
+  /// @param [out] pipelineDumpFile : Handle of pipeline dump file
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pipelineDumpFile = nullptr) = 0;
+
+  /// Build the whole graphics pipeline. If missing elfPackage of a certain stage, we will build it first, and
+  /// link them to the full pipeline last.
+  ///
+  /// @param [in]  pipelineInfo : Info to build this shader module
+  /// @param [out] pipelineOut  : Output of building this shader module
+  /// @param [in]  elfPackage   : Early compiled elfPackage; it is an array which size is UnlinkedStageCount.
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *elfPackage) = 0;
+
   /// Build graphics pipeline from the specified info.
   ///
   /// @param [in]  pPipelineInfo  Info to build this graphics pipeline


amdvlk-admin:

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,248 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    &pipelineInfo->fs,
+  };
+  // clang-format on
+
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+
+  if (result == Result::Success && pPipelineDumpFile) {
```
tsymalla, line 760: I think, the first part of this conditional, checking for `Result::Success` is not neccessary, as the value of `result` is not getting overwritten until here.

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build shader from the specified info.
+  ///
+  /// @param [in]  pShaderInfo    Info to build this shader module
```
tsymalla, line 237: Naming inside the comments does not correspond to the actual argument names.

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
...
+  /// @param [out] pipelineDumpFile : Handle of pipeline dump file
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pPipelineDumpFile = nullptr) = 0;
+
+  /// Build shader from the specified info.
+  ///
+  /// @param [in]  pShaderInfo    Info to build this shader module
```
tsymalla, line 249: Same.

llpc/context/llpcCompiler.cpp
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
 }
 
+// Build shader from the specified info.
+//
+// @param [in]  pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param [in]  stage : Shader stage of needing to compile
+// @param pPipelineDumpFile : Handle of pipeline dump file
+//
+// @returns : Result::Success if successful. Other return codes indicate failure.
```
kuhar, line 738: nit: please fix the documentation: missing separator line (`====`), no need to specify `[in]` only, extra line before `@returns`. Also in other documentation comments.

llpc/include/llpc.h
```diff
@@ -232,6 +232,28 @@ class ICompiler {
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build shader from the specified info.
```
kuhar, line 235: Can you add a comment explaining the difference between different different build functions? The current comment is not specific enough to tell without understanding the implementation.

llpc/include/llpc.h
```diff
@@ -232,6 +232,28 @@ class ICompiler {
...
+  ///
+  /// @param [in]  pipelineInfo    Info to build this shader module
+  /// @param [out] pipelineOut : Output of building this shader module
+  /// @param [in]  stage          Shader stage of needing to compile
+  /// @param [out] pipelineDumpFile : Handle of pipeline dump file
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pPipelineDumpFile = nullptr) = 0;
```
kuhar, line 245: nit: `pipelineDumpFile`? I think the `p` prefix is not necessary anymore

llpc/include/llpc.h
```diff
@@ -232,6 +232,28 @@ class ICompiler {
...
+
+  /// Build shader from the specified info.
```
kuhar, line 247: Also here

llpc/include/llpc.h
```diff
@@ -232,6 +232,28 @@ class ICompiler {
...
+  ///
+  /// @param [in]  pipelineInfo    Info to build this shader module
+  /// @param [out] pipelineOut  : Output of building this shader module
+  /// @param [in]  pElfpackage    Early compiled elfPackage
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *pElfpackage) = 0;
```
kuhar, line 255: Why not pass `BinaryData` by value? This `struct` is small.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+
+  if (pPipelineDumpFile) {
+    std::stringstream strStream;
+    strStream << ";Compiler Options: ";
+    for (auto &option : m_options)
+      strStream << option << " ";
```
kuhar, line 764: If you use `raw_string_ostream` you can join all options with `llvm::join`.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+// @param [in]  pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param [in]  stage : Shader stage of needing to compile
+// @param pPipelineDumpFile : Handle of pipeline dump file
+//
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, UnlinkedShaderStage stage,
+                                          void *pPipelineDumpFile) {
+  Result result = Result::Success;
```
kuhar, line 742: Don't define `result` until it's necessary. Prefer early returns when possible. See https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/CodingStandards.md#result-handling

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    for (auto &option : m_options)
+      strStream << option << " ";
+    std::string extraInfo = strStream.str();
+    PipelineDumper::DumpPipelineExtraInfo(reinterpret_cast<PipelineDumpFile *>(pPipelineDumpFile), &extraInfo);
+  }
+
+  Optional<CacheAccessor> cacheAccessor;
+  ElfPackage candidateElf;
+
+  if (!cacheAccessor || !cacheAccessor->isInCache()) {
```
kuhar, line 772: Cache accessor is always `None` here. Why do we check for `!cacheAccessor`?

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+
+    result = buildUnlinkedShaderInternal(context, shaderInfo, stage, candidateElf, pipelineOut->stageCacheAccesses);
+    releaseContext(context);
+
+    if (result == Result::Success) {
+      elfBin.codeSize = candidateElf.size();
+      elfBin.pCode = candidateElf.data();
+    }
+
+    if (cacheAccessor && pipelineOut->pipelineCacheAccess == CacheAccessInfo::CacheNotChecked)
```
kuhar, line 788: Is the `then` branch ever taken?

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    LLPC_OUTS("Cache hit for graphics pipeline.\n");
+    elfBin = cacheAccessor->getElfFromCache();
+    if (cacheAccessor->isInCache()) {
+      pipelineOut->pipelineCacheAccess =
+          cacheAccessor->hitInternalCache() ? CacheAccessInfo::InternalCacheHit : CacheAccessInfo::CacheHit;
+    }
+  }
+
+  if (result == Result::Success) {
+    void *allocBuf = nullptr;
```
kuhar, line 800: `allocBuf` can be defined inside the next if.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+//
+// @param [in]  pipelineInfo    Info to build this shader module
+// @param [out] pipelineOut  : Output of building this shader module
+// @param [in]  pElfpackage :  Early compiled elfPackage
+//
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *pElfpackage) {
+
+  Result result = Result::Success;
```
kuhar, line 832: Same as above.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true);
+
+  GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
+  Context *context = acquireContext();
+  context->attachPipelineContext(&graphicsContext);
+
+  for (unsigned idx = 0; idx < UnlinkedStageCount; idx++) {
```
kuhar, line 854: Can you iterate over all unlinked stages instead (with `enumRange`)?

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  bool hasError = false;
+  context->setDiagnosticHandler(std::make_unique<LlpcDiagnosticHandler>(&hasError));
+  hasError |= !linkRelocatableShaderElf(elf, &pipelineElf, context);
+  context->setDiagnosticHandler(nullptr);
+  if (hasError)
+    result = Result::ErrorInvalidShader;
+  releaseContext(context);
+
+  if (result == Result::Success) {
+    void *allocBuf = nullptr;
```
kuhar, line 877: Also here

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,246 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    if (pipelineInfo->pfnOutputAlloc) {
+      allocBuf = pipelineInfo->pfnOutputAlloc(pipelineInfo->pInstance, pipelineInfo->pUserData, pipelineElf.size());
+
+      uint8_t *code = static_cast<uint8_t *>(allocBuf);
+      memcpy(code, pipelineElf.data(), pipelineElf.size());
+
+      pipelineOut->pipelineBin.codeSize = pipelineElf.size();
+      pipelineOut->pipelineBin.pCode = code;
+    } else {
+      // Allocator is not specified
```
kuhar, line 887: nit: missing `.` at the end

kuhar (changes requested):
In the commit/PR message:
> When shaders are as library, we sometimes need to fastly link them. In
this case,we must to compile early and store Elfpackage, then link them
to a whole pipeline.

Typos/cosmetics: 'are as library' -> 'are a library' or 'are given as a library', 'need to fastly link them' -> 'need to quickly link them', ' case,we' -> 'case, we', 'Elfpackage' -> 'ElfPackage'.

------------

This PR needs some clean up work.

Also, can you confirm that we want to use unlinked full compilation for pipeline libraries instead of relocatable compilation?

+cc @s-perron @dstutt @trenouf

llpc/context/llpcCompiler.cpp
AMD-dwang: I think it is better to add another change to fix it.

llpc/include/llpc.h
AMD-dwang: Update the comments. Thanks.

llpc/include/llpc.h
AMD-dwang: It seems 'p' prefix is added for the pointer type in this file, I keep the same format. Thanks.

llpc/include/llpc.h
AMD-dwang: ElfPackage is array here.

llpc/context/llpcCompiler.cpp
AMD-dwang: Fixed, Thanks.

llpc/context/llpcCompiler.cpp
AMD-dwang: Yes, but index is also needed.

llpc/context/llpcCompiler.cpp
AMD-dwang: what are the differences between them?

llpc/context/llpcCompiler.cpp
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
kuhar: > I think it is better to add another change to fix it.

For the existing code in this file, fully agreed. For new code, I'd strongly prefer to get it right from the beginning.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  // Compile
+  GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
+  Context *context = acquireContext();
+  context->attachPipelineContext(&graphicsContext);
+  result = buildUnlinkedShaderInternal(context, shaderInfo, stage, candidateElf, pipelineOut->stageCacheAccesses);
+  releaseContext(context);
+  if (result == Result::Success) {
+    elfBin.codeSize = candidateElf.size();
+    elfBin.pCode = candidateElf.data();
+  }
```
kuhar, line 778: You can use an early return here

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+      void *allocBuf = pipelineInfo->pfnOutputAlloc(pipelineInfo->pInstance, pipelineInfo->pUserData, elfBin.codeSize);
+
+      uint8_t *code = static_cast<uint8_t *>(allocBuf);
+      memcpy(code, elfBin.pCode, elfBin.codeSize);
+
+      pipelineOut->pipelineBin.codeSize = elfBin.codeSize;
+      pipelineOut->pipelineBin.pCode = code;
+    } else {
+      // Allocator is not specified.
+      result = Result::ErrorInvalidPointer;
```
kuhar, line 791: Can we check that before doing all the work?

llpc/context/llpcCompiler.cpp
kuhar: There's not that much difference here. `raw_string_ostream` is much simpler than stringstream (in terms of the implementation and the API) and has a bunch of utility functions made with llvm types in mind. What I'm suggesting would be something like:

```c++
std::string buf;
raw_string_ostream os;
os << ";Compiler Options: " << join(m_options, " ");
os.flush();
PipelineDumper::DumpPipelineExtraInfo(reinterpret_cast<PipelineDumpFile *>(pipelineDumpFile), &buf);
```

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  }
+
+  Result result = Result::Success;
+  ElfPackage candidateElf;
+
+  // Compile
+  GraphicsContext graphicsContext(m_gfxIp, pipelineInfo, &pipelineHash, &cacheHash);
+  Context *context = acquireContext();
+  context->attachPipelineContext(&graphicsContext);
+  result = buildUnlinkedShaderInternal(context, shaderInfo, stage, candidateElf, pipelineOut->stageCacheAccesses);
```
kuhar, line 773: Can we define `result` here?

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+//
+// @param [in]  pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param [in]  stage : Shader stage of needing to compile
+// @param pipelineDumpFile : Handle of pipeline dump file
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, UnlinkedShaderStage stage,
+                                          void *pipelineDumpFile) {
+  BinaryData elfBin = {};
```
kuhar, line 741: Can we define `elfBin` closer to where it's first used? This function is long and I cannot quickly find its uses outside of an IDE.

llpc/context/llpcCompiler.cpp
kuhar: Also, this could be a helper (member) function, if it's used elsewhere in this file.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  context->attachPipelineContext(&graphicsContext);
+
+  Result result = Result::Success;
+  for (unsigned idx = 0; idx < UnlinkedStageCount; idx++) {
+    if (elfPackage[idx].pCode) {
+      auto data = reinterpret_cast<const char *>(elfPackage[idx].pCode);
+      elf[idx].assign(data, data + elfPackage[idx].codeSize);
+    } else {
+      graphicsContext.setUnlinked(true);
+      result = buildUnlinkedShaderInternal(context, shaderInfo, static_cast<UnlinkedShaderStage>(idx), elf[idx],
```
kuhar, line 835: What happens when one build fails and the next one succeeds?

kuhar (changes requested):
How do you plan to test the new functions?

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+
+  return result;
+}
+
+// Build the whole graphics pipeline. If missing elfPackage of a certain stage, we will build it firstly, and
+// link them to the full pipeline at last.
+//
+// @param [in]  pipelineInfo    Info to build this shader module
+// @param [out] pipelineOut  : Output of building this shader module
+// @param [in]  elfPackage :  Early compiled elfPackage
```
kuhar, line 803: Can you explain that this is an array?

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
...
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pPipelineInfo,
+                                          GraphicsPipelineBuildOut *pPipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pPipelineDumpFile = nullptr) = 0;
+
+  /// Build the whole graphics pipeline. If missing elfPackage of a certain stage, we will build it firstly, and
+  /// link them to the full pipeline at last.
+  ///
+  /// @param [in]  pPipelineInfo    Info to build this shader module
+  /// @param [out] pPipelineOut  : Output of building this shader module
+  /// @param [in]  pElfPackage    Early compiled elfPackage
```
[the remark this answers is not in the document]
kuhar: Also, please mention that elfPackage is an array and how to find its size

kuhar (changes requested):

llpc/context/llpcCompiler.cpp
AMD-dwang: Fixed.  Once it fails, we will return error code.

llpc/include/llpc.h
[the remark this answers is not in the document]
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,221 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+          pipelineInfo->pfnOutputAlloc(pipelineInfo->pInstance, pipelineInfo->pUserData, pipelineElf.size());
+
+      uint8_t *code = static_cast<uint8_t *>(allocBuf);
+      memcpy(code, pipelineElf.data(), pipelineElf.size());
+
+      pipelineOut->pipelineBin.codeSize = pipelineElf.size();
+      pipelineOut->pipelineBin.pCode = code;
+    } else {
+      // Allocator is not specified.
+      result = Result::ErrorInvalidPointer;
```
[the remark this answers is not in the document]
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
AMD-dwang: Yes, Fixed. Thanks.

llpc/context/llpcCompiler.cpp
AMD-dwang: Add a member function.

llpc/context/llpcCompiler.cpp
AMD-dwang: > How do you plan to test the new functions?

This change is to prepare for a new feature, some cts tests will be added. We can't  test it now, because there are still some changes in driver.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,222 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  return result;
+}
+
+// =====================================================================================================================
+// Build unlinked shader from the specified info.
+//
+// @param context : Acquired context
+// @param shaderInfo : Shader info of this pipeline
+// @param stage :  Shader stage of needing to compile
+// @param [out] pipelineElf : Output Elf package
```
kuhar, line 877: There is no argument called `pipelineElf`.

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build unlinked shader to ElfPackage with pipeline info.
+  ///
+  /// @param [in]  pPipelineInfo       Info to build this shader module
+  /// @param [out] pPipelineOut      : Output of building this shader module
+  /// @param [in]  stage               Shader stage of needing to compile
```
kuhar, line 239: Missing `:`

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
...
+  ///
+  /// @returns : Result::Success if successful. Other return codes indicate failure.
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pPipelineInfo,
+                                          GraphicsPipelineBuildOut *pPipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pPipelineDumpFile = nullptr) = 0;
+
+  /// Build the whole graphics pipeline. If missing elfPackage of a certain stage, we will build it first, and
+  /// link them to the full pipeline last.
+  ///
+  /// @param [in]  pPipelineInfo   Info to build this shader module
```
kuhar, line 250: Missing `:`

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
...
+  /// @param [out] pPipelineOut  : Output of building this shader module
+  /// @param [in]  pElfPackage     Early compiled elfPackage; it is an array which size is UnlinkedStageCount.
```
kuhar, line 252: Missing `:`

kuhar (dismissed):
Thanks for the changes, I find it easier to follow the code now.

My remaining concern is lack of tests and not being able to exercise the new functions with `amdllpc`.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,222 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    if (elfPackage[idx].pCode) {
+      auto data = reinterpret_cast<const char *>(elfPackage[idx].pCode);
+      elf[idx].assign(data, data + elfPackage[idx].codeSize);
+    } else {
+      graphicsContext.setUnlinked(true);
+      result = buildUnlinkedShaderInternal(context, shaderInfo, static_cast<UnlinkedShaderStage>(idx), elf[idx],
+                                           pipelineOut->stageCacheAccesses);
+      if (result != Result::Success) {
+        releaseContext(context);
+        return result;
```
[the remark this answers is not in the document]
kuhar: You can grep for existing uses of `make_scope_exit` in llpc for more examples

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,222 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
   return format_bytes(bytes, /* FirstByteOffset = */ None, /* NumPerLine = */ 16, /* ByteGroupSize = */ 1);
 }
 
+// =====================================================================================================================
+// Helper function for dumping compiler options
+//
+// @param pipelineDumpFile : Handle of pipeline dump file
+void Compiler::dumpCompilerOptions(void *pipelineDumpFile) {
+  if (pipelineDumpFile) {
```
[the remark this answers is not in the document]
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,222 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  }
+}
+
+// =====================================================================================================================
+// Build unlinked shader to ElfPackage with pipeline info.
+//
+// @param pipelineInfo : Info to build this shader module
+// @param [out] pipelineOut : Output of building this shader module
+// @param stage : Shader stage of needing to compile
+// @param pipelineDumpFile : Handle of pipeline dump file
```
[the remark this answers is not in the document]
AMD-dwang: Fixed. Thanks.

llpc/context/llpcCompiler.cpp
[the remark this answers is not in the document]
AMD-dwang: It is useful. Thanks.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +728,222 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+// @param context : Acquired context
+// @param shaderInfo : Shader info of this pipeline
+// @param stage :  Shader stage of needing to compile
+// @param [out] pipelineElf : Output Elf package
+// @param [out] stageCacheAccesses : Stage cache access result. All elements
+//                                   must be initialized in the caller as
+//                                   CacheAccessInfo::CacheNotChecked
+// @returns : Result::Success if successful. Other return codes indicate failure.
+Result Compiler::buildUnlinkedShaderInternal(Context *context, ArrayRef<const PipelineShaderInfo *> shaderInfo,
+                                             Vkgc::UnlinkedShaderStage stage, ElfPackage &elfPackage,
```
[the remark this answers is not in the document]
AMD-dwang: They are the same. removed namespace.

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build unlinked shader to ElfPackage with pipeline info.
+  ///
+  /// @param [in]  pPipelineInfo       Info to build this shader module
```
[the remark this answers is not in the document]
AMD-dwang: In this file, there is no ":" for [in] only. 
But I add it now. Thanks.

AMD-dwang:
> @AMD-dwang
> 
> > This change is to prepare for a new feature, some cts tests will be added. We can't test it now, because there are still some changes in driver.
> 
> Can't we call those from amdllpc? Or have limited unit tests that check for edge cases like missing allocators.

These are for pipeline library, amd pc can't build library now, If we hope to support it, more work is needed. For unit tests, I don't know what should be done, are there some example?

kuhar:
@AMD-dwang 
>  For unit tests, I don't know what should be done, are there some example?
You can add a new unit test file under `llpc/unittests/context`. Here, I was only thinking about some really simple tests that do not do any compilation and only ensure we handle edge cases that result in error results, e.g., missing allocation functions.

> These are for pipeline library, amdllpc can't build library now, If we hope to support it, more work is needed.

But what's the plan for amdllpc to exercise these functions eventually? I assumed that the unlinked compilation would fit the existing unliked compilation scheme in amdllpc.

I think this is really important for both manual investigations and having a robust regression test suite. We should be able to compile code the same way as the driver without having to run a system with a GPU. For example, when we suspect a miscompilation or experience a compiler crash, we should be able to take the dumped `.pipe` file and replay the same compilation with amdllpc. In my options, as long as you have a design document ([sample existing ddns](https://github.com/GPUOpen-Drivers/llpc/tree/dev/docs)) that explains how things like this will interact and lead to a robust solution, it's fine to start with a partial solution that is not fully fleshed out. But again, this assumes that you start the design with testability and debuggability in mind.

AMD-dwang:
> @AMD-dwang
> 
> > For unit tests, I don't know what should be done, are there some example?
> 
> You can add a new unit test file under `llpc/unittests/context`. Here, I was only thinking about some really simple tests that do not do any compilation and only ensure we handle edge cases that result in error results, e.g., missing allocation functions.
> 
> > These are for pipeline library, amdllpc can't build library now, If we hope to support it, more work is needed.
> 
> But what's the plan for amdllpc to exercise these functions eventually? I assumed that the unlinked compilation would fit the existing unliked compilation scheme in amdllpc.
> 
> I think this is really important for both manual investigations and having a robust regression test suite. CTS is not enough by itself, as we can't run it on all development workstations and in the public CI, and where we can it is very slow to run all tests. We should be able to compile code the same way as the driver without having to run a system with a GPU. For example, when we suspect a miscompilation or experience a compiler crash, we should be able to take the dumped `.pipe` file and replay the same compilation with amdllpc. In my options, as long as you have a design document ([sample existing ddns](https://github.com/GPUOpen-Drivers/llpc/tree/dev/docs)) that explains how things like this will interact and lead to a robust solution, it's fine to start with a partial solution that is not fully fleshed out. But again, this assumes that you start the design with testability and debuggability in mind.

Okay, I will add a new issue to record the task of adding the test. Could you approve this change?

kuhar:
> Okay, I will add a new issue to record the task of adding the test. Could you approve this change?

I don't think moving forward with this PR is reasonable without considering the things I mentioned first. It should only progress once there is agreement on the higher-level design and test plan. If the PR went in as-is, it will leave a lot of unexercised and untested code that's nearly impossible to maintain.

kuhar:
Another open question is whether to use relocatable compilation for pipeline libraries or not. Last time we discussed this during the bi-weekly sync, @trenouf was supposed to be the point of contact for this development work.

AMD-dwang:
> > Okay, I will add a new issue to record the task of adding the test. Could you approve this change?
> 
> I don't think moving forward with this PR is reasonable without considering the things I mentioned first. It should only progress once there is agreement on the higher-level design and test plan. If the PR went in as-is, it will leave a lot of unexercised and untested code that's nearly impossible to maintain.

But it should  be done step by step. If a large feature must to be completed in a change, then the amount of code will become very large. And this change will affect nothing.

kuhar:
> But it should be done step by step. If a large feature must to be completed in a change, then the amount of code will become very large. And this change will affect nothing.

Absolutely, but I'm not convinced this is the right first step.

If I were to develop this feature, I'd approach it something like this:
-  Start a thread on the mailing list or set up a meeting/video call where you discuss what compiler work needs to be done to support pipeline libraries. Gather requirements and constraints.
-  Add placeholder entry points (the unlinked compilation functions) to llpc.h and llpcCompiler.h. Make them return errors and not compile anything. This does not introduce any new logic, allows us to agree on the new API, and we can make XGL use it.
-  Develop a proof of concept on my personal fork. Make sure things work on a limited set of tests.
-  Based on this prototype, present a high-level design (mailing list, design document, or a meeting/video call). Explain what the main work items are, what disruptive changes you intend to make, let people know what to expect in the future PRs.
  *  In this specific case, we definitely need to make sure that we can compile pipeline libraries offline. Discuss how this will be enabled in amdllpc, discuss how this affects the `.pipe` file format, explain what tests we need to make sure this works as intended.
- Modify the design based on the feedback. Concurrently, start sending out non-functional changes that you know be necessary to implement the feature. For example, we could start by cleaning up the existing header file documentation.
-  Gradually clean up the prototype and add missing features. Split this into multiple and independent PRs that are easy to review. Ideally, each PR should be either exercised by existing LIT or unit tests, or new tests should be added together with the new code, so that no new code is unused in our test suite.

I strongly object moving forward with this PR without discussing the design. In addition to the maintainability issues I mentioned, my concern is that I'm not yet convinced that we want to implement pipeline libraries this way (without reloc shaders), and I don't want us to re-do most of the implementations later if we discover this is not what we wanted/needed.

AMD-dwang:
> > But it should be done step by step. If a large feature must to be completed in a change, then the amount of code will become very large. And this change will affect nothing.
> 
> Absolutely, but I'm not convinced this is the right first step.
> 
> If I were to develop this feature, I'd approach it something like this:
> 
> * Start a thread on the mailing list or set up a meeting/video call where you discuss what compiler work needs to be done to support pipeline libraries. Gather requirements and constraints.
> * Add placeholder entry points (the unlinked compilation functions) to llpc.h and llpcCompiler.h. Make them return errors and not compile anything. This does not introduce any new logic, allows us to agree on the new API, and we can make XGL use it.
> * Develop a proof of concept on my personal fork. Make sure things work on a limited set of tests.
> * Based on this prototype, present a high-level design (mailing list, design document, or a meeting/video call). Explain what the main work items are, what disruptive changes you intend to make, let people know what to expect in the future PRs.
>   
>   * In this specific case, we definitely need to make sure that we can compile pipeline libraries offline. Discuss how this will be enabled in amdllpc, discuss how this affects the `.pipe` file format, explain what tests we need to make sure this works as intended.
> * Modify the design based on the feedback. Concurrently, start sending out non-functional changes that you know be necessary to implement the feature. For example, we could start by cleaning up the existing header file documentation.
> * Gradually clean up the prototype and add missing features. Split this into multiple and independent PRs that are easy to review. Ideally, each PR should be either exercised by existing LIT or unit tests, or new tests should be added together with the new code, so that no new code is unused in our test suite.
> 
> I strongly object moving forward with this PR without discussing the design. In addition to the maintainability issues I mentioned, my concern is that I'm not yet convinced that we want to implement pipeline libraries this way (without reloc shaders), and I don't want us to re-do most of the implementations later if we discover this is not what we wanted/needed.

Yes, if the design is unreasonable, I can accept it. But if it's because no tests have been added to block the change, I can't.
Thanks. Let's wait for @trenouf's reply.

Flakebi:
As a sidenote:
LLVM has the policy that if a patch changes behavior, i.e. it’s not only a refactoring, it can only be accepted if it also has an accompanying test. While I found it a bit cumbersome when I started contributing to LLVM, I now find it useful to have all these tests.
It means, I can do large refactorings and if the test suite still passes afterwards, I can be confident that I did not break anything because every feature or past regression has a test.
With the large number of patches that land in LLVM every day, I think the tests are what makes contributing possible at all. Without them, we would be hit with dozens of regressions every week.

For features that are work in progress and not yet tested in CTS (e.g. because it’s not yet implemented in the driver), it’s even more important to have *some* tests that are executed, so regressions can’t slip in unnoticed in other compiler changes.
When working on compute libraries (and the related – currently internal – feature), I didn’t add compiler tests at first. That lead to a state where every few weeks, when I was manually testing, some part of the new code were broken because of other changes in the compiler. (After adding tests, we still got regressions now and then that failed the tests, but it’s easy to pinpoint a commit that introduce a unit-test failure.)
I hope we don’t repeat the mistake I made back then :)

s-perron:
> Another open question is whether to use relocatable compilation for pipeline libraries or not. Last time we discussed this during the bi-weekly sync, @trenouf was supposed to be the point of contact for this development work.

I think this implementation is fine at a high level.  It is reusing the same code to actually compile the unlinked shader.  The only difference is that this interface gets the pipeline info in a state where it can be compiled as a unlinked shader.  For relocatable shaders, we get the entire pipeline state and we have to divide it up.  That seems to be the only real difference.

s-perron:
I agree that we need to be able to have a way to test this.  I've already pointed out on error with the hashing.  I have questions on where we are going with this longer term.  Will we cache the full pipepline?

Also the interface is not well defined.  For example, if you call `buildGraphicsShaderStage` for the fragment stage, must all shader stages except for "fs" be null?  If not, then I think we will have some errors.  If so, then it should be mentioned in the function comments and have asserts that will check it.  It is things like this that a proper tests will help us check more easily.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +729,218 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+        return result;
+    }
+  }
+
+  // Link
+  graphicsContext.setUnlinked(false);
+  ElfPackage pipelineElf;
+  bool hasError = false;
+  context->setDiagnosticHandler(std::make_unique<LlpcDiagnosticHandler>(&hasError));
+  hasError |= !linkRelocatableShaderElf(elf, &pipelineElf, context);
```
s-perron, line 854: You should probably be looking for cases where the linking cannot be done.  Not all of them are recognized at link time.  You should probably call `canUseRelocatableComputeShaderElf` at some point in this function.

llpc/context/llpcCompiler.cpp
```diff
@@ -728,6 +729,218 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  if (!pipelineInfo->pfnOutputAlloc)
+    return Result::ErrorInvalidPointer;
+
+  // clang-format off
+  SmallVector<const PipelineShaderInfo *, ShaderStageGfxCount> shaderInfo = {
+    &pipelineInfo->vs,
+    &pipelineInfo->tcs,
+    &pipelineInfo->tes,
+    &pipelineInfo->gs,
+    &pipelineInfo->fs,
```
s-perron, line 766: Are you sure you want every shader stage from the pipeline info?  What if all you are doing is the vertex processing stage?

s-perron: I see where the stages get cleared out now.  Sorry.  This is fine.

llpc/context/llpcCompiler.cpp
s-perron, line 854: I there a plan for how the handle the glue code?  This currently compiles the glue code at link time.  Will we be exposing an interface to compile the vertex fetch shader and color export shader as well?

AMD-dwang:
> As a sidenote: LLVM has the policy that if a patch changes behavior, i.e. it’s not only a refactoring, it can only be accepted if it also has an accompanying test. While I found it a bit cumbersome when I started contributing to LLVM, I now find it useful to have all these tests. It means, I can do large refactorings and if the test suite still passes afterwards, I can be confident that I did not break anything because every feature or past regression has a test. With the large number of patches that land in LLVM every day, I think the tests are what makes contributing possible at all. Without them, we would be hit with dozens of regressions every week.
> 
> For features that are work in progress and not yet tested in CTS (e.g. because it’s not yet implemented in the driver), it’s even more important to have _some_ tests that are executed, so regressions can’t slip in unnoticed in other compiler changes. When working on compute libraries (and the related – currently internal – feature), I didn’t add compiler tests at first. That lead to a state where every few weeks, when I was manually testing, some part of the new code were broken because of other changes in the compiler. (After adding tests, we still got regressions now and then that failed the tests, but it’s easy to pinpoint a commit that introduce a unit-test failure.) I hope we don’t repeat the mistake I made back then :)

Thank you for saying so much, the test is definitely needed, but I want to delay to add them. This change is only a part, I want to complete the whole process first, may as @kuhar mentioned, if a certain interface design is unreasonable, then we must to refine them, so the current tests will become meaningless. Anyway thank you and @kuhar very much.

AMD-dwang:
I updated this change for pipeline library.  @kuhar @s-perron

amdvlk-admin:
### Test summary for commit 8130d954834dc56c951cd3133ed951588a78fe28

</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/context/llpcCompiler.cpp
```diff
@@ -571,6 +572,253 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+    llvm_unreachable("");
+    break;
+  }
+
+  dumpCompilerOptions(pipelineDumpFile);
+
+  MetroHash::Hash cacheHash = {};
+  MetroHash::Hash pipelineHash = {};
+  cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
+  pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
```
s-perron, line 634: This should be

```
pipelineHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, /* isCacheHash = */ false, true, stage);
```

because this is not the cache hash.

llpc/context/llpcCompiler.cpp
```diff
@@ -571,6 +572,253 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+        if (result != Result::Success)
+          return result;
+      }
+    }
+
+    // Link
+    graphicsContext.setUnlinked(false);
+    bool hasError = false;
+    context->setDiagnosticHandler(std::make_unique<LlpcDiagnosticHandler>(&hasError));
+    hasError |= !linkRelocatableShaderElf(elf, &pipelineElf, context);
```
s-perron, line 725: I think I mentioned this before.  You should call `Compiler::canUseRelocatableGraphicsShaderElf` at some point early in this function.  There are some cases that will generate incorrect code if you try to link the unlinked elf.  That function lets you know.

llpc/context/llpcCompiler.cpp
```diff
@@ -571,6 +572,253 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+  MetroHash::Hash cacheHash = {};
+  if (context->isGraphics()) {
+    auto pipelineInfo = reinterpret_cast<const GraphicsPipelineBuildInfo *>(context->getPipelineBuildInfo());
+    cacheHash = PipelineDumper::generateHashForGraphicsPipeline(pipelineInfo, true, true, stage);
+  } else {
+    auto pipelineInfo = reinterpret_cast<const ComputePipelineBuildInfo *>(context->getPipelineBuildInfo());
+    cacheHash = PipelineDumper::generateHashForComputePipeline(pipelineInfo, true, true);
+  }
+  // Note that this code updates m_pipelineHash of the pipeline context. It
+  // must be restored before we link the pipeline ELF at the end of this for-loop.
```
s-perron, line 778: This comment is stale because it is in a new function.

llpc/context/llpcCompiler.cpp
```diff
@@ -571,6 +572,253 @@ template <typename T> static FormattedBytes formatBytesLittleEndian(ArrayRef<T>
...
+                                   stageCacheAccesses);
+    if (result == Result::Success) {
+      // Add the result to the cache.
+      BinaryData elfBin = {elfPackage.size(), elfPackage.data()};
+      cacheAccessor.setElfInCache(elfBin);
+      LLPC_OUTS("Updating the cache for unlinked shader stage " << getUnlinkedShaderStageName(stage) << "\n");
+    }
+  }
+  context->getPipelineContext()->setShaderStageMask(originalShaderStageMask);
+  context->getPipelineContext()->setHashForCacheLookUp(originalCacheHash);
```
s-perron, line 818: Since we are refactoring this code.  Could you put this into an `make_scope_exit` like you did to release the context above?

llpc/context/llpcCompiler.cpp
```diff
@@ -602,66 +849,9 @@ Result Compiler::buildPipelineWithRelocatableElf(Context *context, ArrayRef<cons
...
-    const PipelineShaderInfo *singleStageShaderInfo[ShaderStageNativeStageCount] = {nullptr, nullptr, nullptr,
-                                                                                    nullptr, nullptr, nullptr};
-    for (ShaderStage stage : shaderStages)
-      singleStageShaderInfo[stage] = shaderInfo[stage];
-
-    Vkgc::ElfPackage &stageElf = elf[stage];
-    result = buildPipelineInternal(context, singleStageShaderInfo, PipelineLink::Unlinked, nullptr, &stageElf,
-                                   stageCacheAccesses);
-    if (result != Result::Success)
+    if (buildUnlinkedShaderInternal(context, shaderInfo, stage, elf[stage], stageCacheAccesses) != Result::Success)
```
s-perron, line 852: I like this.  This function was too long.  It helps readability as well as code reuse.

llpc/context/llpcCompiler.h
```diff
@@ -102,6 +102,13 @@ class Compiler : public ICompiler {
 
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *shaderInfo, ShaderModuleBuildOut *shaderOut);
 
+  virtual Result buildGraphicsShaderStage(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                          GraphicsPipelineBuildOut *pipelineOut, Vkgc::UnlinkedShaderStage stage,
+                                          void *pipelineDumpFile = nullptr);
+
+  virtual Result buildGraphicsPipelineWithElf(const GraphicsPipelineBuildInfo *pipelineInfo,
+                                              GraphicsPipelineBuildOut *pipelineOut, const BinaryData *elfPackage);
+
```
s-perron, line 111: Do you have a testing strategy for these two functions? `buildUnlinkedShaderInternal` should be testeded using the relocatable shader tests.  However the new public functions should be tested, I'm just not sure how extensively.

s-perron (changes requested):
This generally looks good.  There are somethings that are missing, but I think you stated this is just the start.  I'd like to see the larger design if possible.  For example, I believe you might want to be able to compile the glue shader ahead of time.  This interface does not seem to allow passing in the glue shaders.

llpc/context/llpcCompiler.h
AMD-dwang: For `buildGraphicsShaderStage`, we will generate a part pipeline file which only includes  relevant shaders information. For  `buildGraphicsPipelineWithElf`, we don't have a good solution yet, currently only test it with cts.

amdvlk-admin:
### Test summary for commit d6de93bfb62a2dd90632fe0f3c271e94ed83074e
<details><summary>CTS tests (Failed: 0/187823)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36645/65226 (56.2%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 28581/65226 (43.8%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>

<li>Passed: 31111/57371 (54.2%)</li>
<li>Failed: 0/57371 (0.0%)</li>
<li>Not Supported: 26260/57371 (45.8%)</li><li>Warnings: 0/57371 (0.0%)</li>
</ul></details>

<li>Passed: 37804/65226 (58.0%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 27422/65226 (42.0%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>


</ul></details>

amdrexu:
There are code conflicts. Please rebase this PR.

amdvlk-admin:
### Test summary for commit 650da5dc9c24a03cd18df096dd79226740b95a9f
<details><summary>CTS tests (Failed: 0/186422)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36645/65226 (56.2%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 28581/65226 (43.8%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>

<li>Passed: 30002/55970 (53.6%)</li>
<li>Failed: 0/55970 (0.0%)</li>
<li>Not Supported: 25968/55970 (46.4%)</li><li>Warnings: 0/55970 (0.0%)</li>
</ul></details>

<li>Passed: 38946/65226 (59.7%)</li>
<li>Failed: 0/65226 (0.0%)</li>
<li>Not Supported: 26280/65226 (40.3%)</li><li>Warnings: 0/65226 (0.0%)</li>
</ul></details>


</ul></details>

amdrexu:
@kuhar @s-perron We would invite you to review the update of this PR. Feel free to make further comments.

amdrexu:
Any further comments for this PR?

s-perron (commented):
I don't have any more comments.

llpc/include/llpc.h
```diff
@@ -232,6 +232,30 @@ class ICompiler {
   /// @returns : Result::Success if successful. Other return codes indicate failure.
   virtual Result BuildShaderModule(const ShaderModuleBuildInfo *pShaderInfo, ShaderModuleBuildOut *pShaderOut) = 0;
 
+  /// Build unlinked shader to ElfPackage with part pipeline info.
+  ///
+  /// @param [in]  pPipelineInfo     : Info to build this shader module
```
amdrexu, line 237: For historical reason, we keep old coding style for existing codes. But for newly added interfaces, please conform to current coding conventions: Don't use "p" prefix. Remove it in naming and comments.

Please go through this PR and close all pending review conversation if it is resolved. Thanks.

AMD-dwang: Fixed. Thanks.

amdvlk-admin:
### Test summary for commit da3c60786f4d517f2d267b10a8722bb8095c0380
<details><summary>CTS tests (Failed: 0/177309)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 35753/62859 (56.9%)</li>
<li>Failed: 0/62859 (0.0%)</li>
<li>Not Supported: 27106/62859 (43.1%)</li><li>Warnings: 0/62859 (0.0%)</li>
</ul></details>

<li>Passed: 27845/51589 (54.0%)</li>
<li>Failed: 0/51589 (0.0%)</li>
<li>Not Supported: 23744/51589 (46.0%)</li><li>Warnings: 0/51589 (0.0%)</li>
</ul></details>

<li>Passed: 37998/62861 (60.4%)</li>
<li>Failed: 0/62861 (0.0%)</li>
<li>Not Supported: 24863/62861 (39.6%)</li><li>Warnings: 0/62861 (0.0%)</li>
</ul></details>


</ul></details>

amdrexu (approved):

PR #1612, opened 2022-01-04
[Windows] _fsopen for multi-thread

dstutt:
fopen_s can't be used for multi-threaded - shareable isn't allowed. Use _fsopen instead.

### DIFF
--- a/llpc/util/llpcFile.cpp
+++ b/llpc/util/llpcFile.cpp
@@ -111,8 +111,13 @@ Result File::open(const char *filename, unsigned accessFlags) {
 
     if (result == Result::Success) {
 #if defined(_WIN32)
-      // MS compilers provide fopen_s, which is supposedly "safer" than traditional fopen.
-      fopen_s(&m_fileHandle, filename, &fileMode[0]);
+      // MS compilers provide _fsopen, which allows the shareable modes to be
+      // specified.
+      // In this case _SH_DENYNO which effectively makes it the equivalent of
+      // fopen.
+      // Note: changed from fopen_s which doesn't allow shareable access
+      // (required for multi-thread).
+      m_fileHandle = _fsopen(filename, &fileMode[0], _SH_DENYNO);
 #else
       // Just use the traditional fopen.
       m_fileHandle = fopen(filename, &fileMode[0]);


jayfoad:
Not sure I'm qualified to review this. When is "sharable" required - is it for multiple threads in the same process reading the same input file? (What about multiple processes reading the same input file?) Is it harmless in other cases, like opening an output file for writing to?

dstutt:
This change makes the Windows open call act more like the non-windows version (fopen has no restrictions like this). The potential fly in the ointment here is that the implementation as it stands here will also allow multiple threads to write to the same file, which can have strange results if those writes aren't coordinated (e.g. using a mutex).
I could make this safer by only allowing the shareable version of the call when the mode is read-only, but then there's different behaviour on the 2 platforms.

dstutt:
Answering your other questions: this is only a problem when trying to access the same file in different threads. It isn't really a thread-safe issue, but rather a concurrent file access problem, so it's the same whether it is multiple threads in the same process, or multiple processes. (Of course, how you might resolve it if required could be a thread-safe problem - but I'm not attempting to address that).

dstutt:
BTW - this fixes an issue in the llpc tests on Windows. Any test that runs more than one thread can have this issue (although depending on timings, it sometimes passes). e.g. llpc\test\shaderdb\multiple_inputs\SpirvTwoEntryPoints.spvasm

llpc/util/llpcFile.cpp
```diff
@@ -111,8 +111,11 @@ Result File::open(const char *filename, unsigned accessFlags) {
 
     if (result == Result::Success) {
 #if defined(_WIN32)
-      // MS compilers provide fopen_s, which is supposedly "safer" than traditional fopen.
-      fopen_s(&m_fileHandle, filename, &fileMode[0]);
+      // MS compilers provide _fsopen, which is supposedly "safer" than
```
jayfoad, line 114: Does the "supposedly safer" comment still apply? I thought it was only for _s versions of standard library functions.

jayfoad (dismissed):
Seems reasonable.

amdvlk-admin:
### Test summary for commit 81340df7c3ffade5fe9feeeb4a54f9357a003a09

<li>CWPACK: amd-master 39f8940199e60c44d4211cf8165dfd12876316fa</li>
<li>PAL: dev bcec463efe5260776d486a5e3da0c549bc0a75d2</li>
<li>SPVGEN: dev 580ade71d630a304805e9468ef22ad263c37d137</li>
<li>XGL: dev 04534321e450d220d404a240a047c7e1cd82f3fd</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev c6950e03a4c97beadc88a8bf5bea3dc5aa2649ee</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/225376)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 32541/56344 (57.8%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23803/56344 (42.2%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>

<li>Passed: 32381/56344 (57.5%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23962/56344 (42.5%)</li><li>Warnings: 1/56344 (0.0%)</li>
</ul></details>

<li>Passed: 32717/56344 (58.1%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23627/56344 (41.9%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>

<li>Passed: 35698/56344 (63.4%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 20646/56344 (36.6%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>


</ul></details>

llpc/util/llpcFile.cpp
dstutt: Reading the docs... fopen_s is mainly safer because it prevents shareable access. I guess this one is less safe.
Perhaps it makes more sense to use the "_SH_DENYWR" option - this one doesn't prevent multiple read only accesses, but does stop anyone else writing to the file until we're done with it. That should fix the issue that's the problem in the testing, but still be safer in terms of concurrent write. (Then the comment can stay the same).

kuhar: +1. I think that the original comment might have referred to https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170:
> More-secure versions of these functions that perform additional parameter validation and return error codes are available; see fopen_s, _wfopen_s.

`_fsopen` does not return any error codes, so the comment does not seem to apply anymore

kuhar (dismissed):

llpc/util/llpcFile.cpp
dstutt: I've re-worded it (I also tried using SH_DENYWR, but that still failed).

kuhar (approved):

vettoreldaniele:
Note that using one of the sharing constants by name requires `<share.h>` to be imported.

kuhar:
retest this please

amdvlk-admin:
### Test summary for commit bfd8dc80e856315f82d6f57aaf8c554d44d426c6

<li>CWPACK: amd-master 39f8940199e60c44d4211cf8165dfd12876316fa</li>
<li>PAL: dev bcec463efe5260776d486a5e3da0c549bc0a75d2</li>
<li>SPVGEN: dev 580ade71d630a304805e9468ef22ad263c37d137</li>
<li>XGL: dev 04534321e450d220d404a240a047c7e1cd82f3fd</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev c6950e03a4c97beadc88a8bf5bea3dc5aa2649ee</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/225376)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 32541/56344 (57.8%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23803/56344 (42.2%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>

<li>Passed: 32381/56344 (57.5%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23962/56344 (42.5%)</li><li>Warnings: 1/56344 (0.0%)</li>
</ul></details>

<li>Passed: 32717/56344 (58.1%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 23627/56344 (41.9%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>

<li>Passed: 35698/56344 (63.4%)</li>
<li>Failed: 0/56344 (0.0%)</li>
<li>Not Supported: 20646/56344 (36.6%)</li><li>Warnings: 0/56344 (0.0%)</li>
</ul></details>


</ul></details>

amdrexu (approved):

JaxLinAMD:
retest this please

dstutt:
I think that the diagnosis of the problem is incorrect, or at least incomplete.
I've seen the problem on Linux as well now.

In the test case where it is going wrong, the input is an spvasm file, which is assembled into a binary spv file before being processed. In the multi-threaded case it is writing this file out in 2 threads, which causes the issue on windows (both reading and writing to the same file in 2 threads, neither of which is shareable by default), plus on Linux writing to the same file from both threads - this results in fewer failures, but timing will occasionally produce inconsistent results and failures.

Closing this PR. I'm going to work on a different solution.

kuhar:
> In the test case where it is going wrong, the input is an spvasm file, which is assembled into a binary spv file before being processed. In the multi-threaded case it is writing this file out in 2 threads, which causes the issue on windows (both reading and writing to the same file in 2 threads, neither of which is shareable by default), plus on Linux writing to the same file from both threads - this results in fewer failures, but timing will occasionally produce inconsistent results and failures.

Perhaps we should create temporary spv files in unique temporary directories? https://llvm.org/doxygen/FileSystem_8h.html has support for this.

PR #1615, opened 2022-01-06
[CI] Remove conditional based on secrets.

vettoreldaniele:
Conditionals based on secrets are not supported. The `github.repository == 'GPUOpen-Drivers/llpc'` condition should suffice now that the workflow code is checked in.

Also pin the `setup-gcloud` action to version `v0` as suggested by the warning.


amdvlk-admin:

kuhar (approved):

kuhar:
@amdrexu or @dstutt or @trenouf Seems like this needs approval before the workflow is allowed to pick up the secret and succeed. Could you try approving it and then hitting re-run all checks for the LLPC Docker CI workflow, so that we can confirm this work? My approval alone does not suffice.

vettoreldaniele:
Closing this as I need to create a new one re-applying the reverted changes and the fix.

PR #1634, opened 2022-01-13
Add provoking vertex mode

AMD-dwang:
Provoking vertex mode Specifies which vertex of a primitive is the _provoking
vertex_, this impacts which vertex's "flat" VS outputs are passed to the PS.
We will need this to fix the barycentric coordinate.


amdvlk-admin:

PR #1636, opened 2022-01-13
Fix primitiveID

AMD-dwang:
Currently we are using vertex0 as provoking vertex to distribute primitive ID,
but when provoking vertex is last, we should use vertex2, at that time,
we will get dirty data, so we need to write correct value for each vertex, this
behavior is consistent with HW.


amdvlk-admin:

amdrexu (approved):

PR #1640, opened 2022-01-13
Fix writing of relocation section in elf

bsaleil:
When writing relocations to elf files, we currently use a wrong relocation size. Because of that, we write random bytes after each relocation, causing nondeterminism. I don't really understand how this never caused issues before when we have >1 relocations.


s-perron (commented):
It probably does not trigger because there are very few cases where there are relocataions that need to be passed to PAL.  I think this should have a test.  This part of the code is probably not commonly run.

amdvlk-admin:
### Test summary for commit 1330e65074479a48520e4dc680cda6fedb5a03d9

<li>CWPACK: amd-master 39f8940199e60c44d4211cf8165dfd12876316fa</li>
<li>PAL: dev bcec463efe5260776d486a5e3da0c549bc0a75d2</li>
<li>SPVGEN: dev 580ade71d630a304805e9468ef22ad263c37d137</li>
<li>XGL: dev 0429e80c4fa09141362e586d785743ee80cfc551</li>
<li>LLVM-PROJECT: amd-gfx-gpuopen-dev c20ba7614ecdc78b3d99601423783100835200c1</li>
</ul></details>

<details><summary>CTS tests (Failed: 0/228192)</summary><ul><li>Built with version 1.2.5.2</li>

<li>Passed: 33955/57048 (59.5%)</li>
<li>Failed: 0/57048 (0.0%)</li>
<li>Not Supported: 23093/57048 (40.5%)</li><li>Warnings: 0/57048 (0.0%)</li>
</ul></details>

<li>Passed: 33797/57048 (59.2%)</li>
<li>Failed: 0/57048 (0.0%)</li>
<li>Not Supported: 23250/57048 (40.8%)</li><li>Warnings: 1/57048 (0.0%)</li>
</ul></details>

<li>Passed: 34085/57048 (59.7%)</li>
<li>Failed: 0/57048 (0.0%)</li>
<li>Not Supported: 22963/57048 (40.3%)</li><li>Warnings: 0/57048 (0.0%)</li>
</ul></details>

<li>Passed: 35882/57048 (62.9%)</li>
<li>Failed: 0/57048 (0.0%)</li>
<li>Not Supported: 21166/57048 (37.1%)</li><li>Warnings: 0/57048 (0.0%)</li>
</ul></details>


</ul></details>

bsaleil:
@s-perron, this is actually often triggered by the tests. I originally thought this would cause an issue when we have more than one relocation, but the random bytes are actually inserted after all the relocations and not after each relocation, so that's not really an issue for PAL, it is only for determinism.

piotrAMD:
Thanks for the fix.

Would it be possible to add a reliable test for that (for example using llvm-objdump)?

s-perron:
> @s-perron, this is actually often triggered by the tests. I originally thought this would cause an issue when we have more than one relocation, but the random bytes are actually inserted after all the relocations and not after each relocation, so that's not really an issue for PAL, it is only for determinism.

If it is triggered by existing tests, can you update one of those tests to look at the elf to make sure the bytes are not ransom.  @piotrAMD  suggestion seems pretty reasonable.

bsaleil:
Added a test case. I copied an existing test case generating several relocations and added run lines to check the relocation size is as expected. These lines are a bit hacky but I didn't find any cleaner way to check that.

lgc/test/ElfRelocationSize.lgc
```diff
@@ -0,0 +1,221 @@
...
+; RUN: readelf -W -S %t.pipe.elf | grep -P -o ".rodata.cst32\s+PROGBITS\s+[0-9]+\s+\S+" | awk '{ print $4 }' >%t.offs2
+; RUN: echo "obase=16;ibase=16;$(cat %t.offs2 | tr a-z A-Z)-$(cat %t.offs1 | tr a-z A-Z)" | bc | FileCheck %s
+
+; This test checks that no random extra bytes are generated after the relocations
+; in the relocation section (.rel.text). To check that, we extract the offsets of
+; the relocation section and the following section from the elf and subtract
+; them. Here we write 4 16-bit relocations, so the printed value should be
+; around 0x40 (because of section alignment).
+
+; CHECK: 41
```
kuhar, line 14:
```suggestion
; This test checks that no random extra bytes are generated after the relocations
; in the relocation section (`.rel.text`). To check that, we extract the offsets of
; the relocation section and the following section from the elf and subtract
; them. Here we write 4 16-bit relocations, so the printed value should be
; around 0x40 (65 dec) (because of section alignment).

; RUN: lgc -mcpu=gfx1030 -extract=2 -o %t.fs.elf %s
; RUN: lgc -mcpu=gfx1030 -extract=3 -other=%t.fs.elf -o %t.vs.elf %s
; RUN: lgc -mcpu=gfx1030 -extract=1 -l %s -o %t.pipe.elf %t.vs.elf %t.fs.elf
;
; RUN: llvm-readelf %t.pipe.elf --section-headers --elf-output-style=LLVM | FileCheck %s --match-full-lines
;
; CHECK-LABEL: Name: .rel.text ({{[0-9]+}})
; CHECK:       Offset: 0x[[#%X,OFFSET1:]]
; CHECK-LABEL: Name: .rodata.cst32 ({{[0-9]+}})
; CHECK:       Offset: 0x[[#OFFSET1 + 65]]
; CHECK-LABEL: Name: .note.GNU-stack ({{[0-9]+}})
```

kuhar (changes requested):
Hi @bsaleil,

Thanks for adding the test. I think it's tricky to write a concise test given the tools that we have in available lgc.

I added suggestion that should make the test easier to work with should we want to update it in the future. A few key things:
-  We should prefer to use llvm-* tools over binutils. This is because system-provided binaries might have slight differences in the way they define flags and output format. If we stick to llvm-provided, we are guaranteed to get the same behavior on all systems used to build LLPC.
-  FileCheck supports arithmetic checks: https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-numeric-substitution-blocks. I think using those is cleaner as it does not require invoking subshells and making sure all parameters are in the accepted format. It should be also easier to debug as FileCheck prints the surrounding context and possible intended matches, in case things stop working in the future.

One thing that @s-perron suggested to me would be to check if the test could be made more general but not assuming the section order. I did not know enough about it to tell, but maybe you can check. If this is hard, I think the test should be fine as long as we confirm that the order is stable across runs.

lgc/test/ElfRelocationSize.lgc
```diff
@@ -0,0 +1,220 @@
+; RUN: lgc -mcpu=gfx1030 -extract=2 -o %t.fs.elf %s
+; RUN: lgc -mcpu=gfx1030 -extract=3 -other=%t.fs.elf -o %t.vs.elf %s
+; RUN: lgc -mcpu=gfx1030 -extract=1 -l %s -o %t.pipe.elf %t.vs.elf %t.fs.elf
+; RUN: llvm-readelf -W -S %t.pipe.elf | grep ".rel.text\|.rodata.cst32" | FileCheck %s
+
+; This test checks that no random extra bytes are generated after the relocations
+; in the relocation section (.rel.text). We check that we have the expected
+; number of bytes between this section and the one following it (.rodata.cst32).
+; Here we write 4 16-bit relocations, so we expect the following section to be at
+; OFFS+0x40 (or a bit more because of section alignment).
```
s-perron, line 10: Did you mean "byte" instead of "bit"?

s-perron (approved):

lgc/test/ElfRelocationSize.lgc
bsaleil: @s-perron sorry, I pushed again to use @kuhar suggested solution instead.

s-perron (approved):

kuhar (approved):

trenouf (approved):
LGTM. Ping me for a merge.

PR #1672, opened 2022-01-29
Derive array/matrix stride when it is missing but is requested

amdrexu:
This is for future use. When array/matrix stride is missing, we can try
to derive it. This is because certain categories of blocks could meet
such conditions. They don't specify ArrayStride and MatrixStride but
expect the data layout is decided by compiler.

Change-Id: I7824a622a3ca0adeb1da78c9d538cf7f4dd43430

### DIFF
--- a/llpc/translator/lib/SPIRV/libSPIRV/SPIRVType.cpp
+++ b/llpc/translator/lib/SPIRV/libSPIRV/SPIRVType.cpp
@@ -63,6 +63,11 @@ uint64_t SPIRVType::getArrayLength() const {
       ->getZExtIntValue();
 }
 
+unsigned SPIRVType::getDerivedArrayStride() const {
+  assert((OpCode == OpTypeArray || OpCode == OpTypeRuntimeArray) && "Not array type");
+  return getArrayElementType()->getSizeInBytes();
+}
+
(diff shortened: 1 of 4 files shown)


llpc/translator/lib/SPIRV/libSPIRV/SPIRVType.cpp
```diff
@@ -86,6 +91,32 @@ SPIRVWord SPIRVType::getIntegerBitWidth() const {
   return static_cast<const SPIRVTypeInt *const>(this)->getBitWidth();
 }
 
+unsigned SPIRVType::getSizeInBytes() const {
+  if (isTypeVector())
+    return getVectorComponentType()->getSizeInBytes() * getVectorComponentCount();
+  if (isTypeMatrix())
+    return getMatrixColumnType()->getSizeInBytes() * getMatrixColumnCount();
```
linqun, line 98: it is  better to  add a comment about no alignment is expected.

linqun (approved):

llpc/translator/lib/SPIRV/libSPIRV/SPIRVType.cpp
amdrexu: Will add it.

PR #1683, opened 2022-02-09
Fix OpPtrEqual/OpPtrNotEqual

AMD-dwang:
The two operands of OpPtrEqual/OpPtrNotEqual have the same SPIRV type,
but the IR type could be different. When type is different, we return
false for OpPtrEqual, true for OpPtrNotEqual.


amdvlk-admin:

PR #1684, opened 2022-02-09
Fix the issue on opudotaccsatkhr

reallyzhao:
This issue has been found during the vulkan cts on VK.spirv_assembly.instruction.compute.opudotaccsatkhr.*_out32
It's due to the incorrect using on llvm.add.sat intrinsic, it should specify whether the operator is signed.


amdvlk-admin:

kuhar (commented):
Thanks for the fix.

1. Could you add a shaderdb test to make sure that each kind of additions intrinsics is emitted as expected? This is so that we don't regress in the future when we modify the code, etc. See the [contribution guidelines](https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/Contributing.md#write-useful-tests).

2. Could you rebase the changes and make sure these exactly one commit?

PR #1696, opened 2022-02-15
Fix missing merged Hs-Ls shader operations in NGG mode

crystalJinamd:
There are tessellation issues caused by a regression point 03d9e6cb.


amdvlk-admin:

JaxLinAMD:
ok to test

crystalJinamd:
> 

Yes, that is what I mean. The merged Hs-Ls shader in NGG mode also is needed to be handled.

PR #1703, opened 2022-02-18
Add build option in amdllpc to link spvgen statically or dynamically

AMD-dwang:
Add an option 'USING_STATIC_SPVGEN' to handle amdllpc to link spvgen
dynamically or statically.

### DIFF
--- a/imported/spirv/spvgen.h
+++ b/imported/spirv/spvgen.h
@@ -240,11 +240,9 @@ void SH_IMPORT_EXPORT vfxPrintDoc(
 }
 #endif
 
-static inline bool InitSpvGen(
-    const char* pSpvGenDir = nullptr)
-{
-    return true;
-}
+bool InitSpvGen(const char* pSpvGenDir = nullptr);
+
+void FinalizeSpvgen();
 
 #else
 
@@ -390,6 +388,8 @@ DECL_EXPORT_FUNC(vfxGetPipelineDoc);
 DECL_EXPORT_FUNC(vfxPrintDoc);
 
 bool SPVAPI InitSpvGen(const char* pSpvGenDir = nullptr);
+
+void SPVAPI FinalizeSpvgen();
 
 #endif
 
@@ -552,6 +552,10 @@ bool SPVAPI InitSpvGen(
     }
     return success;
 }
+
+// =====================================================================================================================
+// Finalize spvgen; Clean up resource
+void SPVAPI FinalizeSpvgen() {}
 
 #endif
 
--- a/llpc/CMakeLists.txt
+++ b/llpc/CMakeLists.txt
@@ -397,6 +397,8 @@ add_executable(amdllpc tool/amdllpc.cpp)
 add_dependencies(amdllpc llpc_standalone_compiler)
 target_link_libraries(amdllpc PRIVATE llpc_standalone_compiler)
 set_compiler_options(amdllpc ${LLPC_ENABLE_WERROR})
+add_compile_definitions(amdllpc PRIVATE SH_EXPORTING)
+target_link_libraries(amdllpc PRIVATE spvgen_static)
 
 endif()
 ### Add Subdirectories #################################################################################################
--- a/llpc/tool/amdllpc.cpp
+++ b/llpc/tool/amdllpc.cpp
@@ -542,6 +542,8 @@ int main(int argc, char *argv[]) {
 
   // Cleanup code that gets run automatically before returning.
   auto onExit = make_scope_exit([compiler, &result] {
+    FinalizeSpvgen();
+
     if (compiler)
       compiler->Destroy();
 
--- a/llpc/tool/llpcCompilationUtils.cpp
+++ b/llpc/tool/llpcCompilationUtils.cpp
@@ -75,10 +75,6 @@
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/raw_ostream.h"
 
-#ifndef LLPC_ENABLE_SPIRV_OPT
-#define SPVGEN_STATIC_LIB 1
-#endif
-
 #include "spvgen.h"
 #include "vfx.h"
 #include <cassert>
--- a/llpc/context/llpcCompiler.cpp
+++ b/llpc/context/llpcCompiler.cpp
@@ -74,7 +74,6 @@
 #include <unordered_set>
 
 #ifdef LLPC_ENABLE_SPIRV_OPT
-#define SPVGEN_STATIC_LIB 1
 #include "spvgen.h"
 #endif
 
--- a/llpc/unittests/vfx/testVfxParser.cpp
+++ b/llpc/unittests/vfx/testVfxParser.cpp
@@ -30,9 +30,6 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
-#define SPVGEN_STATIC_LIB 1
-#include "spvgen.h"
-
 using namespace llvm;
 using ::testing::ElementsAre;
 using ::testing::ElementsAreArray;
--- a/tool/vfx/CMakeLists.txt
+++ b/tool/vfx/CMakeLists.txt
@@ -51,7 +51,7 @@ target_sources(vfx PRIVATE
     vfxVkSection.cpp
 )
 
-target_compile_definitions(vfx PRIVATE VFX_SUPPORT_VK_PIPELINE)
+target_compile_definitions(vfx PRIVATE VFX_SUPPORT_VK_PIPELINE SH_EXPORTING)
 
 target_include_directories(vfx
 PUBLIC
@@ -61,5 +61,6 @@ PRIVATE
     ${PROJECT_SOURCE_DIR}/../../include
 )
 
+target_link_libraries(vfx PRIVATE spvgen_static)
 target_link_libraries(vfx PRIVATE khronos_vulkan_interface)
 target_link_libraries(vfx PRIVATE khronos_spirv_interface)


imported/spirv/spvgen.h
```diff
@@ -240,11 +240,9 @@ void SH_IMPORT_EXPORT vfxPrintDoc(
 }
 #endif
 
-static inline bool InitSpvGen(
-    const char* pSpvGenDir = nullptr)
-{
-    return true;
-}
+bool InitSpvGen(const char* pSpvGenDir = nullptr);
```
kuhar, line 243: This is not related to this PR per se, but why is the spvgen header duplicated in LLPC?

llpc/CMakeLists.txt
```diff
@@ -391,6 +393,13 @@ add_executable(amdllpc tool/amdllpc.cpp)
 add_dependencies(amdllpc llpc_standalone_compiler)
 target_link_libraries(amdllpc PRIVATE llpc_standalone_compiler)
 set_compiler_options(amdllpc ${LLPC_ENABLE_WERROR})
+#if USING_STATIC_SPVGEN
```
kuhar, line 396: nit: remove the comments?

llpc/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ project(LLPC C CXX)
 ### Create LLPC Library ################################################################################################
 add_library(llpc STATIC "")
 
+option(USING_STATIC_SPVGEN "Enable to link spvgen statically." ON)
```
kuhar, line 31: nit: This should be called something like `LLPC_USE_STATIC_SPVGEN`. This would be more consistent with the other cmake flags that sound like we are telling the build system what to do, e.g., "build something", "use something", "enable something", etc.

Also, can move it closer to the other LLPC options down below?

llpc/CMakeLists.txt
```diff
@@ -391,6 +393,13 @@ add_executable(amdllpc tool/amdllpc.cpp)
 add_dependencies(amdllpc llpc_standalone_compiler)
 target_link_libraries(amdllpc PRIVATE llpc_standalone_compiler)
 set_compiler_options(amdllpc ${LLPC_ENABLE_WERROR})
+#if USING_STATIC_SPVGEN
+if(USING_STATIC_SPVGEN)
+    target_link_libraries(amdllpc PRIVATE spvgen_static)
+    add_compile_definitions(amdllpc PRIVATE SH_EXPORTING)
+    add_compile_definitions(amdllpc PRIVATE USING_STATIC_SPVGEN)
```
kuhar, line 400: You can provide multiple definitions at once: https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

llpc/tool/amdllpc.cpp
```diff
@@ -577,6 +577,8 @@ int main(int argc, char *argv[]) {
     return EXIT_FAILURE;
   }
 
+  FinalizeSpvgen();
```
kuhar, line 580: Don't we have to finalize it in the failure case as well?

kuhar (changes requested):
How do you plan to handle the tests in llpc/llpc/test/shaderdb/error_reporting` that check that we print an error when a bad spvgen direcotry is provided or spvgen is not specified at all? Do we need to introduce a new `requires` lit attribute?

Also, is there a plan to eventually deprecate using spvgen as a shared library and remove the related code from amdllpc?

imported/spirv/spvgen.h
AMD-dwang: we didn't provide the spvgen directory before, so import this file.

llpc/CMakeLists.txt
AMD-dwang: Fixed. 
I wanted to use the same macro for other tools, so I didn't add `LLPC_`.

llpc/tool/amdllpc.cpp
AMD-dwang: Fixed, thanks.

llpc/CMakeLists.txt
AMD-dwang: > Also, is there a plan to eventually deprecate using spvgen as a shared library and remove the related code from amdllpc?

If there is nothing wrong with the static library, I will add other changes to deprecate shared library. 

> How do you plan to handle the tests in `llpc/llpc/test/shaderdb/error_reporting` that check that we print an error when a bad spvgen direcotry is provided or spvgen is not specified at all? Do we need to introduce a new `requires` lit attribute?

We may remove this directory.

imported/spirv/spvgen.h
kuhar: I don't follow. The spvgen file has been in spvgen for a very long time https://github.com/GPUOpen-Drivers/spvgen/commits/dev/include. Why do we need a second copy within llpc?

llpc/CMakeLists.txt
kuhar: If we want both the dynamic/shared and static compilation options to co-exist for some time, then I don't think we should just delete these test. Refactoring related to the static library may introduce unexpected breakage in the shared library.

llpc/CMakeLists.txt
kuhar: It's not a macro, it's a CMake option. It's cleanest when each project has its own options. For example, we have cmake code in XGL that propagated ICD options to LLPC. I'd prefer to follow what XGL does to keep things consistent.

imported/spirv/spvgen.h
AMD-dwang: Yes, this file is unnecessary. This file is historical issue, we have plans to remove it.

imported/spirv/spvgen.h
```diff
@@ -240,11 +240,9 @@ void SH_IMPORT_EXPORT vfxPrintDoc(
...
+
+void FinalizeSpvgen();
```
tsymalla, line 245: For consistency, this should be `FinalizeSpvGen`.

llpc/tool/llpcCompilationUtils.cpp
```diff
@@ -76,7 +76,9 @@
 #include "llvm/Support/raw_ostream.h"
 
 #ifndef LLPC_ENABLE_SPIRV_OPT
```
s-perron, line 78: What does spirv-opt have to do with this?

AMD-dwang: I didn't change it, what I have changed is below.
LLPC_ENABLE_SPIRV_OPT is not currently used.

s-perron: Even if you did not add that line you chose how to combine your change with that line.  Why did you not do

```
#if !defined(LLPC_ENABLE_SPIRV_OPT) || !defined(LLPC_ENABLE_SPIRV_OPT)
#define SPVGEN_STATIC_LIB
#endif
```

The names of the macros are very confusing.  You say if that if you are not using static spvgen, then you define spvgen_static_lib.  That seems backwards.

You might want to have a comment that explains the reason for making these decisions.  Also, if the macros need to change name, then we should do that.

AMD-dwang: > The names of the macros are very confusing. You say if that if you are not using static spvgen, then you define spvgen_static_lib. That seems backwards.
> 
Yes, I agree with you. this name is backwards, but it is defined in `spvgen`, other projects are using it, I can't change it.





> The other way of combining these two options is to have the cmake file define `LLPC_USE_STATIC_SPVGEN` if `LLPC_ENABLE_SPIRV_OPT` is defined.
> 
> > LLPC_ENABLE_SPIRV_OPT is not currently used.
> 
> As long as it is in the code, we should be thinking carefully how all of the different options should interact. We need to know the reason that SPVGEN_STATIC_LIB is set when LLPC_ENABLE_SPIRV_OPT is not defined. If it is because SPVGEN_STATIC_LIB must be set, then you may have broken something because it sometimes won't be set. If it is that when LLPC_ENABLE_SPIRV_OPT is set, then SPVGEN_STATIC_LIB cannot be set, then original code is wrong because SPVGEN_STATIC_LIB could be set somewhere else.

Yes, Here is also very confusing. My understanding is that we need `spvgen` when `LLPC_ENABLE_SPIRV_OPT' is defined, then we could select static or shared library; if  `LLPC_ENABLE_SPIRV_OPT' is undefined, we don't need `spvgen`, which means `spvgen.h` should be gated into macros.
However, I am worried that I've misunderstood, so I didn't make changes. I only make sure `amdllpc` works.

s-perron: Okay.

Flakebi:
Linking spvgen statically like in this PR fixes a double free.

Longer explanation:
With libc 2.34, most amdllpc tests start to fail because of a memory corruption caused by a double free. `vfxSection.cpp` defines a global variable `std::map<…> Section::m_sectionInfo` and is compiled as part of the `vfx` library. vfx is linked into amdllpc, so the constructor of `m_sectionInfo` is called when amdllpc starts and the destructor is called on exit.
`vfxSection.cpp` is also compiled into `spvgen` ([CMakeLists.txt](https://github.com/GPUOpen-Drivers/spvgen/blob/1b6b362984a2d99d897a7edd2e90cb230c933b87/CMakeLists.txt#L95-L117)) (why??), also defining `Section::m_sectionInfo` and calling the constructor when the library is loaded and the destructor when the library is unloaded.
When amdllpc starts, it creates the `m_sectionInfo` map. When it loads `spvgen.so`, it will re-initialize the map, using the same physical address, because `Section::m_sectionInfo` is a globally visible and unique symbol. On exit of amdllpc, `spvgen.so` is unloaded, freeing the map. Then amdllpc exits, freeing the map a second time.
I assume the double free has been there for a long time, but libc < 2.34 did not assert.

kuhar:
@Flakebi 
I'm not sure if you are looking for solutions to the double free problem, but a common way to handle this is IIRC to keep pointers to non-trivial globals, e.g., `Section global_sec;` -> `Section* global_sec = new Section();` or `Section& global_sec = *new Section;` and potentially doing a `delete` in some shutdown/deinit function. (https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables)

Flakebi:
This was merged as part of #1774.

PR #1719, opened 2022-03-02
Fix workgroupSize

AMD-dwang:
When workgroupSize is defined by ExecutionModeLocalSizeId, real size
should be got after translation.

### DIFF
--- a/llpc/test/shaderdb/core/OpExecutionModeId_TestLocalSizeId.spvasm
+++ b/llpc/test/shaderdb/core/OpExecutionModeId_TestLocalSizeId.spvasm
@@ -1,0 +1,101 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} final ELF info
+; The workgroup sizes will be set to the values in the LocalSizeId instruction.  In this case,
+; we should use the default values for the spec constants %7, %8, and %9.
+; SHADERTEST: COMPUTE_NUM_THREAD_X 0x0000000000000002
+; SHADERTEST: COMPUTE_NUM_THREAD_Y 0x0000000000000008
+; SHADERTEST: COMPUTE_NUM_THREAD_Z 0x0000000000000020
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
+
+
+; SPIR-V
+; Version: 1.4
+; Generator: Khronos Glslang Reference Front End; 10
+; Bound: 57
+; Schema: 0
+OpCapability Shader
+OpCapability GroupNonUniform
+%1 = OpExtInstImport "GLSL.std.450"
+OpMemoryModel Logical GLSL450
+OpEntryPoint GLCompute %4 "main" %14 %29 %43 %50
+OpExecutionModeId %4 LocalSizeId %7 %8 %9
+OpDecorate %7 SpecId 0
+OpDecorate %8 SpecId 1
+OpDecorate %9 SpecId 2
+OpDecorate %14 BuiltIn NumWorkgroups
+OpDecorate %16 SpecId 0
+OpDecorate %17 SpecId 1
+OpDecorate %18 SpecId 2
+OpDecorate %29 BuiltIn GlobalInvocationId
+OpDecorate %43 RelaxedPrecision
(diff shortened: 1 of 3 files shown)


AMD-dwang:
> Can you add a test for this? See the [contributing guidelines ](https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/Contributing.md#write-useful-tests) about tests.

This change don't update any instructions,  adding a test is useless. Thanks.

s-perron (changes requested):
Can you add a test for a case for this?  The coverage run shows that the new code is not executed.  A new test is needed:



https://storage.googleapis.com/amdvlk-llpc-github-ci-artifacts-public/coverage_release_clang_coverage_1920555000/coverage/vulkandriver/drivers/llpc/llpc/translator/lib/SPIRV/SPIRVReader.cpp.html#L6885

s-perron:
> > Can you add a test for this? See the [contributing guidelines ](https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/Contributing.md#write-useful-tests) about tests.
> 
> This change don't update any instructions, adding a test is useless. Thanks.

It should change the pal metadata, which can be checked in a shaderdb test.  See https://github.com/GPUOpen-Drivers/llpc/blob/f439d6b79b8cd678fc8804bd0020eaec35534e48/llpc/test/shaderdb/relocatable_shaders/PipelineVsFs_FillPsInput.pipe.  This test check the pal metadata.

AMD-dwang:
> > > Can you add a test for this? See the [contributing guidelines ](https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/Contributing.md#write-useful-tests) about tests.
> > 
> > 
> > This change don't update any instructions, adding a test is useless. Thanks.
> 
> It should change the pal metadata, which can be checked in a shaderdb test. See https://github.com/GPUOpen-Drivers/llpc/blob/f439d6b79b8cd678fc8804bd0020eaec35534e48/llpc/test/shaderdb/relocatable_shaders/PipelineVsFs_FillPsInput.pipe. This test check the pal metadata.

I added a test, but this test don't catch this PR. The workgroupSize should come from VkPipelineShaderStageCreateInfo ::VkSpecializationInfo.

AMD-dwang:
retest this please

AMD-dwang:
@s-perron I have added a test. any comments?  I don't know why the checks can't be verified, but it was okay before. Thanks.

llpc/test/shaderdb/core/OpExecutionModeId_TestLocalSizeId.spvasm
```diff
@@ -0,0 +1,99 @@
...
+OpMemberDecorate %48 0 Offset 0
+OpDecorate %48 Block
+OpDecorate %50 DescriptorSet 0
+OpDecorate %50 Binding 0
+%2 = OpTypeVoid
+%3 = OpTypeFunction %2
+%6 = OpTypeInt 32 0
+%7 = OpSpecConstant %6 2
+%8 = OpSpecConstant %6 8
+%9 = OpSpecConstant %6 8
```
s-perron, line 43: Could you make the values for %8 and %9 different to make it easier to know that the value in the pal metadata is being taken from the correct place?

llpc/test/shaderdb/core/OpExecutionModeId_TestLocalSizeId.spvasm
```diff
@@ -0,0 +1,99 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -spvgen-dir=%spvgendir% -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} final ELF info
+; SHADERTEST: COMPUTE_NUM_THREAD_X 0x0000000000000002
```
s-perron, line 4: Add a comment explaining where these values come from.  Something like

```
The workgroup sizes will be set to the values in the LocalSizeId instruction.  In this case, we should use the default values for the spec constants %7, %8, and %9.
```

s-perron (approved):
LGTM.  Just a couple suggestions.

llpc/test/shaderdb/core/OpExecutionModeId_TestLocalSizeId.spvasm
AMD-dwang: Updated it. Thanks.

amdrexu (approved):

PR #1741, opened 2022-03-15
Fix OpImageQueryLod

AMD-dwang:
1. 'coarse' calculation is not precise enough,modify derivative to 'fine' calculation.
2. Refine the test.


amdrexu (approved):
Not sure why 2DShadow and CubeArray are removed from GLSL source. Others look good.

AMD-dwang:
> Not sure why 2DShadow and CubeArray are removed from GLSL source. Others look good.

1D, 2D, 3D are listed one each,  the instructions are the same, remove them to avoid  duplicating. Thanks.

PR #1780, opened 2022-04-21
Add fence between merged shader stages

LLJJDD:
Merged shaders use LDS to pass value, if an s_wait is not performed,
value may be accessed before it is finished written.

### DIFF
--- a/lgc/patch/ShaderMerger.cpp
+++ b/lgc/patch/ShaderMerger.cpp
@@ -192,6 +192,7 @@ Function *ShaderMerger::generateLsHsEntryPoint(Function *lsEntryPoint, Function 
   //     br label %.endLs
   //
   // .endLs:
+  //     fence
   //     call void @llvm.amdgcn.s.barrier()
   //     %hsEnable = icmp ult i32 %threadId, %hsVertCount
   //     br i1 %hsEnable, label %.beginHs, label %.endHs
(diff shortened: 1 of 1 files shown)


lgc/patch/ShaderMerger.cpp
```diff
@@ -371,6 +372,7 @@ Function *ShaderMerger::generateLsHsEntryPoint(Function *lsEntryPoint, Function
   args.clear();
   attribs.clear();
   attribs.push_back(Attribute::NoRecurse);
+  new FenceInst(*m_context, AtomicOrdering::Release, SyncScope::System, endLsBlock);
```
amdrexu, line 375: I think we should send the diff dump to backend. Theoretically, s_barrier (see the next statement) is safe enough to make sure the LDS operations are finished.

amdvlk-admin:
### Test summary for commit 3baa3a288f1dcd985456c0c55dcf781fc37ae2f8

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/ShaderMerger.cpp
nhaehnle: Jay made me aware of the issue, and I think this may not be true. I think what can happen is that ds_store issues in SIMD0 but is stuck in arbitration, e.g. by claused memory instructions from SIMD1. The s_barrier executes, releases a different wave on SIMD1 which executes ds_load, and then the ds_load gets lucky and gets arbitrated before the ds_store from SIMD0.

amdrexu: I think we have some codes that call s_barrier without adding explicit fence. According to the LLVM's change, we have to review those points and might apply the changes? The cases are: HS, NGG, Merged shader sync between shader pair, and mesh shader.

amdrexu:
LLVM backend will revert the causing change.

jayfoad:
The LLVM backend change https://reviews.llvm.org/D120544 was reverted, but the author has said that they would like to reapply it. So we may still need to add explicit fences in LLPC to keep everything working.

amdrexu:
> The LLVM backend change https://reviews.llvm.org/D120544 was reverted, but the author has said that they would like to reapply it. So we may still need to add explicit fences in LLPC to keep everything working.

Thank you for notification. Could you file an internal ticket and assign it to me? I think we need to revisit all codes that need such synchronization apart from merged shader.

PR #1797, opened 2022-05-01
Fix a latent issue of NGG GS

amdrexu:
We should only handle GS message when mutating GS. Other messages from
s_sendmsg should be skipped.


nhaehnle (approved):
LGTM

PR #1802, opened 2022-05-06
Replace some getPointerElementType in SPIRVReader

Flakebi:
Use alternative means like the SPIRVType to get the type behind
pointers.

Also replace uses of getContainedType(0), which is a
getPointerElementType in disguise.

SPIRVReader still has more getPointerElementType uses that are more
difficult to replace, but I wanted to put this up so the work is not
duplicated.

cc @mariusz-sikora-at-amd


amdrexu (approved):
LGTM

PR #1837, opened 2022-06-06
Adding forceCsThreadidSwizzling for shader tuning

ruiminzhao:
This option will be set under vulkan optimization.
Originally SIMDs cover 16x2 regions, if open the switch, swizzling they cover 8x4 regions.
Default value for this option is false.

### DIFF
--- a/lgc/patch/PatchInOutImportExport.cpp
+++ b/lgc/patch/PatchInOutImportExport.cpp
@@ -478,6 +478,19 @@ void PatchInOutImportExport::processShader() {
           reconfigCall->eraseFromParent();
         }
       }
+      // Different with above, this will force the threadID swizzle which will rearrange thread ID within a group into
+      // blocks of 8*4, not to reconfig workgroup automatically and will support to be swizzled in 8*4 block
+      // split.
+      if (func.isDeclaration() && func.getName().startswith(lgcName::SwizzleLocalInvocationId)) {
+        while (!func.use_empty()) {
+          CallInst *swizzleCall = cast<CallInst>(*func.user_begin());
+          Value *localInvocationId = swizzleCall->getArgOperand(0);
+
+          localInvocationId = swizzleLocalInvocationIdIn8x4(localInvocationId, swizzleCall);
+          swizzleCall->replaceAllUsesWith(localInvocationId);
+          swizzleCall->eraseFromParent();
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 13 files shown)


amdvlk-admin:

lgc/patch/PatchInOutImportExport.cpp
```diff
@@ -5436,6 +5449,65 @@ Value *PatchInOutImportExport::reconfigWorkgroup(Value *localInvocationId, Instr
...
+
+  if ((workgroupSizeX >= 16) && (workgroupSizeX % 8 == 0) && (workgroupSizeY % 4 == 0)) {
+    Value *threadIdInGroupX = builder.CreateExtractElement(localInvocationId, (uint64_t)0);
+    Value *threadIdInGroupY = builder.CreateExtractElement(localInvocationId, 1);
+    Value *threadIdInGroupZ = builder.CreateExtractElement(localInvocationId, 2);
+    const unsigned blockSizeX = 8;
+    const unsigned blockSizeY = 4;
+    const unsigned blockDimY = workgroupSizeY / 4; // How many blocks are splitted in Y level.
+
+    unsigned blockSize = blockSizeX * blockSizeY; // 32 threads per-block
```
kuhar, line 5482: Since other variables are `const`, can we mark `blockSize` as well? I don't see any reassignments below.

kuhar (commented):
Can you please add a shaderdb test to check that the generated code looks OK? See https://github.com/GPUOpen-Drivers/llpc/blob/dev/docs/Contributing.md#write-useful-tests

llpc/test/shaderdb/core/LocalInvocationIdNotSwizzle.spvasm
```diff
@@ -0,0 +1,102 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} pipeline before-patching results
+; SHADERTEST-NOT: lgc.swizzle.local.invocation.id
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
```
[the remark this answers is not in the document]
ruiminzhao: Great! I would like to merge two following your suggestion.

llpc/test/shaderdb/core/LocalInvocationIdSwizzle.spvasm
```diff
@@ -0,0 +1,108 @@
+; BEGIN_SHADERTEST
+; RUN: amdllpc -v %gfxip %s --force-compute-shader-thread-id-swizzling | FileCheck -check-prefix=SHADERTEST %s
+; SHADERTEST-LABEL: {{^// LLPC}} pipeline before-patching results
+; SHADERTEST: %{{[0-9]+}} = call <3 x i32> @lgc.swizzle.local.invocation.id(<3 x i32> %{{[0-9]+}})
+; SHADERTEST: AMDLLPC SUCCESS
+; END_SHADERTEST
+; BEGIN_SHADERTEST
+; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
```
kuhar, line 8: The two cases need to use different check prefix instead of SHADERTEST twice.

SHADERTEST has no special meaning, it's just a convention in llpc to use it when there's a single RUN line. See  https://llvm.org/docs/CommandGuide/FileCheck.html#the-filecheck-check-prefix-option.

ruiminzhao: Thanks for your correction. I will update the check-prefix.

kuhar (approved):

amdrexu (approved):

PR #1838, opened 2022-06-06
Disable opaque pointer support

dstutt:
LLVM is moving to opaque pointers as the default.
Temporarily disable this until opaque pointer support is enabled in llpc and
lgc.


trenouf (approved):
LGTM

PR #1853, opened 2022-06-20
lgc/PassManager: use report_fatal_error for bad command-line options

nhaehnle:
llvm_unreachable() is like assert() in that it should only be used to
flag conditions that are programmer error. Depending on the build
options, the compiler may actually optimize code based on the assumption
that it is never reached, and so paths that would reach it become
undefined behavior.

The arguably correct thing to do here is a report_fatal_error(), because
the condition can only be reached via bad debug settings.

### DIFF
--- a/lgc/util/PassManager.cpp
+++ b/lgc/util/PassManager.cpp
@@ -172,7 +172,7 @@ PassManagerImpl::PassManagerImpl(TargetMachine *targetMachine)
       m_instrumentationStandard(cl::DebugPassManager, cl::DebugPassManager || cl::VerifyIr,
                                 /*PrintPassOpts=*/{true, false, true}) {
   if (!cl::DumpCfgAfter.empty())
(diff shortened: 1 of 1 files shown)


Flakebi (approved):

kuhar (approved):
LGTM

Probably also worth noting, the exact behavior is controlled by the `LLVM_UNREACHABLE_OPTIMIZE` cmake flag. When enabled, `llvm_unreachable` behaves like `assert(false); assume(false)`, and when disabled like just `assert(false)`. We use the default, i.e., `=1`. With that, I suspect that there might be places in the codebase where we use `llvm_unreachable` but probably don't want to actually introduce an assumption based on Vulkan API inputs, e.g.:

```c++
if (!inputPtr) {
  llvm_unreachable();
  return Result::Xyz;
}
...
```

Has anyone looked into this or found similar issues?

bsaleil (approved):
LGTM

amdrexu (approved):

PR #1854, opened 2022-06-21
Add optimization option to swizzle thread group

LLJJDD:
This feature generates new thread group ID based on system thread group
ID for user. The algorithm is generating morton code from the flat
system thread group ID. So that the order of new ID would look like
z-order curve.

PR #1855, opened 2022-06-21
Update typo finder CI

Flakebi:
The different cases for LOD are not needed (anymore?) and ba needs to be
a word to fix false positives in #1854.


dstutt:
Why aren't the different LOD cases needed anymore?

Flakebi:
>>    Why aren't the different LOD cases needed anymore?
>
> Aren't these specs case insensitive anyway (or was that your point)?

Not sure if that changed, but they seem to be case insensitive now (which is why I removed them).

dstutt (approved):
LGTM

PR #1878, opened 2022-07-10
CMakeLists: prefer to use llvm_map_components_to_libnames

nhaehnle:
This fixes the build of lgc and lgcdis when LLVM is compiled as a
shared library.

The underlying issue is that the lgc and lgcdis *libraries* are not
linked into the libLLVM-*.so, but the LLVM_LINK_COMPONENTS mechanism
assumes that they are.


kuhar (approved):

dstutt (approved):
LGTM

PR #1895, opened 2022-07-18
Revoke GitHub access

WenqingLiAMD:


jinjianrong (approved):

Flakebi:
I’m not sure why this was done, but if you remove kuhar for some reason, it probably makes sense to remove s-perron as well.

PR #1900, opened 2022-07-19
Fix crash when dumping ELF with empty map entry

LLJJDD:
Recently PAL has a change on pipeline ABI, and our internal compiler
may generate ELF with an empty map entry.

This commit deal with that case by skipping the empty map.

PR #1901, opened 2022-07-19
Handle empty map types in MessagePack iterator.

perlfu:
MessagePack reader cannot assume the next node after MapBegin is a MapPair.


perlfu:
This is a duplicate of #1900

PR #1912, opened 2022-07-22
Remove PatchBufferOp handling for array stores

jayfoad:
PatchBufferOp has special handling for stores of array type, but it
seems unsound because it assumes that the array element type and
smallestType are the same size. Apparently it is never used anyway so
remove it.


nhaehnle (approved):
A potential alternative would be to do the array handling and then fall-through to the bitcast, but array values are a bit strange anyway. So this seems reasonable to me.

jayfoad:
> Apparently it is never used anyway so remove it.

Apparently I was wrong about that. It causes CTS failures.

amdrexu (approved):

PR #1924, opened 2022-07-27
Add mesh shader support: Correct the stage mask for ViewIndex

amdrexu:
ViewIndex could be used in mesh shader. The original stage mask is
incorrect.


nhaehnle (approved):
LGTM

PR #1931, opened 2022-07-28
Fix two issues of SystemValues for mesh shader

amdrexu:
- When getting global table or per-shader table, we have to count in
  additional 8 special SGPRs for mesh shader. That is to say the s8/s9
  are expected value rather than s0/s1.
- Draw data ring is only accessed by task shader. Tighten the assert
  condition.

### DIFF
--- a/lgc/patch/SystemValues.cpp
+++ b/lgc/patch/SystemValues.cpp
@@ -29,6 +29,7 @@
  ***********************************************************************************************************************
  */
 #include "lgc/patch/SystemValues.h"
+#include "ShaderMerger.h"
 #include "lgc/state/PipelineState.h"
 #include "lgc/state/TargetInfo.h"
 #include "llvm/IR/Constants.h"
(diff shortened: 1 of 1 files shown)


lgc/patch/MeshTaskShader.cpp
```diff
@@ -168,7 +168,8 @@ void MeshTaskShader::processMeshShader(Function *entryPoint) {
 // @param readTy : Type of value to read
 // @param byteOffset : Byte offset within the payload entry
 Value *MeshTaskShader::readTaskPayload(Function *entryPoint, Type *readTy, Value *byteOffset) {
-  auto payloadRingBufDesc = m_pipelineSysValues.get(entryPoint)->getTaskPayloadRingBufDesc();
+  auto payloadRingBufDesc =
+      m_pipelineSysValues.get(entryPoint, getShaderStage(entryPoint) == ShaderStageMesh)->getTaskPayloadRingBufDesc();
```
trenouf, line 172: Could we move this condition inside ShaderSystemValues? Then we don't need the interface change on it.

amdrexu: Ah. This is a good question. My initial change was like what you expected:  check mesh stage and keep the interface unchanged. But later, I realized we might need this in NggPrimShader. In that code place, you can not tell whether a stage marked as "Geometry" is for a separate stage or is actually a merged shader. So I add a new explicit parameter. Of course, I could get rid of this consideration and follow your suggestion at present. We can further discuss the interface when there is an example after NggPrimShader rework.

amdvlk-admin:
### Test summary for commit 0be8a16e681b015d675f1c4956bd49d6bb08b20b
<details><summary>CTS tests (Failed: 0/216059)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 35298/73410 (48.1%)</li>
<li>Failed: 0/73410 (0.0%)</li>
<li>Not Supported: 38112/73410 (51.9%)</li><li>Warnings: 0/73410 (0.0%)</li>
</ul></details>

<li>Passed: 33017/69239 (47.7%)</li>
<li>Failed: 0/69239 (0.0%)</li>
<li>Not Supported: 36222/69239 (52.3%)</li><li>Warnings: 0/69239 (0.0%)</li>
</ul></details>

<li>Passed: 37553/73410 (51.2%)</li>
<li>Failed: 0/73410 (0.0%)</li>
<li>Not Supported: 35857/73410 (48.8%)</li><li>Warnings: 0/73410 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
LGTM

PR #1957, opened 2022-08-09
Opaque pointers: Changes to PatchPeepholeOpt

mariusz-sikora-at-amd:
Since we are not able to get pointer element type from IntToPtr
instruction, we have to check users to find base type.
Right now in all tests only load and getelementptr are used as users and
we are able to get base type from these instructions.

### DIFF
--- a/lgc/patch/PatchPeepholeOpt.cpp
+++ b/lgc/patch/PatchPeepholeOpt.cpp
@@ -132,9 +132,32 @@ void PatchPeepholeOpt::visitIntToPtr(IntToPtrInst &intToPtr) {
   if (!constOffset)
     return;
 
+  if (intToPtr.user_empty())
+    return;
+
+  auto *const user = cast<Instruction>(intToPtr.user_back());
+
+  const unsigned userOpcode = user->getOpcode();
+  // Check if user is Load or GEP instruction (right now only these two instructions are used).
+  if (userOpcode != Instruction::Load && userOpcode != Instruction::GetElementPtr)
+    return;
+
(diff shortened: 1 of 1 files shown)


qiaojbao:
retest this please

qiaojbao:
retest this please

amdrexu (approved):

lgc/patch/PatchPeepholeOpt.cpp
```diff
@@ -132,9 +132,32 @@ void PatchPeepholeOpt::visitIntToPtr(IntToPtrInst &intToPtr) {
...
+  Type *elementType = nullptr;
+  if (auto loadInst = dyn_cast<LoadInst>(user)) {
+    elementType = loadInst->getType();
+
+  } else if (auto getElemPtr = dyn_cast<GetElementPtrInst>(user)) {
+    elementType = getElemPtr->getSourceElementType();
+
+  } else {
+    llvm_unreachable("Should never be called!");
+    return;
```
nhaehnle, line 154: You could have simplified this by removing the opcode check above.

Note that, under the hood, the `dyn_cast`s are really the exact same opcode comparisons, so doing the opcode check in advance has no performance benefit.

(This is why LLVM has the `dyn_cast` infrastructure which is used instead of C++'s native dynamic_cast.)

mariusz-sikora-at-amd: I thought it will be more readable to have exit function check in line 142-143.

I will push new PR for the change that you are proposing (since this is merged)

PR #1986, opened 2022-08-24
Add optimization option to reverse thread group

LLJJDD:
This commit add optimization option to reverse thread group.

When the option is set, LLPC will generate code to reverse thread group
order by a driver given control value. Driver is responsible to fill the
internal buffer with either 0 or 1, and LLPC will load that value to
determine whether gl_WorkGroupId needs to be reversed.

It is expected that performance can improve for some compute shaders
that will access the same resource between each dispatch. By reversing
the order, the thread groups will begin accessing resource data that is
still in cache due to the earlier dispatch.


amdvlk-admin:
### Test summary for commit 2b24260579f7fba859c9ea7670ce03b74255cd88
<details><summary>CTS tests (Failed: 0/289625)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Rhel 8.2, Gfx9</summary><ul>

<li>Passed: 32868/69268 (47.5%)</li>
<li>Failed: 0/69268 (0.0%)</li>
<li>Not Supported: 36400/69268 (52.5%)</li><li>Warnings: 0/69268 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 34348/73361 (46.8%)</li>
<li>Failed: 0/73361 (0.0%)</li>
<li>Not Supported: 39013/73361 (53.2%)</li><li>Warnings: 0/73361 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerReverseThreadGroup.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// =====================================================================================================================
+// Executes this SPIR-V lowering pass on the specified LLVM module.
+//
+// @param [in/out] module : LLVM module to be run on (empty on entry)
+// @param [in/out] analysisManager : Analysis manager to use for this transformation
+PreservedAnalyses SpirvLowerReverseThreadGroup::run(Module &module, ModuleAnalysisManager &analysisManager) {
+  runImpl(module);
+  return PreservedAnalyses::none();
```
Flakebi, line 63: If `runImpl` returns false because nothing was changed, this function can return `PreservedAnalysis::all()`.

llpc/lower/llpcSpirvLowerReverseThreadGroup.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  unsigned builtIn = cast<ConstantInt>(callInst.getOperand(0))->getZExtValue();
+
+  if (builtIn == lgc::BuiltInWorkgroupId) {
+    assert(!m_workgroupId);
+    m_workgroupId = &callInst;
+  } else if (builtIn == lgc::BuiltInGlobalInvocationId) {
+    assert(!m_globalInvocationId);
+    m_globalInvocationId = &callInst;
+  }
```
Flakebi, line 110: Can these intrinsics only get called once?

nhaehnle (commented):
I believe this should be a change in LGC instead of LLPC. (cc @trenouf)

The intended overall design is that LLPC is the frontend that concerns itself with SPIR-V-specifics, while features that are common across APIs and shader languages should be in the shared LGC middle-end. As far as I can see, this feature squarely fits into the "LGC" bucket.

trenouf:
> I believe this should be a change in LGC instead of LLPC. (cc @trenouf)
> 
> The intended overall design is that LLPC is the frontend that concerns itself with SPIR-V-specifics, while features that are common across APIs and shader languages should be in the shared LGC middle-end. As far as I can see, this feature squarely fits into the "LGC" bucket.

I agree. But we don't want those hard-wired Vkgc internal set and binding constants inside LGC, so instead the LLPC front-end would have to pass them in to LGC.

Also, maybe I misunderstand, but it seems like this could be implemented on-the-fly as the built-in inputs are lowered in InOutBuilder, rather than as a whole separate pass.

s-perron (commented):
Please add tests.  Every time there is a change, at least one test should be added.  In this case, this is a new feature that might have multiple cases to consider, so I would expect multiple tests:

1. One test for each builtin where the builtin is called once.
2. One test for each builtin where the builtin is called multiple times.  This might need multiple tests if you are relying on commoning.  For example, a case where one call dominates the other and a test where it does not.
3. One test where both builtins are called.

LLJJDD:
> > I believe this should be a change in LGC instead of LLPC. (cc @trenouf)
> > The intended overall design is that LLPC is the frontend that concerns itself with SPIR-V-specifics, while features that are common across APIs and shader languages should be in the shared LGC middle-end. As far as I can see, this feature squarely fits into the "LGC" bucket.
> 
> I agree. But we don't want those hard-wired Vkgc internal set and binding constants inside LGC, so instead the LLPC front-end would have to pass them in to LGC.
> 
> Also, maybe I misunderstand, but it seems like this could be implemented on-the-fly as the built-in inputs are lowered in InOutBuilder, rather than as a whole separate pass.

Thanks! My initial idea is that this feature requires a driver specific internal buffer, so I put it in LLPC for better communication with driver. It should be better to be in LGC.

@trenouf , any idea how we could properly pass the internal buffer information to LGC? e.g., a buffer descriptor loaded as a value as an argument of the @lgc.create.read.builtin.input.* call, or some form of metadata?

LLJJDD:
> Please add tests. Every time there is a change, at least one test should be added. In this case, this is a new feature that might have multiple cases to consider, so I would expect multiple tests:
> 
> 1. One test for each builtin where the builtin is called once.
> 2. One test for each builtin where the builtin is called multiple times.  This might need multiple tests if you are relying on commoning.  For example, a case where one call dominates the other and a test where it does not.
> 3. One test where both builtins are called.

Thanks for reminding, will do in next iteration.

llpc/lower/llpcSpirvLowerReverseThreadGroup.cpp
LLJJDD: I believe SpirvLowerGlobal::mapInputToProxy() will turn all uses of the builtins into proxies, and the @lgc.create.read.builtin.input.* call will only appear once when initializing one proxy.

trenouf:
> @trenouf , any idea how we could properly pass the internal buffer information to LGC? e.g., a buffer descriptor loaded as a value as an argument of the @lgc.create.read.builtin.input.* call, or some form of metadata?

How about new LGC option(s) giving the set and descriptor id of the special buffer for this? Then, when the option(s) are not set, it does not do the optimization.

LLJJDD:
> > @trenouf , any idea how we could properly pass the internal buffer information to LGC? e.g., a buffer descriptor loaded as a value as an argument of the @lgc.create.read.builtin.input.* call, or some form of metadata?
> 
> How about new LGC option(s) giving the set and descriptor id of the special buffer for this? Then, when the option(s) are not set, it does not do the optimization.

Sounds good!

include/vkgcDefs.h
```diff
@@ -82,6 +82,7 @@
 //  %Version History
 //  | %Version | Change Description                                                                                    |
 //  | -------- | ------------------------------------------------------------------------------------------------------|
+//  |     54.6 | Add reverseThreadGroup to PipelineOptions                                                             |
```
nhaehnle, line 85: Make sure to sync the minor version. Right now there's a gap.

nhaehnle (approved):
LGTM

include/vkgcDefs.h
LLJJDD: There are some internal changes occupying those versions.

qiaojbao:
retest this please

PR #1827, opened 2022-05-30
Add support for scalarize waterfall descriptor loads.

perlfu:
Provide support for scalarizing waterfalled non-uniform descriptor
loads in limited circumstances.
This is an alternative to the current behaviour which scalarizes
access to a vector loaded outside the waterfall loop.

While using scalarized loads introduces memory access into each
loop iteration it significantly reduces VGPR pressure created by
the loop.
It also has the positive side effect of reducing the probability
that a s_waitcnt on the vmcnt is introduced into the waterfall loop
by the backend, as lgkmcnt is used for the loads instead.

In practice performance of the scalarized loads can be equivalent
to vector loads when there is medium to high commonality in the
non-uniform values.

Automatically enable this scalarization when a vgpr limit is set.
As this is where it will be most beneficial.

The limitation of this transform is that loads are only scalarized
when all non-uniform values have a common base/index.
This covers the majority of scenarios.
For a general purpose solution instruction would need to be heavily
reordered to ensure the correct parts occur inside/outside the
waterfall loop.

PR #1990, opened 2022-08-25
Peep through image desc modifications during waterfall generation.

perlfu:
An issue was introduced with PR #1827, where waterfall tracing
interacted incorrectly with workaround code that modifies image
descriptors on some GFX10.3 ASICs.
This would result in compilation errors for some shaders.

This patches fixes the issue in two ways:
 * The scalarizer code now correctly updates the non-uniform
   instruction with waterfall replacement values.
 * The tracing code can now peep through workarounds like those
   used on gfx1032, meaning waterfalls output should be identical
   on these ASICs.

### DIFF
--- a/lgc/builder/BuilderImpl.cpp
+++ b/lgc/builder/BuilderImpl.cpp
@@ -348,8 +348,36 @@ BranchInst *BuilderImplBase::createIf(Value *condition, bool wantElse, const Twi
 // @return : Value representing the non-uniform index
 static Value *traceNonUniformIndex(Value *nonUniformVal) {
   auto load = dyn_cast<LoadInst>(nonUniformVal);
-  if (!load)
-    return nonUniformVal;
+  if (!load) {
+    // Workarounds that modify image descriptor can be peeped through, i.e.
+    //   %baseValue = load <8 x i32>, <8 x i32> addrspace(4)* %..., align 16
+    //   %rawElement = extractelement <8 x i32> %baseValue, i64 6
+    //   %updatedElement = and i32 %rawElement, -1048577
+    //   %nonUniform = insertelement <8 x i32> %baseValue, i32 %updatedElement, i64 6
+    auto insert = dyn_cast<InsertElementInst>(nonUniformVal);
+    if (!insert)
+      return nonUniformVal;
+
+    load = dyn_cast<LoadInst>(insert->getOperand(0));
+    if (!load)
+      return nonUniformVal;
+
+    // We found the load, but must verify the chain.
+    // Consider updatedElement as a generic instruction or constant.
+    if (auto updatedElement = dyn_cast<Instruction>(insert->getOperand(1))) {
+      for (Value *operand : updatedElement->operands()) {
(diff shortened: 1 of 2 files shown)


lgc/builder/BuilderImpl.cpp
```diff
@@ -348,8 +348,36 @@ BranchInst *BuilderImplBase::createIf(Value *condition, bool wantElse, const Twi
...
+      for (Value *operand : updatedElement->operands()) {
+        if (auto extract = dyn_cast<ExtractElementInst>(operand)) {
+          // Only dynamic value must be ExtractElementInst based on load.
+          if (dyn_cast<LoadInst>(extract->getOperand(0)) != load)
+            return nonUniformVal;
+        } else if (!dyn_cast<Constant>(operand)) {
+          return nonUniformVal;
+        }
+      }
+    } else if (!dyn_cast<Constant>(insert->getOperand(1))) {
```
nhaehnle, line 377: Use `isa<Constant>`

nhaehnle (commented):
I'm unhappy that this is so specific to a specific pattern and therefore fragile. On the other hand, getting a proper solid foundation for waterfall loops would really require something like regions (explicit nesting) which we can't reasonably represent in LLVM IR, so... I suppose this is okay?

(Hmm, maybe we could have a `callbr`-based representation of waterfall loops?)

perlfu:
> I'm unhappy that this is so specific to a specific pattern and therefore fragile. On the other hand, getting a proper solid foundation for waterfall loops would really require something like regions (explicit nesting) which we can't reasonably represent in LLVM IR, so... I suppose this is okay?
> 
> (Hmm, maybe we could have a `callbr`-based representation of waterfall loops?)

Reasonable concerns.
Technical the waterfall loops work even if the fragile pattern is broken, just they are less efficient.

One of the things that irritated me working on this patch is that the waterfall loop insertion works on the IR during the LGC replay pass.
So there is no direct way to test the waterfall transformation as it happens on IR that is "still being formed".
Ideally it would be a separate pass, working on some kind of LGC markers after replay.

amdvlk-admin:
### Test summary for commit 3f9020836301795e5f5f99003b5ea3dc1aab757e
<details><summary>CTS tests (Failed: 0/220357)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 37494/73361 (51.1%)</li>
<li>Failed: 0/73361 (0.0%)</li>
<li>Not Supported: 35867/73361 (48.9%)</li><li>Warnings: 0/73361 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (approved):
Thanks for the update.

> One of the things that irritated me working on this patch is that the waterfall loop insertion works on the IR during the LGC replay pass.
> So there is no direct way to test the waterfall transformation as it happens on IR that is "still being formed".
> Ideally it would be a separate pass, working on some kind of LGC markers after replay.

Yeah. This is the kind of thing that I hope we can improve more systematically based on the dialects work.

In the meantime, I think it's okay to use this change.

PR #2010, opened 2022-09-06
Add -fno-aligned-new to disable C++17 default aligned new

bsaleil:
Add the `-fno-aligned-new` flag to LLVM, LLPC and LGC to disable C++17 default aligned new.
Games shipped with an old version of the tcmalloc library may crash when allocating aligned memory with `aligned_alloc`.
C++17 defines `__cpp_aligned_new` by default so we are seeing more crashes since LLVM switched to C++17.

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -34,6 +34,10 @@ function(set_compiler_options PROJECT_NAME ENABLE_WERROR)
             -fno-rtti
             -fPIC
             -std=c++17
+            # Some of the games using old versions of the tcmalloc lib are
+            # crashing when allocating aligned memory. C++17 enables aligned new
+            # by default, so we need to disable it to prevent those crashes.
+            -fno-aligned-new
             -Wno-ignored-qualifiers
             -Wno-missing-field-initializers
             -Wno-invalid-offsetof           # offsetof within non-standard-layout type 'x' is undefined
--- a/llpc/CMakeLists.txt
+++ b/llpc/CMakeLists.txt
@@ -117,6 +117,12 @@ if(ICD_BUILD_LLPC)
             Target
             TransformUtils
         )
+        # Some of the games using old versions of the tcmalloc lib are crashing
+        # when allocating aligned memory. C++17 enables aligned new by default,
+        # so we need to disable it to prevent those crashes.
+        foreach (lib ${llvm_libs})
+          target_compile_options(${lib} PRIVATE "-fno-aligned-new")
+        endforeach()
         target_link_libraries(llpc PUBLIC ${llvm_libs})
     endif()
 
--- a/llpc/CMakeLists.txt
+++ b/llpc/CMakeLists.txt
@@ -128,6 +128,9 @@ if(ICD_BUILD_LLPC)
 
     # Always link statically against libLLVMlgc
     llvm_map_components_to_libnames(extra_llvm_libs lgc)
+    foreach (lib ${extra_llvm_libs})
+      target_compile_options(${lib} PRIVATE "-fno-aligned-new")
+    endforeach()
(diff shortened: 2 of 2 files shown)


amdvlk-admin:
### Test summary for commit bae2c3eacbeaaab5442a586c2da64d115e23cf6c
<details><summary>CTS tests (Failed: 0/289643)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 37494/73370 (51.1%)</li>
<li>Failed: 0/73370 (0.0%)</li>
<li>Not Supported: 35876/73370 (48.9%)</li><li>Warnings: 0/73370 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>

<li>Passed: 36014/69277 (52.0%)</li>
<li>Failed: 0/69277 (0.0%)</li>
<li>Not Supported: 33263/69277 (48.0%)</li><li>Warnings: 0/69277 (0.0%)</li>
</ul></details>


</ul></details>

dstutt (approved):
LGTM

I wondered if there was an llvm cmake function or macro for the foreach you implemented, but I couldn't see anything. I guess it's simple enough to not require one.

nhaehnle (commented):
Oh wow, what a mess. What's the long-term solution here?

There is a risk that something inside of LLVM starts genuinely requiring aligned new. What then?

Default builds should only expose the LLPC interface, which has no movement of object ownership as far as I'm aware: all objects are either allocated and freed outside of LLPC, or allocated and freed inside of LLPC. So in principle, we ought to be able to use our own heap allocator and simply not import any allocator symbols. Is that something we can do in practice?

bsaleil:
@nhaehnle I'm not sure I understand how that could work.
The problem is that the LLVM buffer allocator (in [MemAlloc.cpp](https://github.com/GPUOpen-Drivers/llvm-project/blob/797cecf73ce6a96fdb7caa8ea26ca539a27e9d81/llvm/lib/Support/MemAlloc.cpp#L17)) uses the `new` operator with support for over-alignment. These functions are used in various places and AFAIK, there is no way to provide a custom allocator for that.
And I don't see any clean workaround. I guess we could either replace these functions to use something like `posix_memalign` instead and provide `--allow-multiple-definition` to the linker or redefine the global new/delete operators but I don't think either solution is acceptable.

nhaehnle:
I don't have a ready-made solution, but what I'm envisioning in principle is something along the lines of:

* Leave MemAlloc.cpp as-is but identify the relevant symbols which are imported by the compiler.
* Statically link against a working malloc implementation. (I don't know which one, it'd make sense to do some benchmarking and pick the fastest one. There have been reports of people trying different mallocs with LLVM in the past.)
* Somehow hide the relevant symbols from the outside world so that they don't clash with the host application when the driver is loaded.

I wouldn't exactly call that "clean" either, as there'd be some unusual linker magic involved. On the plus side, it would isolate "our" heap implementation from the host application's heap implementation in a way that systematically avoids this kind of issue in the future.

Flakebi:
There was a discussion about using a custom allocator for llvm recently: https://discourse.llvm.org/t/using-a-custom-allocator-for-llvm/64637/7
One interesting [response](https://discourse.llvm.org/t/using-a-custom-allocator-for-llvm/64637/8):
> So, we did try using -Wl, wrap linker option to guarantee wrapping all the allocations to resolve into symbols we want, but it did have a high compile time cost. There are ways to reduce this, but in the end the maintenance burden was too high (when we started looking at wrapping operator new and all it’s mangled variants).
>
> At this point, we’re thinking more in terms of “providing a common set of API for allocations” rather than “providing the option for a custom allocator for LLVM”. We did start off with the latter, but there’s been convincing arguments against that.

nhaehnle (approved):
Let's go with this for now but please remember to at least file an issue.

PR #2035, opened 2022-09-21
Opaque Pointers: Skip lower const imm store if types are not equal

mariusz-sikora-at-amd:
Zero-index GEP elimination (ConstantFold) is taking place more often with opaque pointers and it has side effects for LowerConstImmediateStore optimization.

Before opaque pointers:
%__llpc_output_proxy_ = alloca { <4 x float> }, align 16, addrspace(5) 
%11 = getelementptr inbounds { <4 x float> }, { <4 x float> } addrspace(5)* %__llpc_output_proxy_, i32 0, i32 0 
store <4 x float> <float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00>, <4 x float> addrspace(5)* %11, align 16

After opaque pointers:
%__llpc_output_proxy_ = alloca { <4 x float> }, align 16, addrspace(5) 
store <4 x float> <float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00>, ptr addrspace(5) %__llpc_output_proxy_, align 16

The mismatch of types is taking place when creating GlobalVariable with constant initialization. GlobalVariable type is based on Alloca type which is a { <4 x float> } and will be initialized by the Constant of a type <4 x float>.

This change is not affecting anything for non opaque pointer mode.

### DIFF
--- a/llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
+++ b/llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
@@ -176,10 +176,15 @@ StoreInst *SpirvLowerConstImmediateStore::findSingleStore(AllocaInst *allocaInst
 void SpirvLowerConstImmediateStore::convertAllocaToReadOnlyGlobal(StoreInst *storeInst) {
   auto allocaInst = cast<AllocaInst>(storeInst->getPointerOperand());
   auto globalType = allocaInst->getAllocatedType();
+  auto initVal = cast<Constant>(storeInst->getValueOperand());
+
+  if (globalType != initVal->getType())
+    return;
+
   auto global = new GlobalVariable(*m_module, globalType,
                                    true, // isConstant
-                                   GlobalValue::InternalLinkage, cast<Constant>(storeInst->getValueOperand()), "",
-                                   nullptr, GlobalValue::NotThreadLocal, SPIRAS_Constant);
+                                   GlobalValue::InternalLinkage, initVal, "", nullptr, GlobalValue::NotThreadLocal,
+                                   SPIRAS_Constant);
   global->takeName(allocaInst);
   // Change all uses of pAllocaInst to use pGlobal. We need to do it manually, as there is a change
   // of address space, and we also need to recreate "getelementptr"s.


amdvlk-admin:
### Test summary for commit ea770efd14672dc0bf6e5ed98f45d168c90a10b6
<details><summary>CTS tests (Failed: 0/286514)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36222/72648 (49.9%)</li>
<li>Failed: 0/72648 (0.0%)</li>
<li>Not Supported: 36426/72648 (50.1%)</li><li>Warnings: 0/72648 (0.0%)</li>
</ul></details>

<li>Passed: 40420/72695 (55.6%)</li>
<li>Failed: 0/72695 (0.0%)</li>
<li>Not Supported: 32275/72695 (44.4%)</li><li>Warnings: 0/72695 (0.0%)</li>
</ul></details>

<li>Passed: 37135/72702 (51.1%)</li>
<li>Failed: 0/72702 (0.0%)</li>
<li>Not Supported: 35567/72702 (48.9%)</li><li>Warnings: 0/72702 (0.0%)</li>
</ul></details>

<li>Passed: 34834/68469 (50.9%)</li>
<li>Failed: 0/68469 (0.0%)</li>
<li>Not Supported: 33635/68469 (49.1%)</li><li>Warnings: 0/68469 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle:
Can you give an example of where this is relevant? Perhaps it would be better to change the GV type or the type of the initializer if there is genuinely only a single store?

Though I am even more suspicious that your motivating example is a `%_llpc_output_proxy`: Trivial cases should probably be optimized away before they're ever looked at by a pass that generates constant globals (which tends to be an expensive operation, though sometimes desirable).

Before applying this change, could you look into moving `SpirvLowerConstImmediateStore` to the optimization part of the llpcSpirvLower pipeline? This really ought to be thought of as an optimization-style pass. It needs to come after SROA (which eliminates allocas where possible) and probably fairly late. So why not move it to the end of the optimization block, i.e. just after EarlyCSE (and before SpirvLowerMathFloatOp)?

This particular change here will then likely still be required, I'd assume, but we can analyze it under conditions that "make more sense".

mariusz-sikora-at-amd:
> Can you give an example of where this is relevant? Perhaps it would be better to change the GV type or the type of the initializer if there is genuinely only a single store?
> 
> Though I am even more suspicious that your motivating example is a `%_llpc_output_proxy`: Trivial cases should probably be optimized away before they're ever looked at by a pass that generates constant globals (which tends to be an expensive operation, though sometimes desirable).
> 
> Before applying this change, could you look into moving `SpirvLowerConstImmediateStore` to the optimization part of the llpcSpirvLower pipeline? This really ought to be thought of as an optimization-style pass. It needs to come after SROA (which eliminates allocas where possible) and probably fairly late. So why not move it to the end of the optimization block, i.e. just after EarlyCSE (and before SpirvLowerMathFloatOp)?
> 
> This particular change here will then likely still be required, I'd assume, but we can analyze it under conditions that "make more sense".

Now I noticed that commit message is lacking newlines...

My motivation was to just to "disable" for now the case where alloca and store are operating on different types and revisit this later, when opaque pointers will be working, because this looks like a good case for some optimizations. I was planning to inspect type of alloca and type of store and based on this decide if read-only GV can be created.
This example with "llpc_out_proxy" was just a first failing test that I found, but it could happened for more complex data structures because opaque pointers "are removing" zero-index GEP. I had in my head cases where you are storing only to the part of the alloca.

As you suggested I will try to move this optimization after SROA and see what will happened.

Right now this change is not causing anything for non-opaque mode and for opaque mode it is just skipping optimization, so you are ending with the same result.

mariusz-sikora-at-amd:
> > Can you give an example of where this is relevant? Perhaps it would be better to change the GV type or the type of the initializer if there is genuinely only a single store?
> > Though I am even more suspicious that your motivating example is a `%_llpc_output_proxy`: Trivial cases should probably be optimized away before they're ever looked at by a pass that generates constant globals (which tends to be an expensive operation, though sometimes desirable).
> > Before applying this change, could you look into moving `SpirvLowerConstImmediateStore` to the optimization part of the llpcSpirvLower pipeline? This really ought to be thought of as an optimization-style pass. It needs to come after SROA (which eliminates allocas where possible) and probably fairly late. So why not move it to the end of the optimization block, i.e. just after EarlyCSE (and before SpirvLowerMathFloatOp)?
> > This particular change here will then likely still be required, I'd assume, but we can analyze it under conditions that "make more sense".
> 
> Now I noticed that commit message is lacking newlines...
> 
> My motivation was to just to "disable" for now the case where alloca and store are operating on different types and revisit this later, when opaque pointers will be working, because this looks like a good case for some optimizations. I was planning to inspect type of alloca and type of store and based on this decide if read-only GV can be created. This example with "llpc_out_proxy" was just a first failing test that I found, but it could happened for more complex data structures because opaque pointers "are removing" zero-index GEP. I had in my head cases where you are storing only to the part of the alloca.
> 
> As you suggested I will try to move this optimization after SROA and see what will happened.
> 
> Right now this change is not causing anything for non-opaque mode and for opaque mode it is just skipping optimization, so you are ending with the same result.

I moved `SpirvLowerConstImmediateStore` pass as you suggested (after `EarlyCSE` and before `SpirvLowerMathFloatOp`) and it helped. I will run some additional tests and prepare new PR (and close this one) if this will not have any impact on performance.

mariusz-sikora-at-amd:
> > > Can you give an example of where this is relevant? Perhaps it would be better to change the GV type or the type of the initializer if there is genuinely only a single store?
> > > Though I am even more suspicious that your motivating example is a `%_llpc_output_proxy`: Trivial cases should probably be optimized away before they're ever looked at by a pass that generates constant globals (which tends to be an expensive operation, though sometimes desirable).
> > > Before applying this change, could you look into moving `SpirvLowerConstImmediateStore` to the optimization part of the llpcSpirvLower pipeline? This really ought to be thought of as an optimization-style pass. It needs to come after SROA (which eliminates allocas where possible) and probably fairly late. So why not move it to the end of the optimization block, i.e. just after EarlyCSE (and before SpirvLowerMathFloatOp)?
> > > This particular change here will then likely still be required, I'd assume, but we can analyze it under conditions that "make more sense".
> > 
> > 
> > Now I noticed that commit message is lacking newlines...
> > My motivation was to just to "disable" for now the case where alloca and store are operating on different types and revisit this later, when opaque pointers will be working, because this looks like a good case for some optimizations. I was planning to inspect type of alloca and type of store and based on this decide if read-only GV can be created. This example with "llpc_out_proxy" was just a first failing test that I found, but it could happened for more complex data structures because opaque pointers "are removing" zero-index GEP. I had in my head cases where you are storing only to the part of the alloca.
> > As you suggested I will try to move this optimization after SROA and see what will happened.
> > Right now this change is not causing anything for non-opaque mode and for opaque mode it is just skipping optimization, so you are ending with the same result.
> 
> I moved `SpirvLowerConstImmediateStore` pass as you suggested (after `EarlyCSE` and before `SpirvLowerMathFloatOp`) and it helped. I will run some additional tests and prepare new PR (and close this one) if this will not have any impact on performance.

@nhaehnle If we move `SpirvLowerConstImmediateStore` after `Early CSE` then the overall compilation time will be **up by 1.3%** There are also some shaders which will compile 3x slower. 
Using `-time-passes` I found two places with the biggest slow down. `Machine Instruction Scheduler` 0.3216 -> 1.6512 and `Code sinking` 0.0013 -> 0.2817.
additionally generated shader will be bigger `_amdgpu_ps_main` 4040 - > 8720 B

I would rather stay with the original change from this PR and create ticket to improve it later, after opaque transition is finished. Current change do not have any performance impacts, because without opaque pointers this will not happened (difference in types) and for opaque pointers mode it will behave as in non opaque pointer mode.

nhaehnle (approved):

PR #2055, opened 2022-09-27
Opaque Pointers: Handle name mangling.

mariusz-sikora-at-amd:
With opaque pointers we are not able to get base type of the pointer, so while creating mangled function name we are able only to indicate that we are dealing with pointer type (func_name.p7 instead of func_name.p7v4f32). This is way return type of the function has been added while mangling the function name.
Second change (which will be later reverted) which is added in this patch is casting all pointer types used in function to i8*.

### DIFF
--- a/llpc/lower/llpcSpirvLowerGlobal.cpp
+++ b/llpc/lower/llpcSpirvLowerGlobal.cpp
@@ -314,6 +314,12 @@ void SpirvLowerGlobal::handleCallInst(bool checkEmitCall, bool checkInterpCall) 
             mangledName.startswith(gSPIRVName::InterpolateAtVertexAMD)) {
           // Translate interpolation functions to LLPC intrinsic calls
           auto loadSrc = callInst->getArgOperand(0);
+
+          // TODO: Remove this when LLPC will switch fully to opaque pointers.
+          // This is needed to handle additional bitcast which was added while creating function call.
+          if (auto bitCast = dyn_cast<BitCastInst>(loadSrc))
+            loadSrc = bitCast->getOperand(0);
+
           unsigned interpLoc = InterpLocUnknown;
           Value *auxInterpValue = nullptr;
 
@@ -1801,7 +1807,16 @@ void SpirvLowerGlobal::lowerBufferBlock() {
 
               // Run the users of the GEP to check for any nonuniform calls.
               for (User *const user : getElemPtr->users()) {
-                CallInst *const call = dyn_cast<CallInst>(user);
+                // TODO: Remove this when LLPC will switch fully to opaque pointers.
+                // TODO: revert this back to 'const'
+                CallInst *call = dyn_cast<CallInst>(user);
+
+                // TODO: Remove this when LLPC will switch fully to opaque pointers.
+                // This is needed to handle additional bitcast which was added while creating function call.
+                if (auto bitCast = dyn_cast<BitCastInst>(user)) {
+                  User *bitCastUser = bitCast->user_back();
+                  call = dyn_cast<CallInst>(bitCastUser);
+                }
                 // If the user is not a call or the call is the function pointer call, bail.
                 if (!call)
                   continue;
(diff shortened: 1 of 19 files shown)


amdvlk-admin:
### Test summary for commit 9a441dadbb83f6d32eb137200a2af55ca6650184
<details><summary>CTS tests (Failed: 0/286530)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36222/72652 (49.9%)</li>
<li>Failed: 0/72652 (0.0%)</li>
<li>Not Supported: 36430/72652 (50.1%)</li><li>Warnings: 0/72652 (0.0%)</li>
</ul></details>

<li>Passed: 40419/72699 (55.6%)</li>
<li>Failed: 0/72699 (0.0%)</li>
<li>Not Supported: 32280/72699 (44.4%)</li><li>Warnings: 0/72699 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -314,6 +314,12 @@ void SpirvLowerGlobal::handleCallInst(bool checkEmitCall, bool checkInterpCall)
             mangledName.startswith(gSPIRVName::InterpolateAtVertexAMD)) {
           // Translate interpolation functions to LLPC intrinsic calls
           auto loadSrc = callInst->getArgOperand(0);
+
+          // TODO: Remove this when LLPC will switch fully to opaque pointers.
+          // This is needed to handle additional bitcast which was added while creating function call.
+          if (auto bitCast = dyn_cast<BitCastInst>(loadSrc))
+            loadSrc = bitCast->getOperand(0);
+
```
nhaehnle, line 322: Can you please move this lower to just above the distinction between `GetElementPtrInst` and `GlobalVariable`? Then all the `loadSrc` handling is together in one place.

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -1801,7 +1807,16 @@ void SpirvLowerGlobal::lowerBufferBlock() {
...
+                // TODO: Remove this when LLPC will switch fully to opaque pointers.
+                // TODO: revert this back to 'const'
+                CallInst *call = dyn_cast<CallInst>(user);
+
+                // TODO: Remove this when LLPC will switch fully to opaque pointers.
+                // This is needed to handle additional bitcast which was added while creating function call.
+                if (auto bitCast = dyn_cast<BitCastInst>(user)) {
+                  User *bitCastUser = bitCast->user_back();
+                  call = dyn_cast<CallInst>(bitCastUser);
+                }
```
nhaehnle, line 1819: This makes awkward assumptions about exactly how the IR looks like at this stage. Consider what happens if `bitcast`s have been CSE'd in the meantime? There could be multiple users of the bitcast, all of which would need to be checked.

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5696,6 +5710,10 @@ static void printTypeName(Type *ty, raw_ostream &nameStream) {
     nameStream << "]";
     return;
   }
+  if (auto pointerTy = dyn_cast<PointerType>(ty)) {
+    nameStream << "p" << pointerTy->getAddressSpace();
+    return;
+  }
```
nhaehnle, line 5716: Suggest an assertion that `pointerTy` is either opaque or points to an i8.

nhaehnle (commented):
This makes a lot of sense to me. Some comments though inline.

llpc/lower/llpcSpirvLowerGlobal.cpp
mariusz-sikora-at-amd: Thanks!, now as I'm thinking about this you may be right. I will fix this !

llpc/translator/lib/SPIRV/SPIRVReader.cpp
mariusz-sikora-at-amd: After adding this assertion I realized that all of this bitcasting may not be needed. Why do I think so? Because after adding this assertion I noticed that I missed one place and still everything was working (I missed adding changes to create row major matrix).
Now I'm thinking about revert bitcasts from this RP and change the way of creating functions.
Right now before creating new function we are checking if function (with the same name) is in the module. If function exists then we are using existing one, if not then we are creating new function.
I'm thinking about removing search of existing function and every time create a new function. From naming/type perspective there will not be any issue, because if function with specific name already exists then LLVM is adding suffix to the end of the newly created function (func, func.1, func.2). All of these functions can have the same argument types or different types.
I have to check (and you know you can save my time :) ) if there is a case in which we rely that only single function (with specific name and arguments) will be created and will have multiple users (multiple calls). The disadvantages of this solution is that we will end with functions which are having single users.
Thoughts ?

nhaehnle: Creating a new function each time may hurt the effectiveness of some optimizing transforms like CSE. I think the bitcasting approach is better, even if it's possible for some issue to slip through temporarily.

mariusz-sikora-at-amd: @nhaehnle please check other version here: https://github.com/GPUOpen-Drivers/llpc/pull/2065

PR #2059, opened 2022-09-29
Remove obsolete CMake code

yavn:
Additionally, that part caused build errors with multi-config generators.


amdvlk-admin:

Flakebi:
ok to test

yavn:
Looks like I removed too much, sorry. I will take another stab at multi-config problem directly.

PR #2064, opened 2022-09-30
Tweak LLVM module path to work with Visual Studio generator

yavn:
I believe there is a bug in LLVM. The CMake modules will get copied to the build directory with a path such as `build_dir/compiler/llpc/llvm/$(Configuration)/lib/cmake/llvm`. With a literal, non-expanded `$(Configuration)` token in there, which doesn't make much sense.

The fix looks generic, but it won't help Ninja Multi-Config. For that I will just try to get the root cause fixed in LLVM.

The execute_process change is not needed, I added it for clarity that this command most of the time doesn't even work (because the file isn't there).

### DIFF
--- a/llpc/CMakeLists.txt
+++ b/llpc/CMakeLists.txt
@@ -144,16 +144,13 @@ set_compiler_options(llpc ${LLPC_ENABLE_WERROR})
 
 ### Defines/Includes/Sources ###########################################################################################
 if(ICD_BUILD_LLPC)
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${XGL_LLVM_BUILD_PATH}/lib/cmake/llvm)
+    list(APPEND CMAKE_MODULE_PATH
+        ${XGL_LLVM_BUILD_PATH}/lib/cmake/llvm
+        ${XGL_LLVM_BUILD_PATH}/${CMAKE_CFG_INTDIR}/lib/cmake/llvm   # Workaround for multi-config generators (e.g. VS)
+    )
     include(LLVMConfig)
     message(STATUS "LLVM executables: " ${LLVM_TOOLS_BINARY_DIR})
     message(STATUS "LLVM libraries: " ${LLVM_BUILD_LIBRARY_DIR})
-    execute_process(
-        COMMAND ${LLVM_TOOLS_BINARY_DIR}/llvm-config --libs amdgpu analysis bitreader bitwriter codegen irreader linker mc passes support target transformutils
-        OUTPUT_VARIABLE LLVM_LINK_FLAGS
-        OUTPUT_STRIP_TRAILING_WHITESPACE
-    )
-    message(STATUS "LLVM link options:" ${LLVM_LINK_FLAGS})
 endif()
 target_compile_definitions(llpc PRIVATE ${TARGET_ARCHITECTURE_ENDIANESS}ENDIAN_CPU)
 target_compile_definitions(llpc PRIVATE _SPIRV_LLVM_API)


amdvlk-admin:

llpc/CMakeLists.txt
```diff
@@ -144,16 +144,21 @@ set_compiler_options(llpc ${LLPC_ENABLE_WERROR})
 
 ### Defines/Includes/Sources ###########################################################################################
 if(ICD_BUILD_LLPC)
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${XGL_LLVM_BUILD_PATH}/lib/cmake/llvm)
+    list(APPEND CMAKE_MODULE_PATH
+        ${XGL_LLVM_BUILD_PATH}/lib/cmake/llvm
+        ${XGL_LLVM_BUILD_PATH}/${CMAKE_CFG_INTDIR}/lib/cmake/llvm   # Workaround for multi-config generators (e.g. VS)
+    )
```
Flakebi, line 150: I think we should only include one if these paths. `CMAKE_CFG_INTDIR` is an empty string for non-multi-config generators, so the second path should work there as well.

llpc/CMakeLists.txt
```diff
@@ -144,16 +144,21 @@ set_compiler_options(llpc ${LLPC_ENABLE_WERROR})
...
-        OUTPUT_VARIABLE LLVM_LINK_FLAGS
-        OUTPUT_STRIP_TRAILING_WHITESPACE
-    )
-    message(STATUS "LLVM link options:" ${LLVM_LINK_FLAGS})
+    if(EXISTS ${LLVM_TOOLS_BINARY_DIR}/llvm-config)
+        execute_process(
+            COMMAND ${LLVM_TOOLS_BINARY_DIR}/llvm-config --libs amdgpu analysis bitreader bitwriter codegen irreader linker mc passes support target transformutils
+            OUTPUT_VARIABLE LLVM_LINK_FLAGS
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
```
Flakebi, line 159: Do we need to run llvm-config at all (I though it was deprecated)?
I don’t like putting it behind an exists flag. If it’s not needed we should remove it. If it’s needed, it should fail here instead of miscompile or fail later due to variables missing.

Flakebi (commented):
Multi-config builds on llvm are difficult and because most of the community doesn’t use it, breaking it goes unnoticed for a long time and nobody tries to replace the deprecated `CMAKE_CFG_INTDIR` with generator expressions.

llpc/CMakeLists.txt
yavn: True, it would work for single- and multi-config generators now, but I decided to add it as a separate path in case LLVM is fixed at some point. If that happens, `CMAKE_CFG_INTDIR` in the path would break VS generator. Having it in path doesn't make sense, so I'd rather have it separately so it's obvious this is a kind of a hack.

yavn:
Can someone help kick off that required pipeline check? Thanks.

Flakebi:
ok to test

qiaojbao:
/jenkins/workspace/vulkan/sanitized-opensource/Github-PR/llpc-github-pr/driver_build/xgl/icd/api/pipeline_compiler.cpp:2615:50: error: ‘struct Vkgc::PipelineShaderOptions’ has no member named ‘forceLateZ’
21:11:34               pCreateInfo->pipelineInfo.fs.options.forceLateZ = pRenderPass->IsForceLateZNeeded();
21:11:34                                                    ^~~~~~~~~~
21:11:34  /jenkins/workspace/vulkan/sanitized-opensource/Github-PR/llpc-github-pr/driver_build/xgl/icd/api/pipeline_compiler.cpp: In member function ‘void vk::PipelineCompiler::ApplyPipelineOptions(const vk::Device*, VkPipelineCreateFlags, Vkgc::PipelineOptions*)’:
21:11:34  /jenkins/workspace/vulkan/sanitized-opensource/Github-PR/llpc-github-pr/driver_build/xgl/icd/api/pipeline_compiler.cpp:2742:15: error: ‘struct Vkgc::PipelineOptions’ has no member named ‘rtMaxRayLength’
21:11:34       pOptions->rtMaxRayLength = settings.rtMaxRayLength;
21:11:34                 ^~~~~~~~~~~~~~

yavn:
Thanks for checking, but yeah, that error looks rather unrelated to this... I'll follow up.

qiaojbao:
Maybe your PR branch too old.

yavn:
Ah, got it, thanks. I thought it would be testing a merge commit automatically.

yavn:
ok to test

amdvlk-admin:
### Test summary for commit 6085a6e8d903a6b06dd14bc6331fb9380eac0b6d
<details><summary>CTS tests (Failed: 0/286649)</summary><ul><li>Built with version 1.3.0.0</li>

</ul></details>

<li>Passed: 40419/72724 (55.6%)</li>
<li>Not Supported: 32305/72724 (44.4%)</li><li>Warnings: 0/72724 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/CMakeLists.txt
nhaehnle: Having read https://cmake.org/cmake/help/latest/variable/CMAKE_CFG_INTDIR.html, I tend to agree with @yavn: we really shouldn't be using CMAKE_CFG_INTDIR here if everything was working as it is supposed to. The only reason for doing so is to workaround a bug elsewhere.

nhaehnle (approved):

yavn:
This was already worked around by another patch. The differences in implementation are minor and we can let this one go.

PR #2065, opened 2022-09-30
Opaque Pointers: Handle name mangling.

mariusz-sikora-at-amd:
With opaque pointers we are not able to get base type of the pointer, so while creating mangled function name we are able only to indicate that we are dealing with pointer type (func_name.p7 instead of func_name.p7v4f32). This is way return type of the function has been added while mangling the function name.

### DIFF
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -1574,6 +1574,21 @@ void SPIRVToLLVM::updateDebugLoc(SPIRVValue *bv, Function *f) {
 }
 
 // =====================================================================================================================
+// This function will be removed when LLPC switch to opaque pointers
+//
+// Find function in the module with specific type and starting with name.
+Function *SPIRVToLLVM::findFunctionInModule(string name, FunctionType *funcType) {
+
+  for (Function &func : m_m->functions()) {
+    if (func.getName().startswith(name) && func.getFunctionType() == funcType) {
+      return &func;
+    }
+  }
+
+  return nullptr;
+}
+
+// =====================================================================================================================
 // Create a call to launder a row major matrix.
 //
 // @param pointerToMatrix : The pointer to matrix to launder.
@@ -1591,7 +1606,8 @@ std::pair<Type *, Value *> SPIRVToLLVM::createLaunderRowMajorMatrix(Type *const 
       FunctionType::get(newMatrixPointerType, {matrixPointerType, matrixType}, false);
   string mangledName(SpirvLaunderRowMajor);
   appendTypeMangling(nullptr, args, mangledName);
-  Function *rowMajorFunc = m_m->getFunction(mangledName);
+  // TODO: Remove this when LLPC will switch fully to opaque pointers.
+  Function *rowMajorFunc = findFunctionInModule(mangledName, rowMajorFuncType);
   if (!rowMajorFunc) {
     rowMajorFunc = Function::Create(rowMajorFuncType, GlobalValue::ExternalLinkage, SpirvLaunderRowMajor, m_m);
   }
@@ -5688,8 +5704,7 @@ static void printTypeName(Type *ty, raw_ostream &nameStream) {
   for (;;) {
     if (auto pointerTy = dyn_cast<PointerType>(ty)) {
       nameStream << "p" << pointerTy->getAddressSpace();
-      ty = pointerTy->getPointerElementType();
-      continue;
+      return;
     }
     if (auto arrayTy = dyn_cast<ArrayType>(ty)) {
       nameStream << "a" << arrayTy->getNumElements();
@@ -5790,16 +5805,17 @@ Instruction *SPIRVToLLVM::transBuiltinFromInst(const std::string &funcName, SPIR
 #endif
 
   std::string mangledName(funcName);
-  appendTypeMangling(nullptr, args, mangledName);
-  Function *func = m_m->getFunction(mangledName);
+  appendTypeMangling(retTy, args, mangledName);
   FunctionType *ft = FunctionType::get(retTy, argTys, false);
+  // TODO: Remove this when LLPC will switch fully to opaque pointers.
+  Function *func = findFunctionInModule(mangledName, ft);
   // ToDo: Some intermediate functions have duplicate names with
   // different function types. This is OK if the function name
   // is used internally and finally translated to unique function
   // names. However it is better to have a way to differentiate
   // between intermediate functions and final functions and make
   // sure final functions have unique names.
-  if (!func || func->getFunctionType() != ft) {
+  if (!func) {
     func = Function::Create(ft, GlobalValue::ExternalLinkage, mangledName, m_m);
     func->setCallingConv(CallingConv::SPIR_FUNC);
     if (isFuncNoUnwind())
@@ -7561,7 +7577,10 @@ bool SPIRVToLLVM::transShaderDecoration(SPIRVValue *bv, Value *v) {
       // needed.
       std::string mangledFuncName(gSPIRVMD::NonUniform);
       appendTypeMangling(nullptr, args, mangledFuncName);
-      auto f = getOrCreateFunction(m_m, voidTy, types, mangledFuncName);
+      // TODO: Remove this when LLPC will switch fully to opaque pointers.
+      Function *f = findFunctionInModule(mangledFuncName, FunctionType::get(voidTy, types, false));
+      if (!f)
+        f = getOrCreateFunction(m_m, voidTy, types, mangledFuncName);
       if (bb->getTerminator()) {
         // NOTE: For OpCopyObject, we use the operand value directly, which may be in another block that already has a
         // terminator. In this case, insert the call before the terminator.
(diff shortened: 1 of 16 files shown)


amdvlk-admin:
### Test summary for commit a4ea8b5b7038f107cc6d448dc94265c165ec619d
<details><summary>CTS tests (Failed: 0/286562)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36222/72692 (49.8%)</li>
<li>Failed: 0/72692 (0.0%)</li>
<li>Not Supported: 36470/72692 (50.2%)</li><li>Warnings: 0/72692 (0.0%)</li>
</ul></details>

<li>Passed: 40419/72691 (55.6%)</li>
<li>Failed: 0/72691 (0.0%)</li>
<li>Not Supported: 32272/72691 (44.4%)</li><li>Warnings: 0/72691 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (commented):
So this is the version without bitcasts, right?

Have you considered the issue of optimizations like CSE no longer working correctly?

mariusz-sikora-at-amd:
> So this is the version without bitcasts, right?
> 
> Have you considered the issue of optimizations like CSE no longer working correctly?

Yes, this is a version without bitcasts and yes, I was thinking about CSE and that's why I changed the approach. This PR is changing only the name of the functions and is NOT introducing the idea that I described here: https://github.com/GPUOpen-Drivers/llpc/pull/2055. (Creating new function each time)
Right now before creating new function we are checking if function with that same name already exists in the module. After removing `getPointerElementType()` but not switching to opaque pointer mode we will end up having the same name for two different functions.
Let's say we want to create two functions. One which will take `<4 x i32> addrspace(0)*` as an argument and second with `<8 x i32> addrspace(0)*` argument and the name will be `dummy` + mangling.
Right now we will create two functions:
`void dummy.p0v4i32(<4 x i32> addrspace(0)*` and `void dummy.p0v8i32(<8 x i32> addrspace(0)*`.
After removing `getPoitnerElementType()` and not switching to opaque pointers we will create single function like `void dommy.p0(<4 x i32> addrspace(0)*` and when creating function for `<8 x i32>` we will see that function with the name `dummy.p0` is already created and we will just take it (and break somewhere later while compiling).
This PR adds additional comparison of function types (type of function which we want to create with type of the function which is already created in the module and has the same name) before creating new function.
Going back to the example with `void dummy.p0(<4 x i32> addrspace(0)* ...` we will end up having two function:
`void dummy.p0(<4 x i32> addrspace(0)* ...` and `void dummy.p0.1(<8 x i32> addrspace(0)* ...`.

nhaehnle:
Okay, so if we first create `dummy.p0(<4 x i32> addrspace(0)* ...)` and then later use dummy twice with an `<8 x i32>` pointer, we end up with `dummy.p0.1(<8 x i32> addrspace(0)* ...)` and `dummy.p0.2(<8 x i32> addrspace(0)* ...)`. That doesn't seem right to me?

mariusz-sikora-at-amd:
For non opaque pointers you will see only changes in naming.
before:
`dummy.p0v4i32(<4 x i32> addrspace(0)* ...)`
`dummy.p0x8i32(<8 x i32> addrspace(0)* ...)`
after:
`dummy.p0(<4 x i32> addrspace(0)* ...)`
`dummy.p0.1(<8 x i32> addrspace(0)* ...)`

No I see that my previous idea (having new function for each call) introduced some uncertainty now.

Flakebi:
> > Okay, so if we first create `dummy.p0(<4 x i32> addrspace(0)* ...)` and then later use dummy twice with an `<8 x i32>` pointer, we end up with `dummy.p0.1(<8 x i32> addrspace(0)* ...)` and `dummy.p0.2(<8 x i32> addrspace(0)* ...)`. That doesn't seem right to me?
> 
> No, you will have single function for `<8 x i32>` and in your example it will be `dummy.p0.1(<8 x i32> addrspace(0)* ...)`. `dummy.p0.2` will not be created for `<8 x i32>` because before creating new function now we are checking if types are the same: `if (!rowMajorFunc || rowMajorFunc->getFunctionType() != rowMajorFuncType)`

I think what Nicolai wanted to point out is that after the first function for `<4 x i32>*` was created,
```c++
Function *rowMajorFunc = m_m->getFunction(mangledName);
```
will always return the first function. So, even when `dummy.p0.1` exists for `<8 x i32>*`, `rowMajorFunc` points to `dummy.p0` and the types don’t match for `<8 x i32>*`, so a new function will created every time and it will end up with `dummy.p0.1` and `dummy.p0.2`.

mariusz-sikora-at-amd:
> > > Okay, so if we first create `dummy.p0(<4 x i32> addrspace(0)* ...)` and then later use dummy twice with an `<8 x i32>` pointer, we end up with `dummy.p0.1(<8 x i32> addrspace(0)* ...)` and `dummy.p0.2(<8 x i32> addrspace(0)* ...)`. That doesn't seem right to me?
> > 
> > 
> > No, you will have single function for `<8 x i32>` and in your example it will be `dummy.p0.1(<8 x i32> addrspace(0)* ...)`. `dummy.p0.2` will not be created for `<8 x i32>` because before creating new function now we are checking if types are the same: `if (!rowMajorFunc || rowMajorFunc->getFunctionType() != rowMajorFuncType)`
> 
> I think what Nicolai wanted to point out is that after the first function for `<4 x i32>*` was created,
> 
> ```c++
> Function *rowMajorFunc = m_m->getFunction(mangledName);
> ```
> 
> will always return the first function. So, even when `dummy.p0.1` exists for `<8 x i32>*`, `rowMajorFunc` points to `dummy.p0` and the types don’t match for `<8 x i32>*`, so a new function will created every time and it will end up with `dummy.p0.1` and `dummy.p0.2`.

Oh... you are right. If I don't think of anything better I'll revert this to bitcasts.

mariusz-sikora-at-amd:
I could try to search through all function in the module and see if function name is starting with `mangledName` and if type is the same.
`bool function_exists = false;
for (Function &func : m_m->functions())
  if (func.getName().startswith(mangledName) && func.getFunctionType() == func_type))
    function_exists = true;

if (!function_exists)
  // Create function
`
This code will be removed after the transition. What do you think about it @Flakebi @nhaehnle

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -5688,8 +5688,7 @@ static void printTypeName(Type *ty, raw_ostream &nameStream) {
   for (;;) {
     if (auto pointerTy = dyn_cast<PointerType>(ty)) {
       nameStream << "p" << pointerTy->getAddressSpace();
-      ty = pointerTy->getPointerElementType();
-      continue;
+      return;
```
Flakebi, line 5691: I’m sorry if I missed something in the discussion, but to get equivalent code with and without opaque pointers, can we do
```c++
if (pointerTy->isOpaque())
  return;
ty = pointerTy->getPointerElementType();
continue;
```
here?

That means we get functions with types in their names and signatures without opaque pointers.
And we get functions without types in both, names and signature, with enabled opaque pointers.

mariusz-sikora-at-amd: We could do this, but we will end up with `getPointerElementType` which is deprecated and can be removed any day since LLVM 15 was released. My plan was to remove every `getPointerElementType` and later fix all issues found while enabling opaque pointers.

amdvlk-admin:
### Test summary for commit e899057310280dfe35438a31fb273e0be9699189
<details><summary>CTS tests (Failed: 0/286639)</summary><ul><li>Built with version 1.3.0.0</li>

</ul></details>

<li>Passed: 40419/72714 (55.6%)</li>
<li>Failed: 0/72714 (0.0%)</li>
<li>Not Supported: 32295/72714 (44.4%)</li><li>Warnings: 0/72714 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/translator/lib/SPIRV/SPIRVReader.cpp
mariusz-sikora-at-amd: I updated this PR with new way of searching for matching functions, but I can leave this change as a backup if tomorrow they will remove the deprecated functions and proceed with changes that you suggested.

Flakebi: Thanks! Both is fine with me, let’s see what @nhaehnle thinks.

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1573,6 +1573,21 @@ void SPIRVToLLVM::updateDebugLoc(SPIRVValue *bv, Function *f) {
...
+Function *SPIRVToLLVM::findFunctionInModule(string name, FunctionType *funcType) {
+
+  for (Function &func : m_m->functions()) {
+    if (func.getName().startswith(name) && func.getFunctionType() == funcType) {
+      return &func;
+    }
+  }
+
+  return nullptr;
+}
```
nhaehnle, line 1589: A bit unorthodox, but I suppose we can live with it briefly.

**However**, please add a fast path that uses the `Module`'s own lookup functions once and checks the type. Those lookups use a hash map and are therefore much faster than iteration. If you don't get any result from the lookup, you can immediately return null. If you get an incorrectly typed result, you can go into the loop.

mariusz-sikora-at-amd: Thanks ! I will move this to draft for now.

PR #2070, opened 2022-10-05
Add aggressiveInvariantLoads and disableInvariantLoads options

perlfu:
Add patching pass to modify invariant load metadata based on aggressiveInvariantLoads and disableInvariantLoads in PipelineShaderOptions.

If aggressiveInvariantLoads is set then all loads in a shader are marked invariant if there are no memory writes in the shader. Note: barriers, discards, and inline asm are considered writes to memory so this generally safe.

If disableInvariantLoads is set then all invariant load metadata is striped from loads in the shader.
This is useful for optimising edges where sink in causing unwanted register pressure increases.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -47,7 +47,7 @@
 #define LLPC_INTERFACE_MAJOR_VERSION 55
 
 /// LLPC minor interface version.
-#define LLPC_INTERFACE_MINOR_VERSION 1
+#define LLPC_INTERFACE_MINOR_VERSION 2
 
 #ifndef LLPC_CLIENT_INTERFACE_MAJOR_VERSION
 #error LLPC client version is not defined
@@ -82,7 +82,8 @@
 //  %Version History
 //  | %Version | Change Description                                                                                    |
 //  | -------- | ----------------------------------------------------------------------------------------------------- |
-//  |     55.1 | Add nsaThreshold to PipelineShaderOptions
+//  |     55.2 | Add aggressiveInvariantLoads and disableInvariantLoads to PipelineShaderOptions                       |
+//  |     55.1 | Add nsaThreshold to PipelineShaderOptions                                                             |
 //  |     55.0 | Remove isInternalRtShader from module options                                                         |
 //  |     54.9 | Add internalRtShaders to PipelineOptions to allow for dumping this data                               |
 //  |     54.6 | Add reverseThreadGroup to PipelineOptions                                                             |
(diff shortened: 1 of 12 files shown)


amdvlk-admin:
### Test summary for commit 945edbb0334efd74e6adea41f22fa64be2791ea9
<details><summary>CTS tests (Failed: 0/286570)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36222/72692 (49.8%)</li>
<li>Failed: 0/72692 (0.0%)</li>
<li>Not Supported: 36470/72692 (50.2%)</li><li>Warnings: 0/72692 (0.0%)</li>
</ul></details>

<li>Passed: 40419/72699 (55.6%)</li>
<li>Failed: 0/72699 (0.0%)</li>
<li>Not Supported: 32280/72699 (44.4%)</li><li>Warnings: 0/72699 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (commented):
Why is the addition of invariant load metadata behind a flag? Either it is safe and correct to add it, then we should do it unconditionally; or it's not safe, then there's a question of whether we should do it at all.

dstutt:
> Why is the addition of invariant load metadata behind a flag? Either it is safe and correct to add it, then we should do it unconditionally; or it's not safe, then there's a question of whether we should do it at all.

You mean the aggressive option? Presumably this isn't a safe/unsafe issue, but more one of performance? Although @perlfu can confirm.
Since enabling it by default can have performance impact on some shaders (presumably by increasing register pressure), it isn't a good idea to blanket enable it?

perlfu:
> > Why is the addition of invariant load metadata behind a flag? Either it is safe and correct to add it, then we should do it unconditionally; or it's not safe, then there's a question of whether we should do it at all.
> 
> You mean the aggressive option? Presumably this isn't a safe/unsafe issue, but more one of performance? Although @perlfu can confirm. Since enabling it by default can have performance impact on some shaders (presumably by increasing register pressure), it isn't a good idea to blanket enable it?

@dstutt's explanation is correct.  Both options are safe on or off.  In some edge cases increasing sinking (beyond current levels) has an adverse in patch on register pressure.  In time I suspect we might be able to validate aggressive to be default on, or determine an appropriate heuristic, but I'd prefer not to suddenly cause a bunch of performance drops in edge cases.

nhaehnle (approved):
Fine. We really need to do better on this front though.

PR #2073, opened 2022-10-07
Remove explicit setting of zlib to be removed from cmake

mariusz-sikora-at-amd:
Explicit disable of ZLIB in LLPC is in conflict with new upstream llvm changes for ZSTD.


Flakebi:
Looks like a good cleanup to me.
Maybe it makes sense to put up the same change as an internal PR to get CI results for Windows?

dstutt:
Explicitly adding @amdrexu  as a reviewer to check if there a reason that zlib was disabled.
Otherwise this LGTM - I think @Flakebi 's suggestion is a good one.

amdrexu (approved):
LGTM

nhaehnle (approved):
Yay, fewer options!

PR #2079, opened 2022-10-11
Add lgc::Pipeline::getShaderStage()

trenouf:
... allowing the front-end to get the shader stage for a function back that it set earlier. Also allow setting a shader stage of invalid, removing any previous shader stage mark for the function.

### DIFF
--- a/lgc/interface/lgc/Pipeline.h
+++ b/lgc/interface/lgc/Pipeline.h
@@ -702,7 +702,7 @@ public:
   // @param stage : Shader stage, or ShaderStageInvalid if none
   static void markShaderEntryPoint(llvm::Function *func, ShaderStage stage);
 
-  // Get a function's shader stage, as marked by markShaderEntryPoint.
+  // Get a function's shader stage.
   //
   // @param func : Function to check
   // @returns stage : Shader stage, or ShaderStageInvalid if none
--- a/lgc/interface/lgc/Pipeline.h
+++ b/lgc/interface/lgc/Pipeline.h
@@ -698,9 +698,15 @@ public:
   // with irLink(). This is a static method in Pipeline, as it does not need a Pipeline object, and can be used
   // in the front-end before a shader is associated with a pipeline.
   //
-  // @param func : Shader entry-point function
-  // @param stage : Shader stage
+  // @param func : Function to mark
+  // @param stage : Shader stage, or ShaderStageInvalid if none
   static void markShaderEntryPoint(llvm::Function *func, ShaderStage stage);
+
+  // Get a function's shader stage, as marked by markShaderEntryPoint.
+  //
+  // @param func : Function to check
+  // @returns stage : Shader stage, or ShaderStageInvalid if none
+  static ShaderStage getShaderStage(llvm::Function *func);
 
   // Link the individual shader modules into a single pipeline module. The front-end must have
   // finished calling Builder::Create* methods and finished building the IR. In the case that
--- a/lgc/state/Compiler.cpp
+++ b/lgc/state/Compiler.cpp
@@ -73,7 +73,7 @@ void Pipeline::markShaderEntryPoint(Function *func, ShaderStage stage) {
 }
 
 // =====================================================================================================================
-// Get a function's shader stage, as marked by markShaderEntryPoint.
+// Get a function's shader stage.
 //
 // @param func : Function to check
 // @returns stage : Shader stage, or ShaderStageInvalid if none
--- a/lgc/state/Compiler.cpp
+++ b/lgc/state/Compiler.cpp
@@ -58,15 +58,27 @@ ElfLinker *createElfLinkerImpl(PipelineState *pipelineState, llvm::ArrayRef<llvm
 // in the front-end before a shader is associated with a pipeline.
 //
 // @param func : Shader entry-point function
-// @param stage : Shader stage
+// @param stage : Shader stage or ShaderStageInvalid
 void Pipeline::markShaderEntryPoint(Function *func, ShaderStage stage) {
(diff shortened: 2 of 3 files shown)


amdvlk-admin:
### Test summary for commit 458794c647740aff62069e63a14d2338bebcfc0f
<details><summary>CTS tests (Failed: 0/286649)</summary><ul><li>Built with version 1.3.0.0</li>

</ul></details>

<li>Passed: 40419/72724 (55.6%)</li>
<li>Not Supported: 32305/72724 (44.4%)</li><li>Warnings: 0/72724 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/interface/lgc/Pipeline.h
```diff
@@ -698,10 +698,16 @@ class Pipeline {
   // with irLink(). This is a static method in Pipeline, as it does not need a Pipeline object, and can be used
   // in the front-end before a shader is associated with a pipeline.
   //
-  // @param func : Shader entry-point function
-  // @param stage : Shader stage
+  // @param func : Function to mark
+  // @param stage : Shader stage, or ShaderStageInvalid if none
   static void markShaderEntryPoint(llvm::Function *func, ShaderStage stage);
 
+  // Get a function's shader stage, as marked by markShaderEntryPoint.
```
nhaehnle, line 705: This isn't entirely correct, is it? getShaderStage will return a shader stage as set even if not marked by markShaderEntryPoint.

The distinction is actually important if we don't inline functions: you could have free functions that still "belong" to a shader stage but aren't entry points. So, they wouldn't have been marked by markShaderEntryPoint, but it probably makes sense to set their shader stage. And on that note, perhaps setShaderStage should be available to the frontend as well?

Cc @tsymalla

tsymalla: I agree with the first point, in the case of `NoInline` non-entrypoint functions, which we could get by just marking them in `llpcSpirvLowerTranslator`, the comment is not entirely correct.

Making `setShaderStage` available to the frontend sounds good. When having free functions in the module, they are initially without shader stage, the assignment can only happen by examining the call instructions inside an entrypoint function. To do the assignment, all calls inside a entrypoint function need to be processed before entering the middle-end (inside `llpcSpirvLowerTranslator`), right after the entrypoint has been marked with the corresponding shader stage. This also needs to happen for nested `NoInline` functions. So, having a way to do the `NoInline` <-> ShaderStage mapping as early as possible could reduce trouble, for instance, when processing the calls to the function inside `PatchEntryPointMutate`, where we currently assume that only entry-point functions and declarations appear inside a module which would not be the case with `NoInline` functions, which would require juggling around with the shader stages for such functions.

trenouf: Right, we don't really support non-inlined funcs currently. If we do want to in the future, I would argue that it is not by a non-inlined function being "associated" with a shader. What if there is some IR transformation that commons up two identical functions, or mostly commons up two similar functions? It is then up to the middle-end to determine how a function needs to be compiled (wave size etc) based on where it is called from, and that might require duplicating it if it has ended up being called from two shaders with different requirements.

So I would say that a non-shader function would not be marked with a shader stage by the front-end.

Unless you're going to tell me that spir-v allows a function to access stuff that is local to its calling shader, such as the shader inputs. In which case I need to rethink...

trenouf: Apparently spir-v does allow a function to access shader inputs etc from the shader it is associated with.

Currently we always inline those, so it is not a problem. If we ever decide not to, I think that should be sorted out in the front-end (e.g. by modifying the IR to pass in extra args), and not left to PatchEntryPointMutate. So the front-end would only mark the shader stage on actual shaders, not functions associated with them.

On that basis, is this change ok?

nhaehnle: I think it's fine to hold off on exposing setShaderStage until we know exactly how we want to deal with this situation; i.e., whether the middle-end is responsible for untangling calls and duplicating functions if necessary, or if an earlier pass inserts an explicit "context" argument, or what.

Personally, I'd still tone down the comment and just say "get a function's shader stage". After all, `irLink` does set this for non-entry points if I'm reading the code right.

trenouf: Ah, you're right, `irLink` does do that. But we only use and test that for compute, and I'm not even sure we do that.

Yes, I'll tone down the comment.

amdvlk-admin:
### Test summary for commit 94ac543117ac80a07c1bd42d727670bf84a219f7
<details><summary>CTS tests (Failed: 0/277760)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 34841/70472 (49.4%)</li>
<li>Failed: 0/70472 (0.0%)</li>
<li>Not Supported: 35631/70472 (50.6%)</li><li>Warnings: 0/70472 (0.0%)</li>
</ul></details>

<li>Passed: 38907/70483 (55.2%)</li>
<li>Failed: 0/70483 (0.0%)</li>
<li>Not Supported: 31576/70483 (44.8%)</li><li>Warnings: 0/70483 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70483 (50.7%)</li>
<li>Failed: 0/70483 (0.0%)</li>
<li>Not Supported: 34733/70483 (49.3%)</li><li>Warnings: 0/70483 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66322 (50.5%)</li>
<li>Failed: 0/66322 (0.0%)</li>
<li>Not Supported: 32846/66322 (49.5%)</li><li>Warnings: 0/66322 (0.0%)</li>
</ul></details>


</ul></details>

PR #2110, opened 2022-11-01
Add SRD stride variable in resource node struct

crystalJinamd:
The SRD stride variable is used to load a descriptor with an offset value
rather than a fixed image descritor size. For example, if the client uses
full combined image sampler descriptor set layout with multi-planar
YCbCr image, the descriptor stride between each planes should be the size of
image descriptor and sampler descriptor.

The load with offset changes would be in the later commits.

### DIFF
--- a/include/vkgcDefs.h
+++ b/include/vkgcDefs.h
@@ -347,6 +347,7 @@ struct ResourceMappingNode {
     struct {
       unsigned set;     ///< Descriptor set
       unsigned binding; ///< Descriptor binding
+      unsigned stride;  ///< Descriptor stride
       unsigned reserv0;
       unsigned reserv1;
       unsigned reserv2;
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

nhaehnle (commented):
I feel like this change needs more context and assorted tests.

For example, if we just add a `stride` field like that, it's very tempting to think it's used for Vulkan descriptor arrays. But the commit title talks about YCbCr, so that's probably not what you have in mind.

It seems plausible to add this kind of field, but this really calls for a larger PR with multiple commits where the context becomes clear.

crystalJinamd:
> 

Thanks for your suggestion, I will add more detail comments in it.

crystalJinamd:
> I would like to hold off pushing this until the real changes on top of it can be reviewed.

But this commit is required to be merged firstly, then SRD stride could be set in XGL client side.

> From the conversation so far, I'm not entirely convinced that this is really necessary. In the YCbCr case, is it not possible to derive the correct stride based on the ResourceNodeType?

Yes, currently it handles DescriptorYCbCrSampler type as DescriptorResource type that also is array element type of image descriptor array. And the current stride variable is the sum of each plane descriptor size.

trenouf:
Surely there are further changes in LLPC and/or LGC to implement this new field. If we could see that in a PR, then it would become clearer why we need the new field at all.

crystalJinamd:
> Surely there are further changes in LLPC and/or LGC to implement this new field. If we could see that in a PR, then it would become clearer why we need the new field at all.

Got it, I will prepare more commits related with it. Thanks !

PR #2116, opened 2022-11-07
Fix a IR verification error of switch

amdrexu:
SPIR-V switch could have the same destination block referenced by case branch and the default branch. See this:

  %11 = OpLabel
        OpStore %value %63
        OpStore %y %float_0_5
        OpSelectionMerge %40 None
        OpSwitch %63 %39 0 %37 1 %38 2 %39
  %39 = OpLabel
  %65 = OpPhi %float %float_0_5 %11 %45 %38

One case branch and the default branch have the same destination block. In the destination block, the PHI node has two incomings from the same predecessor block. LLVM requires each incoming of PHI node must be specified. We handled such case before but didn't process default block at that time.


qiaojbao:
retest this please

ruiling (approved):
LGTM

nhaehnle (approved):
LGTM

PR #2124, opened 2022-11-09
Fixes for upstream llvm changes to function memory attributes

jayfoad:
Upstream LLVM commit 304f1d59ca41 "[IR] Switch everything to use memory attribute" changed readnone, readonly and writeonly from function attributes into parameter attributes.

### DIFF
--- a/lgc/builder/BuilderBase.cpp
+++ b/lgc/builder/BuilderBase.cpp
@@ -61,8 +61,22 @@ CallInst *BuilderCommon::CreateNamedCall(StringRef funcName, Type *retTy, ArrayR
     func->setCallingConv(CallingConv::C);
     func->addFnAttr(Attribute::NoUnwind);
 
-    for (auto attrib : attribs)
-      func->addFnAttr(attrib);
+    for (auto attrib : attribs) {
+      switch (attrib) {
+      default:
+        func->addFnAttr(attrib);
+        break;
+      case Attribute::ReadNone:
+        func->setDoesNotAccessMemory();
+        break;
+      case Attribute::ReadOnly:
+        func->setOnlyReadsMemory();
+        break;
+      case Attribute::WriteOnly:
+        func->setOnlyWritesMemory();
+        break;
+      }
+    }
   }
 
   auto call = CreateCall(func, args, instName);
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -65,6 +65,7 @@
 #include "lgc/util/AddressExtender.h"
 #include "lgc/util/BuilderBase.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Analysis/AliasAnalysis.h" // for MemoryEffects
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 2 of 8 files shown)


jayfoad:
This is my own updated and modified version of #2119.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -950,8 +951,15 @@ void PatchEntryPointMutate::setFuncAttrs(Function *entryPoint) {
...
   // NOTE: Remove "readnone" attribute for entry-point. If GS is empty, this attribute will allow
   // LLVM optimization to remove sendmsg(GS_DONE). It is unexpected.
+#if LLVM_MAIN_REVISION && LLVM_MAIN_REVISION < 440919
+  // Old version of the code
   if (entryPoint->hasFnAttribute(Attribute::ReadNone))
     entryPoint->removeFnAttr(Attribute::ReadNone);
+#else
+  // New version of the code (also handles unknown version, which we treat as
+  // latest)
+  entryPoint->setMemoryEffects(MemoryEffects::unknown());
```
jayfoad, line 961: @amdrexu does this look OK? I don't have a good understanding of this code and there is no lit test for it.

dstutt: Can you remove the REVISION check here too - or is this one required? (Based on Nicolai's comments in #2119 )
If it is, then I think that the revision number I used there was wrong. It probably requires calculating again.

jayfoad: I couldn't find an easy way to do this without an LLVM_MAIN_REVISION check. APIs like `func->setOnlyReadsMemory()` generally only let you add more restrictions to the function's memory effects. There does not seem to be an easy way to remove restrictions.

I did double check this revision number, and it corresponds exactly to the relevant upstream commit.

dstutt: Ok - thanks - I thought you'd re-used the one I did in #2119 - but I see now they're slightly different?

jayfoad: Yes it is intentionally slightly different, 440919 instead of 440909.

lgc/builder/BuilderBase.cpp
```diff
@@ -61,8 +61,22 @@ CallInst *BuilderCommon::CreateNamedCall(StringRef funcName, Type *retTy, ArrayR
...
+      case Attribute::ReadNone:
+        func->setDoesNotAccessMemory();
+        break;
+      case Attribute::ReadOnly:
+        func->setOnlyReadsMemory();
+        break;
+      case Attribute::WriteOnly:
+        func->setOnlyWritesMemory();
+        break;
+      }
```
dstutt, line 78: One reason I wasn't sure about #2119 was that this function is just hiding the fact that ReadNone, ReadOnly and WriteOnly aren't supported as function attributes any more.
I wondered if an approach that added an extra parameter for memory effects might be the cleaner way to go? Not really sure though.

jayfoad: Yes, I have been working on such a patch, but I don't think it can (easily) be done in a way that is backwards compatible with the current state of https://github.com/GPUOpen-Drivers/llvm-project/commits/amd-gfx-gpuopen-dev because all callers of CreateNamedCall would have to pass in a MemoryEffects object, but MemoryEffects does not even exist in that branch.

dstutt: Ok - makes sense.

dstutt (dismissed):
LGTM

jayfoad:
retest this please

WenqingLiAMD:
retest this please

amdvlk-admin:
### Test summary for commit d06a993bbe0b5bfdcfa6b8739dc6ec6feb69f970
<details><summary>CTS tests (Failed: 0/207247)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Ubuntu 22.04, Navi2x</summary><ul>

<li>Passed: 38905/70464 (55.2%)</li>
<li>Failed: 0/70464 (0.0%)</li>
<li>Not Supported: 31559/70464 (44.8%)</li><li>Warnings: 0/70464 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70472 (50.7%)</li>
<li>Failed: 0/70472 (0.0%)</li>
<li>Not Supported: 34722/70472 (49.3%)</li><li>Warnings: 0/70472 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66311 (50.5%)</li>
<li>Failed: 0/66311 (0.0%)</li>
<li>Not Supported: 32835/66311 (49.5%)</li><li>Warnings: 0/66311 (0.0%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchEntryPointMutate.cpp
dstutt: Sorry - should not have marked this as resolved.

PR #2126, opened 2022-11-09
Remove support for building against older LLVMs

jayfoad:
Remove workarounds for building against any LLVM source that is older than the current state of the amd-gfx-gpuopen-dev branch of https://github.com/GPUOpen-Drivers/llvm-project.


jayfoad:
This is just intended as a code cleanup. Do we actually want to keep these workarounds for any reason?

amdrexu (approved):

PR #2142, opened 2022-11-21
Opaque Pointers: Handle lack of zero-index GEPs in LowerGlobal

mariusz-sikora-at-amd:
Opaque pointers are removing zero-indexes from GEP instructions
which are not changing pointer address. These leads to elimination of
all trailing zero-indexes from GEP instruction or even deletion of
entire GEP instruction (if all indices are zero).
LLPC uses GEP's indices while doing GlobalVariable Lowering. Indices
are used to unpack GlobalVariable type and MetaData.
    
This patch is changing API of functions such as handleLoadInstGEP,
handleStoreGEP and interpolateInputElement. From now functions will accept
array of indices instead of GEP instruction (interpolateInputElement
was getting GEP internally from CallInst).
Additionally this patch is removing functions, such as
handleLoadInstGlobal and handleStoreInstGlobal. From now
handle(Load/Store)InstGEP will handle cases where global user is a load/store,
GEP or load/store but with mismatch of types of load/store and global.
Mismatch can happen if zero-index GEP was removed and global is used
directly by load/store.

### DIFF
--- a/llpc/lower/llpcSpirvLowerGlobal.cpp
+++ b/llpc/lower/llpcSpirvLowerGlobal.cpp
@@ -343,27 +343,17 @@ void SpirvLowerGlobal::handleCallInst(bool checkEmitCall, bool checkInterpCall) 
             auxInterpValue = callInst->getArgOperand(1); // Vertex no.
           }
 
-          if (isa<GetElementPtrInst>(loadSrc)) {
+          GlobalVariable *gv = nullptr;
+          SmallVector<Value *, 6> indexOperands;
+          if (auto getElemPtr = dyn_cast<GetElementPtrInst>(loadSrc)) {
             // The interpolant is an element of the input
-            interpolateInputElement(interpLoc, auxInterpValue, *callInst);
+            for (auto &index : getElemPtr->indices())
+              indexOperands.push_back(m_builder->CreateZExtOrTrunc(index, m_builder->getInt32Ty()));
+            gv = cast<GlobalVariable>(getElemPtr->getPointerOperand());
           } else {
-            // The interpolant is an input
-            assert(isa<GlobalVariable>(loadSrc));
-
-            auto input = cast<GlobalVariable>(loadSrc);
-            auto inputTy = input->getValueType();
-
-            MDNode *metaNode = input->getMetadata(gSPIRVMD::InOut);
-            assert(metaNode);
-            auto inputMeta = mdconst::dyn_extract<Constant>(metaNode->getOperand(0));
-
-            m_builder->SetInsertPoint(callInst);
-            auto loadValue = addCallInstForInOutImport(inputTy, SPIRAS_Input, inputMeta, nullptr, 0, nullptr, nullptr,
-                                                       interpLoc, auxInterpValue, false);
-
-            m_interpCalls.insert(callInst);
-            callInst->replaceAllUsesWith(loadValue);
+            gv = cast<GlobalVariable>(loadSrc);
           }
+          interpolateInputElement(interpLoc, auxInterpValue, *callInst, gv, indexOperands);
         }
       }
     }
@@ -411,70 +401,24 @@ static bool hasPrimitiveIdx(const Constant &metaVal) {
 }
 
 // =====================================================================================================================
-// Handle a single "load" instruction directly loading a global.
+// Handle a single "load" instruction loading a global.
 //
+// @param inOut : Global Variable instruction
+// @param indexOperands : Indices of GEP instruction
 // @param loadInst : Load instruction
-// @param addrSpace : Address space
-void SpirvLowerGlobal::handleLoadInstGlobal(LoadInst &loadInst, const unsigned addrSpace) {
-  Value *loadSrc = loadInst.getOperand(0);
-  auto inOut = cast<GlobalVariable>(loadSrc);
-  auto inOutTy = inOut->getValueType();
+void SpirvLowerGlobal::handleLoadInstGEP(GlobalVariable *inOut, ArrayRef<Value *> indexOperands, LoadInst &loadInst) {
 
-  const bool isTaskPayload = addrSpace == SPIRAS_TaskPayload;
-  MDNode *metaNode = inOut->getMetadata(isTaskPayload ? gSPIRVMD::Block : gSPIRVMD::InOut);
-  assert(metaNode);
-  auto inOutMetaVal = mdconst::dyn_extract<Constant>(metaNode->getOperand(0));
+  assert(indexOperands.empty() || cast<ConstantInt>(indexOperands.front())->isZero() && "Non-zero GEP first index\n");
+  if (!indexOperands.empty())
+    indexOperands = indexOperands.drop_front();
 
   m_builder->SetInsertPoint(&loadInst);
-
-  Value *loadValue = UndefValue::get(inOutTy);
-
-  // If the input/output is arrayed, the outermost index might be used for vertex indexing
-  if (!isTaskPayload && inOutTy->isArrayTy() && hasVertexIdx(*inOutMetaVal)) {
-    auto elemTy = inOutTy->getArrayElementType();
-    auto elemMeta = cast<Constant>(inOutMetaVal->getOperand(1));
-
-    const unsigned elemCount = inOutTy->getArrayNumElements();
-    for (unsigned i = 0; i < elemCount; ++i) {
-      Value *vertexIdx = m_builder->getInt32(i);
-      auto elemValue = addCallInstForInOutImport(elemTy, addrSpace, elemMeta, nullptr, 0, nullptr, vertexIdx,
-                                                 InterpLocUnknown, nullptr, false);
-      loadValue = m_builder->CreateInsertValue(loadValue, elemValue, {i});
-    }
-  } else if (isTaskPayload) {
-    loadValue = loadValueFromTaskPayload(inOutTy, inOutMetaVal, nullptr);
-  } else {
-    loadValue = addCallInstForInOutImport(inOutTy, addrSpace, inOutMetaVal, nullptr, 0, nullptr, nullptr,
-                                          InterpLocUnknown, nullptr, false);
-  }
-  m_loadInsts.insert(&loadInst);
-  loadInst.replaceAllUsesWith(loadValue);
-}
-
-// =====================================================================================================================
-// Handle a single "load" instruction loading a global through a GEP instruction
-//
-// @param getElemPtr : Load source GEP instruction
-// @param loadInst : Load instruction
-// @param addrSpace : Address space
-void SpirvLowerGlobal::handleLoadInstGEP(GetElementPtrInst *const getElemPtr, LoadInst &loadInst,
-                                         const unsigned addrSpace) {
-  std::vector<Value *> indexOperands;
-
-  assert(cast<ConstantInt>(getElemPtr->getOperand(1))->isZero() && "Non-zero GEP first index\n");
-
-  GlobalVariable *inOut = cast<GlobalVariable>(getElemPtr->getPointerOperand());
-  assert(!isa<GetElementPtrInst>(getElemPtr->getPointerOperand()) &&
-         "Chained GEPs should have been coalesced by SpirvLowerAccessChain.");
-
-  m_builder->SetInsertPoint(&loadInst);
-
-  for (auto &index : drop_begin(getElemPtr->indices()))
-    indexOperands.push_back(m_builder->CreateZExtOrTrunc(index, m_builder->getInt32Ty()));
 
   Value *vertexIdx = nullptr;
   auto inOutTy = inOut->getValueType();
 
+  auto addrSpace = inOut->getType()->getPointerAddressSpace();
+
   const bool isTaskPayload = addrSpace == SPIRAS_TaskPayload;
   MDNode *metaNode = inOut->getMetadata(isTaskPayload ? gSPIRVMD::Block : gSPIRVMD::InOut);
   assert(metaNode);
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

nhaehnle:
ok to test

llpc/lower/llpcSpirvLowerUtil.cpp
```diff
@@ -87,4 +88,32 @@ void setShaderStageToModule(Module *module, ShaderStage shaderStage) {
...
+//
+// This function is used to workaround the elimination of zero-index GEP instructions which is taking place
+// when opaque pointers are enabled.
+//
+// @param builder : LGC Builder
+// @param indexOperands : vector to which zero-index elements will be added
+// @param typeToMatch : type used as destination of unpacking "packType"
+// @param packType : packed type which will be unpacked.
+void appendZeroIndexToMatchTypes(lgc::Builder *builder, SmallVectorImpl<Value *> &indexOperands, Type *typeToMatch,
+                                 Type *packType) {
```
nhaehnle, line 102: Rename `packType` to `baseType`?

llpc/lower/llpcSpirvLowerGlobal.h
```diff
@@ -54,11 +54,12 @@ class SpirvLowerGlobal : public SpirvLower, public llvm::PassInfoMixin<SpirvLowe
 
   void handleLoadInst();
   void handleLoadInstGlobal(LoadInst &loadInst, const unsigned addrSpace);
-  void handleLoadInstGEP(GetElementPtrInst *const getElemPtr, LoadInst &loadInst, const unsigned addrSpace);
+  void handleLoadInstGEP(SmallVectorImpl<Value *> &indexOperands, LoadInst &loadInst, const unsigned addrSpace,
+                         GlobalVariable *inOut);
```
nhaehnle, line 58: Please keep related operands close together. I would suggest:
```
void handleLoadInstGEP(GlobalVariable *inOut, SmallVectorImpl<Value *> &indexOperands, LoadInst &loadInst);
```
(addrSpace is implied by the GlobalVariable, isn't it?)

llpc/lower/llpcSpirvLowerGlobal.h
```diff
@@ -54,11 +54,12 @@ class SpirvLowerGlobal : public SpirvLower, public llvm::PassInfoMixin<SpirvLowe
...
 
   void handleStoreInst();
   void handleStoreInstGlobal(StoreInst &storeInst);
-  void handleStoreInstGEP(GetElementPtrInst *const getElemPtr, StoreInst &storeInst);
+  void handleStoreInstGEP(SmallVectorImpl<Value *> &indexOperands, StoreInst &storeInst, GlobalVariable *output);
```
nhaehnle, line 62: Same here, re-order the operands.

llpc/lower/llpcSpirvLowerGlobal.h
```diff
@@ -116,7 +117,8 @@ class SpirvLowerGlobal : public SpirvLower, public llvm::PassInfoMixin<SpirvLowe
   llvm::Value *atomicOpWithValueInTaskPayload(llvm::Instruction *atomicInst, llvm::Constant *metadata,
                                               llvm::Value *extraByteOffset);
 
-  void interpolateInputElement(unsigned interpLoc, llvm::Value *interpInfo, llvm::CallInst &callInst);
+  void interpolateInputElement(unsigned interpLoc, llvm::Value *interpInfo, llvm::CallInst &callInst,
+                               SmallVectorImpl<Value *> &indexOperands, GlobalVariable *gv);
```
nhaehnle, line 121: For consistency, gv before indexOperands.

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -343,26 +343,42 @@ void SpirvLowerGlobal::handleCallInst(bool checkEmitCall, bool checkInterpCall)
...
+              callInst->replaceAllUsesWith(loadValue);
+            } else {
+              // Type of GlobalVariable and return type of the Call are not the same. This can happen for opaque
+              // pointers because zero-index GEP instructions are removed. In this case we have to build zero-index
+              // vector which is equivalent to zero-index GEP and use "interpolateInputElement".
+              SmallVector<Value *, 6> indexOperands;
+              indexOperands.push_back(m_builder->getInt32(0));
+              appendZeroIndexToMatchTypes(m_builder, indexOperands, funcRetType, inputTy);
+              interpolateInputElement(interpLoc, auxInterpValue, *callInst, indexOperands, input);
+            }
```
nhaehnle, line 381: In the spirit of leaving the code in a slightly better state than we found it in, please refactor this into a code sequence that calls `interpolateInputElement` in only one place (unified for the GEP and non-GEP case) and never calls `addCallInstForInOutImport` directly (let `interpolateInputElement` handle that for us).

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -454,23 +470,17 @@ void SpirvLowerGlobal::handleLoadInstGlobal(LoadInst &loadInst, const unsigned a
...
+// @param inOut : Global Variable instruction
+void SpirvLowerGlobal::handleLoadInstGEP(SmallVectorImpl<Value *> &indexOperands, LoadInst &loadInst,
+                                         const unsigned addrSpace, GlobalVariable *inOut) {
+  assert(cast<ConstantInt>(indexOperands.front())->isZero() && "Non-zero GEP first index\n");
 
   m_builder->SetInsertPoint(&loadInst);
 
-  for (auto &index : drop_begin(getElemPtr->indices()))
-    indexOperands.push_back(m_builder->CreateZExtOrTrunc(index, m_builder->getInt32Ty()));
+  indexOperands.erase(indexOperands.begin());
```
nhaehnle, line 483: Move this to be closer to the assert above.

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -524,15 +534,40 @@ void SpirvLowerGlobal::handleLoadInst() {
...
+              indexOperands.push_back(m_builder->CreateZExtOrTrunc(index, m_builder->getInt32Ty()));
+            Type *gepRetType = GetElementPtrInst::getIndexedType(gep->getSourceElementType(), indexOperands);
+            // Check if a Load type is the same as indexed type of GEP instruction. Types can be different if opaque
+            // pointers are enabled. Opaque pointers are removing zero-index elements which are at the end of GEP
+            // instruction. These additional zero-index elements are needed to proper unpack metadata for global
+            // variable.
+            if (loadInst->getType() != gepRetType)
+              appendZeroIndexToMatchTypes(m_builder, indexOperands, loadInst->getType(), gepRetType);
+            handleLoadInstGEP(indexOperands, *loadInst, addrSpace, &global);
+          }
```
nhaehnle, line 570: The first branch of the outer if absolutely needs braces.

nhaehnle (changes requested):
I started a review and left some inline comments, but taking a step back I think this is the wrong high-level design here.

I have two important general rules:

1. When we touch code, we should try to leave it in at least a slightly better state than it was before.
2. There is often a temptation to fix an issue or to implement a feature by adding something on top of an already existing abstraction. However, it is very often better to instead re-mold the existing abstraction to better fit the problem.

In this particular case, we have existing abstractions (`handleLoadInstGEP` etc.) that were built around a world of typed pointers and have certain expectations because of it. As-is, this PR tries to keep the abstractions the same, or at least as close to how they were before, and adds stuff on top to maintain the illusion of typed pointers. This will be confusing in the future.

Instead, let us embrace opaque pointers and the fact that 0-indexing is "for free". This means:

* No `appendZeroIndexToMatchTypes`
* However, `handleLoadInstGEP` etc. are still changed to take (`ArrayRef`s of) index operands.
* Re-mold the call sites to prefer calling the GEP-variants of functions, even if that means calling it with an empty index list. For example, in `handleLoadInst` call `handleLoadInstGEP` with empty operands instead of `handleLoadInstGlobal`. Use opportunities to remove code by unifying the paths.
* Change `loadInOutMember` to handle the case where the index operands are empty but the `loadTy` doesn't match the `inOutTy` (this requires adding a new parameter to `loadInOutMember`). Do the analogous change in the store and interpolate paths.

amdvlk-admin:
### Test summary for commit 75d133e25f62c97a46b1f9c8701e55a2f5ed7332
<details><summary>CTS tests (Failed: 0/207274)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Ubuntu 22.04, Navi2x</summary><ul>

<li>Passed: 38905/70473 (55.2%)</li>
<li>Failed: 0/70473 (0.0%)</li>
<li>Not Supported: 31568/70473 (44.8%)</li><li>Warnings: 0/70473 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70481 (50.7%)</li>
<li>Failed: 0/70481 (0.0%)</li>
<li>Not Supported: 34731/70481 (49.3%)</li><li>Warnings: 0/70481 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66320 (50.5%)</li>
<li>Failed: 0/66320 (0.0%)</li>
<li>Not Supported: 32844/66320 (49.5%)</li><li>Warnings: 0/66320 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit 48f97d3170b35d278121e27fe60867a60e1a4192
<details><summary>CTS tests (Failed: 0/207274)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Ubuntu 22.04, Navi2x</summary><ul>

<li>Passed: 38905/70473 (55.2%)</li>
<li>Failed: 0/70473 (0.0%)</li>
<li>Not Supported: 31568/70473 (44.8%)</li><li>Warnings: 0/70473 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70481 (50.7%)</li>
<li>Failed: 0/70481 (0.0%)</li>
<li>Not Supported: 34731/70481 (49.3%)</li><li>Warnings: 0/70481 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66320 (50.5%)</li>
<li>Failed: 0/66320 (0.0%)</li>
<li>Not Supported: 32844/66320 (49.5%)</li><li>Warnings: 0/66320 (0.0%)</li>
</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -522,81 +470,42 @@ void SpirvLowerGlobal::handleLoadInst() {
     if (!shouldHandle(addrSpace))
       continue;
     for (User *user : global.users()) {
-      if (LoadInst *loadInst = dyn_cast<LoadInst>(user))
-        // The user is a load
-        handleLoadInstGlobal(*loadInst, addrSpace);
-      else if (GetElementPtrInst *gep = dyn_cast<GetElementPtrInst>(user)) {
+      if (LoadInst *loadInst = dyn_cast<LoadInst>(user)) {
+        handleLoadInstGEP(&global, ArrayRef<Value *>(), *loadInst);
```
nhaehnle, line 474: I would expect `{}` to work instead of the explicit ArrayRef.

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -2419,6 +2377,15 @@ void SpirvLowerGlobal::lowerBufferBlock() {
...
 
+              // Check if Global Variable type is the same as GEP source type. For opaque pointers these types can be
+              // different because opaque pointers are removing zero-index GEP instructions. Without opaque pointers
+              // LLVM-IR would contain additional zero-index GEP which will be between "getElemPtr" and "global". These
+              // two GEP instructions would be at some point merged to single one.
+              if (getElemPtr->getSourceElementType() != global.getValueType()) {
+                appendZeroIndexToMatchTypes(m_builder, indices, getElemPtr->getSourceElementType(),
+                                            global.getValueType());
+              }
+
```
nhaehnle, line 2388: Is this really necessary? The indices are really only used to create a new GEP, and the new GEP doesn't need trailing zero indices with opaque pointers.

The only case to watch out for is if the block indices themselves are missing, but that case could be handled by not inserting a new GEP at all -- right?

nhaehnle (commented):
Thank you, this looks so much better. I do have two notes inline, still.

llpc/lower/llpcSpirvLowerGlobal.cpp
mariusz-sikora-at-amd: You covered all cases why I did it, but this change is for leading zeros, not for trailing.

Type and indices of GEP are different for opaque pointer and since Block Index is equal to GEP's indices[1], I had to add these leading zeros. Adding leading zeros also solved issue with new GEP which was based on Global Variable Type.

For new GEP I could used existing indices (without adding leading zeros) and change Type of the GEP while creating, but not sure if I can assume that Block Index == 0 if Types are different for oldGEP and GV ?

amdvlk-admin:
### Test summary for commit 48f97d3170b35d278121e27fe60867a60e1a4192
<details><summary>CTS tests (Failed: 0/207274)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Ubuntu 22.04, Navi2x</summary><ul>

<li>Passed: 38905/70473 (55.2%)</li>
<li>Failed: 0/70473 (0.0%)</li>
<li>Not Supported: 31568/70473 (44.8%)</li><li>Warnings: 0/70473 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70481 (50.7%)</li>
<li>Failed: 0/70481 (0.0%)</li>
<li>Not Supported: 34731/70481 (49.3%)</li><li>Warnings: 0/70481 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66320 (50.5%)</li>
<li>Failed: 0/66320 (0.0%)</li>
<li>Not Supported: 32844/66320 (49.5%)</li><li>Warnings: 0/66320 (0.0%)</li>
</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerGlobal.cpp
nhaehnle: Ah, thank you for the explanation.

If the oldGEP base type is different from the GV type, then yes, I think you can assume that the block index is 0 and form the newGEP with the same base type as oldGEP.

amdvlk-admin:
### Test summary for commit 8afec56ab763bb21dc38fd11dd30e5c8c0a85df5
<details><summary>CTS tests (Failed: 0/207253)</summary><ul><li>Built with version 1.3.0.0</li>
<details open><summary>Ubuntu 22.04, Navi2x</summary><ul>

<li>Passed: 38905/70466 (55.2%)</li>
<li>Failed: 0/70466 (0.0%)</li>
<li>Not Supported: 31561/70466 (44.8%)</li><li>Warnings: 0/70466 (0.0%)</li>
</ul></details>

<li>Passed: 35750/70474 (50.7%)</li>
<li>Failed: 0/70474 (0.0%)</li>
<li>Not Supported: 34724/70474 (49.3%)</li><li>Warnings: 0/70474 (0.0%)</li>
</ul></details>

<li>Passed: 33476/66313 (50.5%)</li>
<li>Failed: 0/66313 (0.0%)</li>
<li>Not Supported: 32837/66313 (49.5%)</li><li>Warnings: 0/66313 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
Thank you!

PR #2137, opened 2022-11-17
Rework of TF store

amdrexu:
Defer TF store from the phase of in/out handling to the phase of preparing pipeline ABI. Add two functions readTessFactors and writeTessFactors. These functions will be used by future TF optimization in shader merger as well.

The basic logic of TF store is not changed. Just some coding refactoring.

PR #2143, opened 2022-11-22
Fix the regression of PR#2137

amdrexu:
The PR (https://github.com/GPUOpen-Drivers/llpc/pull/2137) has a latent issue of reading TF from on-chip LDS. We mistakenly use CreateLoad which always assumes the alignment is from the specified load type. Rather, we can only ensure the alignment is dword. So CreateAlignedLoad should be used instead to explicitly specify dword alignment.


qiaojbao:
retest this please

PR #2263, opened 2023-02-06
Fix latent issues of mesh shader implementation

amdrexu:
1. The enableSwXfb function should return false if mesh pipeline is found. XFB is only applied to traditional graphics pipeline.

2. When checking hasNoVertexAttrib in PS reg config, we should rely on the exsitence of per-vertex inputs in fragment shader. This is because we could encounter part pipeline compilation of mesh pipeline. In such cases, fragment shader is compiled without the knowlegde of previous stages.

3. The ready bit of draw data is supposed to be written by task shader with the lowest 8 bits of the dword (last dword). Other bits of the dword have other purposes.

4. Task payload could have such usage:

     taskpayloadSharedEXT uint outVar[32];

   In this case, we incorrectly handle the array dimension of task payload. The original logic is to handle block array by eliminating its array dimension (see the block array as folow):

     layout(binding = 0) buffer Data {
       uint outVar;
     } data[32];

   For task payload, this is meaningless since task payload doesn't defined as block array.

### DIFF
--- a/lgc/patch/Gfx9ConfigBuilder.cpp
+++ b/lgc/patch/Gfx9ConfigBuilder.cpp
@@ -1648,16 +1648,14 @@ template <typename T> void ConfigBuilder::buildPsRegConfig(ShaderStage shaderSta
     spiPsInputCntl.bits.FLAT_SHADE = interpInfoElem.flat && !interpInfoElem.isPerPrimitive;
     spiPsInputCntl.bits.OFFSET = interpInfoElem.loc;
     if (gfxIp.major >= 11 && interpInfoElem.isPerPrimitive) {
-      const auto preStage = m_pipelineState->getPrevShaderStage(ShaderStageFragment);
-      if (preStage == ShaderStageMesh) {
-        // NOTE: HW allocates and manages attribute ring based on the register fields: VS_EXPORT_COUNT and
-        // PRIM_EXPORT_COUNT. When VS_EXPORT_COUNT = 0, HW assumes there is still a vertex attribute exported even
-        // though this is not what we want. Hence, we should reserve param0 as a dummy vertex attribute and all
-        // primitive attributes are moved after it.
-        bool hasNoVertexAttrib = m_pipelineState->getShaderResourceUsage(ShaderStageMesh)->inOutUsage.expCount == 0;
-        if (hasNoVertexAttrib)
-          ++spiPsInputCntl.bits.OFFSET;
-      }
+      // NOTE: HW allocates and manages attribute ring based on the register fields: VS_EXPORT_COUNT and
+      // PRIM_EXPORT_COUNT. When VS_EXPORT_COUNT = 0, HW assumes there is still a vertex attribute exported even
+      // though this is not what we want. Hence, we should reserve param0 as a dummy vertex attribute and all
+      // primitive attributes are moved after it.
+      const bool hasNoVertexAttrib = resUsage->inOutUsage.inputMapLocCount == 0; // No vertex attribute
+      if (hasNoVertexAttrib)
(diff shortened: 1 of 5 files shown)


amdvlk-admin:
### Test summary for commit 93f990819b27a73150e658b643400d8a01dbf015
<details><summary>CTS tests (Failed: 0/211498)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 39250/70506 (55.7%)</li>
<li>Failed: 0/70506 (0.0%)</li>
<li>Not Supported: 31256/70506 (44.3%)</li><li>Warnings: 0/70506 (0.0%)</li>
</ul></details>

<li>Passed: 39498/70495 (56.0%)</li>
<li>Failed: 0/70495 (0.0%)</li>
<li>Not Supported: 30997/70495 (44.0%)</li><li>Warnings: 0/70495 (0.0%)</li>
</ul></details>

<li>Passed: 36345/70497 (51.6%)</li>
<li>Failed: 0/70497 (0.0%)</li>
<li>Not Supported: 34152/70497 (48.4%)</li><li>Warnings: 0/70497 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit f6ab23ecbfdf4c41881abf62ef991aa07fa9c479
<details><summary>CTS tests (Failed: 0/211498)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 39250/70506 (55.7%)</li>
<li>Failed: 0/70506 (0.0%)</li>
<li>Not Supported: 31256/70506 (44.3%)</li><li>Warnings: 0/70506 (0.0%)</li>
</ul></details>

<li>Passed: 39498/70495 (56.0%)</li>
<li>Failed: 0/70495 (0.0%)</li>
<li>Not Supported: 30997/70495 (44.0%)</li><li>Warnings: 0/70495 (0.0%)</li>
</ul></details>

<li>Passed: 36345/70497 (51.6%)</li>
<li>Failed: 0/70497 (0.0%)</li>
<li>Not Supported: 34152/70497 (48.4%)</li><li>Warnings: 0/70497 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (commented):
> The ready bit of draw data is supposed to be written by task shader with the lowest 8 bits of the dword (last dword). Other bits of the dword have other purposes.

I don't understand that part of it. What are the other 24 bits used for?

The initial design was specifically made such that the entire 16 byte block could be written atomically with a single instruction.

If we can no longer do that, that's a potentially significant performance regression and needs to be understood.

The rest looks good to me.

amdrexu:
> I don't understand that part of it. What are the other 24 bits used for?
> 
> The initial design was specifically made such that the entire 16 byte block could be written atomically with a single instruction.
> 
> If we can no longer do that, that's a potentially significant performance regression and needs to be understood.

Sorry. I check the CP doc/code and didn't find clear reason. It is a GFX11 fix. I will send an internal mail.

nhaehnle (approved):
After the internal clarification, I think we should just go ahead.

One annoying subtle aspect here is that the readyBit write needs to "make visible" the groupCount store. So, in a clean representation in terms of the memory model, technically the readyBit store would need atomic release semantics, and the fact that both stores touch the same cacheline could be used to prevent the need for a waitcnt -- but the backend doesn't understand that, and in any case there's also the statistics atomic.

In practice, as long as we're using intrinsics for the store, the important bit is that the stores don't get reordered, but that should be the case in practice.

amdrexu:
> the important bit is that the stores don't get reordered, but that should be the case in practice

Thank you for pointing this out. I didn't realize this previously :)

PR #2267, opened 2023-02-08
Opaque Pointers: Remove support for typed pointers from LLPC/LGC

mariusz-sikora-at-amd:
- Remove flag (-enable-opaque-pointers) which was created for transition purpose.
- Remove usage of enable-opaque-pointers flag from LLPC and LGC tests.
- Opaque Pointers state in LLVM Context is no longer overrid (it was overrid by -enable-opaque-pointers flag).
- Remove last remaining getPointerElementType used for mangling function names. This will remove compilation warnings about getPointerElementType being deprecated.
- Remove assertions used for typed pointers to validate correctnes of types while removing getPointerElementType from LLPC/LGC.


nhaehnle (commented):
Nice.

In terms of merging effort it may be better to split the removal of the -enable-opaque-pointers option into a separate change. Just initially make it so the compiler errors out if the option is overridden to false? But maybe I'm just being overly paranoid.

What's the reason for this only being a draft so far?

mariusz-sikora-at-amd:
> Nice.
> 
> In terms of merging effort it may be better to split the removal of the -enable-opaque-pointers option into a separate change. Just initially make it so the compiler errors out if the option is overridden to false? But maybe I'm just being overly paranoid.

Sure, I will split the change.


> What's the reason for this only being a draft so far?

I was testing changes in internal repo and found some "issues".
Not all internal branches are up to date with llvm. In one branch we are missing https://github.com/llvm/llvm-project/commit/e45cf479231fc144b4e1e and it is causing tests (like: https://github.com/GPUOpen-Drivers/llpc/blob/dev/lgc/test/TextureRange.lgc) to fail when I'm removing 'enable-opaque-pointers' flag. Probably I will also re-write lgc tests to use opaque pointers in llvm-ir instead of typed pointers.
I will start pushing changes tomorrow.

Additionally I see that LGC test https://github.com/GPUOpen-Drivers/llpc/blob/dev/lgc/test/TextureRange.lgc#L9 is expecting `bitcast ptr to ptr` which I thought should be removed in opaque pointers.

piotrAMD:
> Probably I will also re-write lgc tests to use opaque pointers in llvm-ir instead of typed pointers. I will start pushing changes tomorrow.

Agreed that rewriting lgc tests to not use typed pointers is the way to go.

mariusz-sikora-at-amd:
Closing this draft. I created couple PRs based on this single PR.
Rewritting LGC tests is in progress.

PR #2297, opened 2023-02-20
Fix Visual Studio solution generation with CMake

yavn:
This stopped working due to a CMake change in LLVM.

Let's keep both paths, so that it can work with old and new versions of LLVM.

I also removed `LLVM_BIN_DIR`, it is not referenced anywhere.

### DIFF
--- a/llpc/CMakeLists.txt
+++ b/llpc/CMakeLists.txt
@@ -144,7 +144,10 @@ set_compiler_options(llpc ${LLPC_ENABLE_WERROR})
 
 ### Defines/Includes/Sources ###########################################################################################
 if(ICD_BUILD_LLPC)
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${XGL_LLVM_BUILD_PATH}/${CMAKE_CFG_INTDIR}/lib/cmake/llvm")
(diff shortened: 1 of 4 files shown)


Flakebi (dismissed):
Thanks. LGTM

Flakebi:
There are a couple more references to the old `CMAKE_CFG_INTDIR` path in `test/CMakeLists.txt`, `llpc/unittests/CMakeLists.txt` and `llpc/test/CMakeLists.txt`.
The ones in `llpc/test/` can probably be removed alltogether, but the path in `test/CMakeLists.txt` probably needs fixing as well.

Flakebi:
Yes, that would be nice. Thanks!

trenouf (approved):
LGTM. Happy for you to merge this and leave those other changes Sebastian pointed out for a follow-up change.

yavn:
Sorry, just pushed again. Indeed it was failing with -DXGL_BUILD_TESTS=ON. Now it should be fine.

trenouf (approved):
Still LGTM.

yavn:
OK, looks like it can be merged. I don't have an access to do it, please go ahead.

PR #2307, opened 2023-02-22
Enable sample shading whenever gl_sampleId/gl_samplePostion is declared

AMD-dwang:
According to the vulkan spec, sample shading is enabled if the fragment

shader's entry point interface includes input variables decorated with a

BuiltIn of SampleId or SamplePosition built-ins.

If gl_sampleId or gl_samplePosition is declared but not used in a fragment

shader, it will be removed in lower stage, in those cases, we need to add

metadata and use it later.

### DIFF
--- a/lgc/patch/PatchResourceCollect.cpp
+++ b/lgc/patch/PatchResourceCollect.cpp
@@ -1134,9 +1134,9 @@ void PatchResourceCollect::processShader() {
   }
 
   if (m_shaderStage == ShaderStageFragment) {
-    if (m_resUsage->builtInUsage.fs.fragCoord || m_resUsage->builtInUsage.fs.pointCoord ||
-        m_resUsage->builtInUsage.fs.sampleMaskIn) {
-      if (m_pipelineState->getRasterizerState().perSampleShading)
+    if (m_pipelineState->getRasterizerState().perSampleShading) {
+      if (m_resUsage->builtInUsage.fs.fragCoord || m_resUsage->builtInUsage.fs.pointCoord ||
+          m_resUsage->builtInUsage.fs.sampleMaskIn || m_resUsage->resourceWrite)
         m_resUsage->builtInUsage.fs.runAtSampleRate = true;
     }
 
--- a/lgc/interface/lgc/Pipeline.h
+++ b/lgc/interface/lgc/Pipeline.h
@@ -107,6 +107,7 @@ enum class ThreadGroupSwizzleMode : unsigned {
 static const unsigned ShadowDescriptorTableDisable = ~0U;
 
 static const char XfbStateMetadataName[] = "lgc.xfb.state";
+static const char SampleShadingMetaName[] = "lgc.sample.shading";
 
 // Middle-end per-pipeline options to pass to SetOptions.
 // The front-end should zero-initialize it with "= {}" in case future changes add new fields.
--- a/lgc/state/PipelineState.cpp
+++ b/lgc/state/PipelineState.cpp
@@ -1262,6 +1262,10 @@ void PipelineState::recordGraphicsState(Module *module) {
 void PipelineState::readGraphicsState(Module *module) {
   readNamedMetadataArrayOfInt32(module, IaStateMetadataName, m_inputAssemblyState);
   readNamedMetadataArrayOfInt32(module, RsStateMetadataName, m_rasterizerState);
+
+  auto nameMeta = module->getNamedMetadata(SampleShadingMetaName);
+  if (nameMeta)
+    m_rasterizerState.perSampleShading |= 1;
 }
 
 // =====================================================================================================================
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -7692,6 +7692,19 @@ bool SPIRVToLLVM::transShaderDecoration(SPIRVValue *bv, Value *v) {
       if (bv->hasDecorate(DecorationBuiltIn, 0, &builtIn)) {
         inOutDec.IsBuiltIn = true;
         inOutDec.Value.BuiltIn = builtIn;
+
+        // NOTE: According to the vulkan spec, sample shading is enabled if the fragment shader's entry point
+        // interface includes input variables decorated with a BuiltIn of SampleId or SamplePosition built-ins.
+        // If gl_sampleId or gl_samplePosition is declared but not used in a fragment shader, it will be removed
+        // in lower stage, in those cases, we need to add metadata and use it later.
+        if ((m_execModule == ExecutionModelFragment) &&
+            (builtIn == spv::BuiltInSampleId || builtIn == spv::BuiltInSamplePosition)) {
+          auto nameMeta = m_m->getNamedMetadata(lgc::SampleShadingMetaName);
+          if (!nameMeta) {
+            nameMeta = m_m->getOrInsertNamedMetadata(lgc::SampleShadingMetaName);
+            nameMeta->addOperand(MDNode::get(*m_context, MDString::get(*m_context, lgc::SampleShadingMetaName)));
+          }
+        }
 #if VKI_RAY_TRACING
         Llpc::Context *llpcContext = static_cast<Llpc::Context *>(m_context);
         llpcContext->getPipelineContext()->collectBuiltIn(builtIn);


amdvlk-admin:
### Test summary for commit 0b64f0ff917eddadee464991faf4e7ae7dd32779
<details><summary>CTS tests (Failed: 0/205094)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36341/68149 (53.3%)</li>
<li>Failed: 0/68149 (0.0%)</li>
<li>Not Supported: 31808/68149 (46.7%)</li><li>Warnings: 0/68149 (0.0%)</li>
</ul></details>

<li>Passed: 36569/68149 (53.7%)</li>
<li>Failed: 0/68149 (0.0%)</li>
<li>Not Supported: 31580/68149 (46.3%)</li><li>Warnings: 0/68149 (0.0%)</li>
</ul></details>

<li>Passed: 34225/68796 (49.7%)</li>
<li>Failed: 0/68796 (0.0%)</li>
<li>Not Supported: 34571/68796 (50.3%)</li><li>Warnings: 0/68796 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle:
I don't understand why this change is needed and I don't think it is correct as-is.

The current logic is that `runAtSampleRate` is set if `perSampleShading` is requested **and** the shader contains code that causes the use of sample rate to be observable.

Sample rate is only observable if sample ID or position is used directly or indirectly. The only place I can see where this happens indirectly is for FragCoord/PointCoord. But there, in order to actually generate correct code it is guarded by `perSampleShading`.

So if anything, the correct thing to do seems to be to set the `perSampleShading` boolean.

But maybe I'm still lacking some context?

AMD-dwang:
> I don't understand why this change is needed and I don't think it is correct as-is.
> 
> The current logic is that `runAtSampleRate` is set if `perSampleShading` is requested **and** the shader contains code that causes the use of sample rate to be observable.
> 
> Sample rate is only observable if sample ID or position is used directly or indirectly. The only place I can see where this happens indirectly is for FragCoord/PointCoord. But there, in order to actually generate correct code it is guarded by `perSampleShading`.
> 
> So if anything, the correct thing to do seems to be to set the `perSampleShading` boolean.
> 
> But maybe I'm still lacking some context?

Please see https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#primsrast-sampleshading.
Some tests are using the following shader and have not set the `perSampleShading`,  In these cases, we expect to enable sample shading as well.
```
#version 450
layout (location = 0) out vec4 outColor;
layout (std430, binding = 0) buffer Output {
    uint invocationCount;
} buf;
void main() {
    gl_SampleID;
    uint index = atomicAdd(buf.invocationCount, 1);
    outColor = vec4(1.0, 1.0, 0.0, 1.0);
}
```

nhaehnle:
Interesting, thank you. I assume this is from a CTS test? Could you please add the name(s) of the relevant tests to the commit message?

I've tried to dig into this further and it seems to me there are more issues here around sample shading enable. Let's please double-check them.

We enable multiple PS invocations via the `PS_ITER_SAMPLE` field, which is set based on `builtInUsage.runAtSampleRate`. There are some suspicious factors in when that field is set. Section 28.8 (Sample Shading) of the Vulkan spec lists only two conditions under which there should be multiple PS invocations:

> Sample shading is enabled if at least one of the following conditions is true:
>
> *  [VkPipelineMultisampleStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineMultisampleStateCreateInfo)::sampleShadingEnable is set to VK_TRUE, or
> * the fragment shader’s entry point interface includes input variables decorated with a BuiltIn of SampleId or SamplePosition built-ins.

If either of those conditions is true, running multiple pixel shaders is potentially observable, if the pixel shader uses external stores/atomics like in the example you cited. But if neither of those conditions is true, it seems to me as if only a single pixel shader invocation should be used, *even if there are inputs with a `Sample` location*(!).

The logic in LLPC and LGC doesn't match that.

I suspect that what we should be doing is:

1. Set `lgc::RasterizerState::perSampleShading` based on the two conditions from the Vulkan spec listed above. This would be the job of the frontend, since it needs to pass along the pipeline creation state (which it already does) and needs to interpret the entry point interface (which your change is adding, but I think that part should be limited to the frontend).
2. Determine whether per-sample shading is observable or not. It is observable if inputs with `Sample` interpolation location or builtins such as SampleID are used; or if the pixel shader has side effects (`resUsage::resourceWrite`). This part belongs in LGC and more or less matches how we set `ResourceUsage::fs.runAtSampleRate` today, but we should probably just remove that and instead combine the existing `ResourceUsage::fs::{sample,sampleId,...}` fields.
3. LGC should then combine these pieces of information. Only if both are true do we set `PS_ITER_SAMPLE`. This would happen in the config builder (presumably with a `shouldRunAtSampleRate` helper that is shared between the different config builders)

There may be something I am still missing, but this does seem to match what the spec says more closely.

Finally, there is yet another potential issue here. The Vulkan space (also in section 28.8) says:

> If sample shading is enabled, an implementation must invoke the fragment shader at least max(⌈ [VkPipelineMultisampleStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineMultisampleStateCreateInfo)::minSampleShading × [VkPipelineMultisampleStateCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineMultisampleStateCreateInfo)::rasterizationSamples ⌉, 1) times per fragment. If a fragment shader entry point’s interface includes an input variable decorated with a BuiltIn of SampleId or SamplePosition, a value of 1.0 is used instead of minSampleShading.

What I believe this means is that we also need to change the value of `PS_ITER_SAMPLES` (note the 'S'), which the compiler currently doesn't control. It is currently set via `msaaCreateInfo.pixelShaderSamples` when creating a PAL pipeline.

Right?

AMD-dwang:
After some [discussion](https://gerrit.khronos.org/c/vk-gl-cts/+/10719/comment/d9010bc4_78420876/), we still need to implement it.

Sample shading should contain two cases:
1.  Set lgc::RasterizerState::perSampleShading to true in frontend, we need to determine whether per-sample shading is enabled or not, based on `resUsage::resourceWrite, FragCoord/PointCoord`.

3. if `gl_sampleId`,`gl_samplePostion` or  a variable decorated with `sample` is used, sample shading should be enabled.

amdvlk-admin:
### Test summary for commit bb9823102656c91260c3c50f489935f0c1c39823
<details><summary>CTS tests (Failed: 0/213061)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40074/71023 (56.4%)</li>
<li>Failed: 0/71023 (0.0%)</li>
<li>Not Supported: 30949/71023 (43.6%)</li><li>Warnings: 0/71023 (0.0%)</li>
</ul></details>

<li>Passed: 40112/71013 (56.5%)</li>
<li>Failed: 0/71013 (0.0%)</li>
<li>Not Supported: 30901/71013 (43.5%)</li><li>Warnings: 0/71013 (0.0%)</li>
</ul></details>

<li>Passed: 36218/71025 (51.0%)</li>
<li>Failed: 0/71025 (0.0%)</li>
<li>Not Supported: 34807/71025 (49.0%)</li><li>Warnings: 0/71025 (0.0%)</li>
</ul></details>


</ul></details>

AMD-dwang:
ping @nhaehnle for review.

lgc/patch/PatchResourceCollect.cpp
```diff
@@ -1134,9 +1134,9 @@ void PatchResourceCollect::processShader() {
   }
 
   if (m_shaderStage == ShaderStageFragment) {
-    if (m_resUsage->builtInUsage.fs.fragCoord || m_resUsage->builtInUsage.fs.pointCoord ||
-        m_resUsage->builtInUsage.fs.sampleMaskIn) {
-      if (m_pipelineState->getRasterizerState().perSampleShading)
+    if (m_pipelineState->getRasterizerState().perSampleShading) {
+      if (m_resUsage->builtInUsage.fs.fragCoord || m_resUsage->builtInUsage.fs.pointCoord ||
+          m_resUsage->builtInUsage.fs.sampleMaskIn || m_resUsage->resourceWrite)
```
nhaehnle, line 1139: I don't follow this hunk. Why is `m_resUsage->resourceWrite` added into the condition?

nhaehnle (commented):
The metadata change makes sense since the spec now clarifies that it's the "static use" that matters.

But see the inline comment.

lgc/patch/PatchResourceCollect.cpp
AMD-dwang: In theory, as long as `perSampleShading` is true, we should enable sample shading. However, for the performance, we try not to enable it as much as possible. If `m_resUsage->resourceWrite` is true,  whether to enable sample shading has an impact on the output. Please see above shader. Thanks.

nhaehnle (approved):
Ah, I think I see. I still find it a bit confusing but should be okay.

PR #2338, opened 2023-03-08
Extend SpirvLowerTerminator to handle conditional branches

perlfu:
Remove limitation of only handling unconditional branches when culling instructions after an OpKill.

### DIFF
--- a/llpc/lower/llpcSpirvLowerTerminator.cpp
+++ b/llpc/lower/llpcSpirvLowerTerminator.cpp
@@ -117,25 +117,30 @@ void SpirvLowerTerminator::visitCallInst(CallInst &callInst) {
   if (isa<ReturnInst>(*instIter))
     return;
 
-  // If terminator branches to block containing a PHI node which references
-  // this block then instructions following kill may be part of a return
-  // value sequence. In which case we cannot safely modify the instructions.
-  auto blockTerm = parentBlock->getTerminator();
-  if (!blockTerm)
-    return;
-  if (BranchInst *branch = dyn_cast<BranchInst>(blockTerm)) {
-    assert(branch->isUnconditional());
-    BasicBlock *targetBlock = branch->getSuccessor(0);
-    for (PHINode &phiNode : targetBlock->phis()) {
-      for (BasicBlock *incomingBlock : phiNode.blocks()) {
-        if (incomingBlock == parentBlock)
-          return;
+  // We must update any phis that point to this block as stale values confuse later passes
+  for (auto succBlock : successors(parentBlock)) {
+    for (auto &phiNode : succBlock->phis()) {
+      phiNode.removeIncomingValue(parentBlock, false);
+      if (phiNode.getNumOperands() == 0) {
+        // PHI began empty which means this successor block is now unreachable.
+        // Update all references to the PHI with poison and mark it for removal.
+        // We can do this because all references to this PHI are now also unreachable,
+        // or are PHI nodes which will be cleaned when the CFG is simplified.
+        auto poisonValue = PoisonValue::get(phiNode.getType());
+        phiNode.replaceAllUsesWith(poisonValue);
+        m_removalStack.emplace_back(&phiNode);
       }
     }
   }
 
-  // Add return
-  ReturnInst::Create(*m_context, nullptr, &*instIter);
+  // Add return (of poison)
+  auto returnType = parentBlock->getParent()->getReturnType();
+  if (returnType && !returnType->isVoidTy()) {
+    auto returnValue = PoisonValue::get(returnType);
+    ReturnInst::Create(*m_context, returnValue, &*instIter);
+  } else {
+    ReturnInst::Create(*m_context, nullptr, &*instIter);
+  }
 
   // Mark all other instructions for removal
   while (instIter != parentBlock->end()) {
--- a/llpc/test/shaderdb/core/OpKill_TestFunctionInlineReturn_lit.frag
+++ b/llpc/test/shaderdb/core/OpKill_TestFunctionInlineReturn_lit.frag
@@ -4,18 +4,28 @@
 /*
 ; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
 
-; Test that after SPIR-V lowering all operations after second kill
-; have been replaced with a branch to the return block; however,
-; the operations after first kill should not be touched as these
-; are part of return .
+; Test that after SPIR-V lowering there are only two kills and
+; both branch directly to the return block.
+
+; Check that the selection of constants is correctly preserved
+; in the inlined function call. Selection relies on SimplifyCFG
+; correctly lowering a PHI, in a manner that can fail if
+; PHIs are not updated correctly.
 
 ; SHADERTEST-LABEL: {{^// LLPC.*}} SPIR-V lowering results
 ; SHADERTEST: call void{{.*}} @lgc.create.kill
-; SHADERTEST-NEXT: br label %"{{.*}}.exit"
-; SHADERTEST: call void{{.*}} @lgc.create.kill
 ; SHADERTEST-NEXT: br label %[[LABEL:[0-9]*]]
+
+; SHADERTEST: "myfunc1(i1;.exit":
+; SHADERTEST: select i1 %{{.*}}, float 3.000000e+00, float 5.000000e+00
+
+; SHADERTEST: call void{{.*}} @lgc.create.kill
+; SHADERTEST-NEXT: br label %[[LABEL]]
+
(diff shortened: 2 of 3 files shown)


llpc/lower/llpcSpirvLowerTerminator.cpp
```diff
@@ -124,12 +124,12 @@ void SpirvLowerTerminator::visitCallInst(CallInst &callInst) {
...
-    for (PHINode &phiNode : targetBlock->phis()) {
-      for (BasicBlock *incomingBlock : phiNode.blocks()) {
-        if (incomingBlock == parentBlock)
-          return;
+    for (BasicBlock *targetBlock : branch->successors()) {
+      for (PHINode &phiNode : targetBlock->phis()) {
+        for (BasicBlock *incomingBlock : phiNode.blocks()) {
+          if (incomingBlock == parentBlock)
+            return;
+        }
```
nhaehnle, line 132: I realize a lot of this was already there, but there are a number of really strange bits about this code:

1. The check for `blockTerm` -- there has to be a blockTerm in illegal IR, so the check is misleading and can/should be removed. (You could replace it by an assertion if you really want to but personally I don't think that level of paranoia is necessary :))
2. The block terminator could be a `switch` instruction (plus, safest would be to guard against unhandled terminators via `llvm_unreachable`).
3. We unconditionally insert a return instruction with a "null" return value. That's dodgy as hell. How about returning `poison` if the function has a non-void return type?
4. If a successor block contains a phi node, it *must* have `parentBlock` as one of the incoming blocks. Again, if it doesn't, that's invalid IR. So the two inner loops are unnecessary: it's sufficient to check whether the `targetBlock` (better name: `successor`) has a phi node or not.

Also, now that I think about it, even that test isn't quite sufficient because `parentBlock` could be a dominator of its successors, and then no phi nodes are necessary in the first place. Hmm, this is surprisingly tricky to get fully right.

Split the block, insert a conditional branch with constant condition to a block with a `ret` instruction, and let SimplifyCFG take care of it? :)

perlfu: All good points.
I'll have a little bit of a rethink on the whole implementation.

Leaving it all to SimplifyCFG alone cannot necessarily work if the ret containing block relies on values we will skip, etc.
That's said it's OK to just bail on things we cannot handle.
The original purpose of this was to deal with odd stuff like unreachable being added by some compilers to the SPIRV.

perlfu: Reanalyzing the original test case for the branch testing code, it turns out that really what this code was preventing was leaving references to the block containing the kill in PHI nodes.
It seems to be we can always modify this to a return if we updated the PHIs, so I have done this instead.

This now also generates returns of the correct (poison) type.

I tested the SimplifyCFG version as well, and this works for the simple test cases; however, the reason this was written to explicitly erase instruction is to remove any of their references.  If we simply split blocks and leave the work to SimplifyCFG then IR can be invalid leaving this pass as the orphan blocks will references values that they cannot access... Although I guess that is always true of orphan blocks?

nhaehnle: This is a very nice cleanup. It looks like the "code from the book" for this problem. (I should really start to misappropriate that expression :D)

On the last point: A dominates B if all paths from the entry to B go through A.

If B is reachable, then there are no paths from the entry to B. Therefore, all paths from the entry to B go through A.

Therefore, an unreachable block is dominated by all blocks, and so an unreachable block may refer to values from any other basic block and still be valid SSA form.

nhaehnle (dismissed):

amdvlk-admin:
b118d18de459f301fcdfb8e60bfdfe005f285f1b Jenkins build error.
/home/amdvlk/jenkins/workspace/vulkan/sanitized-opensource/Github-PR/llpc-github-pr/driver_build/drivers/llpc/lgc/patch/Patch.cpp:75:10: fatal error: llvm/Transforms/IPO/PassManagerBuilder.h: No such file or directory
 #include  llvm/Transforms/IPO/PassManagerBuilder.h 
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[2870/3046] Building CXX object compiler/llpc/llvm/tools/lgc/CMakeFiles/LLVMlgc.dir/builder/SubgroupBuilder.cpp.o
[2871/3046] Building CXX object compiler/llpc/llvm/tools/lgc/CMakeFiles/LLVMlgc.dir/elfLinker/FetchShader.cpp.o

perlfu:
Rebased to remove Jenkins build error.

nhaehnle (dismissed):

amdvlk-admin:
### Test summary for commit 1d78761ecd4cf81c3edc65e826956c221656baec
<details><summary>CTS tests (Failed: 3/208297)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36985/69439 (53.3%)</li>
<li><details><summary>Failed: 1/69439 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.graphicsfuzz.unreachable-switch-case-with-discards
Stack trace: Crash
```
</p></details></li>

</ul></details>

<li>Passed: 37205/69417 (53.6%)</li>
<li><details><summary>Failed: 1/69417 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.graphicsfuzz.unreachable-switch-case-with-discards
Stack trace: Crash
```
</p></details></li>

</ul></details>

<li>Passed: 34224/69441 (49.3%)</li>
<li><details><summary>Failed: 1/69441 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.graphicsfuzz.unreachable-switch-case-with-discards
Stack trace: Crash
```
</p></details></li>

</ul></details>


</ul></details>

perlfu:
- Rebase
- Fix handling of PHIs when they become empty (need removing), i.e. because a kill successor block became unreachable

amdvlk-admin:
### Test summary for commit d53a9ce7e69275e39ac58ee40079b7014f954ce9

</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerTerminator.cpp
```diff
@@ -117,25 +117,26 @@ void SpirvLowerTerminator::visitCallInst(CallInst &callInst) {
...
       }
     }
   }
 
-  // Add return
-  ReturnInst::Create(*m_context, nullptr, &*instIter);
+  // Add return (of poison)
+  auto returnType = parentBlock->getParent()->getReturnType();
+  auto returnValue = PoisonValue::get(returnType);
+  ReturnInst::Create(*m_context, returnValue, &*instIter);
```
nhaehnle, line 139: Should this be conditional depending on whether the return type is void or not?

nhaehnle (commented):
Mostly LGTM, but I do have one issue inline.

llpc/lower/llpcSpirvLowerTerminator.cpp
perlfu: That's true.  Although this was working apparently.
One oddity I have noticed in LLVM recently is that `void poison` is a valid type to create in the IR data structures, but is illegal to parse.  I suspect this might be a case where this works because `void` is actually a type in the type system, so `getReturnType()` is not `nullptr`.
Anyway, I will fix it.

perlfu:
- Rebase
- Avoid creating `void poison` return values

llpc/test/shaderdb/core/OpKill_TestFunctionInlineReturn_lit.frag
```diff
@@ -4,18 +4,28 @@
...
-; Test that after SPIR-V lowering all operations after second kill
-; have been replaced with a branch to the return block; however,
-; the operations after first kill should not be touched as these
-; are part of return .
+; Test that after SPIR-V lowering there are only two kills and
+; both branch directly to the return block.
+
+; Check that the selection of constants is correctly preserved
+; in the inlined function call. Selection relies on SimplifyCFG
+; correctly lowering a PHI, in a manner than can fail if
```
tsymalla, line 12: nit: than => that ?

perlfu: Fixed, thanks!

nhaehnle (approved):
Thanks!

perlfu:
retest this please

amdvlk-admin:
### Test summary for commit fe72c8972aa4356321b75bb701a2ae60ebb038e5

</ul></details>

</ul></details>

</ul></details>


</ul></details>

PR #2345, opened 2023-03-10
Add a test case for a bug with arrays of variable pointers

nhaehnle:
This came up in an upstream discussion about fat buffer pointers. The underlying issue is that there is legal shader SPIR-V that results in load/store of buffer pointers to survive all the way through to PatchBufferOp.

The impact of the bug is limited because neither GLSL nor HLSL can express this construct.

See also: https://discourse.llvm.org/t/representing-buffer-descriptors-in-the-amdgpu-target-call-for-suggestions/68798/35


nhaehnle:
cc @piotrAMD

piotrAMD (approved):
Nice, thanks!

The shader is valid to my eyes too.

amdrexu (approved):

PR #2112, opened 2022-11-01
Use a better insertion point for buffer descriptor loads

jayfoad:
When inserting buffer descriptor loads and push constant loads, insert them at the nearest common dominator of all uses, instead of always in the entry block.

Currently we run an IR code sinking pass later which would have sunk these loads anyway, but we would like to stop running that pass because it can have bad effects on register pressure.

PR #2358, opened 2023-03-16
Handle unreachable descriptor load uses

dstutt:
#2112 inserts buffer loads in a better location than previously. In some cases, the use it is considering is in an unreachable block (not in the domTree) which causes an assert/seg-fault.

Add additional checks that handle unreachable uses differently and replace the use with a poison value instead.

### DIFF
--- a/llpc/lower/llpcSpirvLowerGlobal.cpp
+++ b/llpc/lower/llpcSpirvLowerGlobal.cpp
@@ -1794,12 +1794,7 @@ static Instruction *findNearestCommonDominator(Instruction *instA, Use &useB, Do
   if (!instA)
     return instB;
 
-  BasicBlock *blockA = instA->getParent();
-  if (blockA == blockB)
-    return instA->comesBefore(instB) ? instA : instB;
-
-  BasicBlock *block = domTree.findNearestCommonDominator(blockA, blockB);
-  return block == blockA ? instA : (block == blockB ? instB : block->getTerminator());
+  return domTree.findNearestCommonDominator(instA, instB);
 }
 
 // =====================================================================================================================
@@ -2529,6 +2524,9 @@ void SpirvLowerGlobal::lowerBufferBlock() {
           insertPoint = findNearestCommonDominator(insertPoint, use, domTree);
         }
 
+        if (!insertPoint)
+          continue;
+
         m_builder->SetInsertPoint(insertPoint);
         unsigned bufferFlags = global.isConstant() ? 0 : lgc::Builder::BufferFlagWritten;
         Value *const bufferDesc =
@@ -2643,6 +2641,9 @@ void SpirvLowerGlobal::lowerPushConsts() {
         usesToReplace.push_back(inst);
         insertPoint = findNearestCommonDominator(insertPoint, use, domTree);
       }
+
+      if (!insertPoint)
+        continue;
 
       m_builder->SetInsertPoint(insertPoint);
 


llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -2510,6 +2510,7 @@ void SpirvLowerGlobal::lowerBufferBlock() {
         DominatorTree &domTree = m_functionAnalysisManager->getResult<DominatorTreeAnalysis>(*func);
 
         SmallVector<Instruction *, 8> usesToReplace;
+        SmallVector<Instruction *, 8> unreachableUses;
```
trenouf, line 2513: Actually, despite what I said earlier, you might be able to do this without the unreachableUses vector as it is already saving the ones it wants to process into usesToReplace. Not a blocker though.

trenouf (approved):
LGTM

llpc/lower/llpcSpirvLowerGlobal.cpp
dstutt: I think you do need to do it this way - it's still replacing a use ultimately (I tried and it didn't work).

dstutt: However, tests show up a case that isn't handled (all uses replaced with poison and insertPoint is still nullptr). So a slight re-arrangement and a null test are required I think.

trenouf: Is it the case that, having processed all the reachable uses in instructions, all that is left are unreachable uses in instructions? I.e. have you also got rid of non-instruction uses by then?

If so, then you can just pick one off the front each time in a loop:
```
while (!global.use_empty())
  *global.use_begin() = PoisonValue::get(global.use_begin()->getType());
```
(pointers and refs are approximate only as I can never remember how to use a Use properly).

trenouf (approved):
LGTM

llpc/lower/llpcSpirvLowerGlobal.cpp
dstutt: Hmm, could be. I think the implementation I've ended up with is clear - unless you think it needs refactoring to be more like what you suggest, I'll leave as-is?

amdvlk-admin:
### Test summary for commit 9512e64519e7148e58c981d722b6323a0faf81be
<details><summary>CTS tests (Failed: 0/207358)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36986/69126 (53.5%)</li>
<li>Failed: 0/69126 (0.0%)</li>
<li>Not Supported: 32140/69126 (46.5%)</li><li>Warnings: 0/69126 (0.0%)</li>
</ul></details>

<li>Passed: 37206/69104 (53.8%)</li>
<li>Failed: 0/69104 (0.0%)</li>
<li>Not Supported: 31898/69104 (46.2%)</li><li>Warnings: 0/69104 (0.0%)</li>
</ul></details>

<li>Passed: 34225/69128 (49.5%)</li>
<li>Failed: 0/69128 (0.0%)</li>
<li>Not Supported: 34903/69128 (50.5%)</li><li>Warnings: 0/69128 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit ebb1cb07e288d8f8830d806d3b0796157a658eae
<details><summary>CTS tests (Failed: 0/207358)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36986/69126 (53.5%)</li>
<li>Failed: 0/69126 (0.0%)</li>
<li>Not Supported: 32140/69126 (46.5%)</li><li>Warnings: 0/69126 (0.0%)</li>
</ul></details>

<li>Passed: 37206/69104 (53.8%)</li>
<li>Failed: 0/69104 (0.0%)</li>
<li>Not Supported: 31898/69104 (46.2%)</li><li>Warnings: 0/69104 (0.0%)</li>
</ul></details>

<li>Passed: 34225/69128 (49.5%)</li>
<li>Failed: 0/69128 (0.0%)</li>
<li>Not Supported: 34903/69128 (50.5%)</li><li>Warnings: 0/69128 (0.0%)</li>
</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -1780,7 +1780,9 @@ void SpirvLowerGlobal::storeOutputMember(Type *outputTy, Type *storeTy, Value *s
 // =====================================================================================================================
 // Find a new insertion point that dominates the current insertion point \p instA and a specified instruction use
 // \p useB.
-static Instruction *findNearestCommonDominator(Instruction *instA, Use &useB, DominatorTree &domTree) {
+static Instruction *findNearestCommonDominator(Instruction *instA, Use &useB, DominatorTree &domTree,
+                                               SmallVector<Instruction *, 8> &usesToReplace,
```
trenouf, line 1784: Hmm, wouldn't it be better to not pass these vectors here, and instead just make findNearestCommonDominator tell the caller that useB is not in the dom tree by returning nullptr or something?

Does that contradict what Jay said?

Also, if you are going to pass vectors here, use SmallVectorImpl so the function def/decl doesn't need to know what default capacity you used in the vector.

If you do go with this approach, the comment above the function is inadequate.

amdvlk-admin:
### Test summary for commit 45770b06bf7fe36c748d67b79179e4b3fb96d951
<details><summary>CTS tests (Failed: 3/207358)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36985/69126 (53.5%)</li>
<li><details><summary>Failed: 1/69126 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.instruction.compute.variable_pointers.complex_types_compute.opphi_outer_struct_single_buffer_first_input
Stack trace: Crash
```
</p></details></li>

<li>Not Supported: 32140/69126 (46.5%)</li><li>Warnings: 0/69126 (0.0%)</li>
</ul></details>

<li>Passed: 37205/69104 (53.8%)</li>
<li><details><summary>Failed: 1/69104 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.instruction.compute.variable_pointers.complex_types_compute.opphi_outer_struct_single_buffer_first_input
Stack trace: Crash
```
</p></details></li>

<li>Not Supported: 31898/69104 (46.2%)</li><li>Warnings: 0/69104 (0.0%)</li>
</ul></details>

<li>Passed: 34224/69128 (49.5%)</li>
<li><details><summary>Failed: 1/69128 (0.0%)</summary><p>


Failures:
```
FAILURE: dEQP-VK.spirv_assembly.instruction.compute.variable_pointers.complex_types_compute.opphi_outer_struct_single_buffer_first_input
Stack trace: Crash
```
</p></details></li>

<li>Not Supported: 34903/69128 (50.5%)</li><li>Warnings: 0/69128 (0.0%)</li>
</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerGlobal.cpp
```diff
@@ -1790,6 +1792,13 @@ static Instruction *findNearestCommonDominator(Instruction *instA, Use &useB, Do
     instB = blockB->getTerminator();
   }
 
+  // Check if useB is actually reachable
+  if (!domTree.getNode(blockB)) {
+    unreachableUses.push_back(instB);
+    return instA;
+  }
+
+  usesToReplace.push_back(instB);
```
ruiling, line 1801: Instead of doing this, I think we can switch to `Instruction *` version of `findNearestCommonDominator()`.
That is replace:
```
  BasicBlock *blockA = instA->getParent();
  if (blockA == blockB)
    return instA->comesBefore(instB) ? instA : instB;
  BasicBlock *block = domTree.findNearestCommonDominator(blockA, blockB);
  return block == blockA ? instA : (block == blockB ? instB : block->getTerminator());
```
with:
```
return domTree.findNearestCommonDominator(instA, instB);
```

ruiling (commented):
Based on what @nhaehnle has said at https://github.com/GPUOpen-Drivers/llpc/pull/2338#discussion_r1133880176, I think we don't need to replace the unreachable uses. We only need to work-around that `BasicBlock *` version of `findNearestCommonDominator()` requires the blocks should be reachable. In fact, the `Instruction *` version of `findNearestCommonDominator()` already supports this well.

llpc/lower/llpcSpirvLowerGlobal.cpp
dstutt: Thanks - that's a good point. See the latest proposal.

llpc/lower/llpcSpirvLowerGlobal.cpp
dstutt: Going with Ruiling's suggestion.

trenouf (approved):
Yeah, Ruiling's idea is much better. :-)

So does that leave the unreachable use pointing at the original global? I guess they will both disappear at some point. Or does it change the unreachable use to the new closest dominator value? If it is still like that by the end of the pass, that is illegal IR, and -verify-ir would complain.

jayfoad:
> So does that leave the unreachable use pointing at the original global? I guess they will both disappear at some point.

Yes if all uses are unreachable, and hopefully the global will get removed soon and won't cause any problems in the interim, but I don't know for sure.

> Or does it change the unreachable use to the new closest dominator value? If it is still like that by the end of the pass, that is illegal IR, and -verify-ir would complain.

Yes if only some uses are unreachable, but I think IR verification should be OK because unreachable blocks are deemed to be dominated by everything.

jayfoad (approved):
Nice!

dstutt:
Bother - I forgot to update the commit message.

PR #2379, opened 2023-04-04
Add a new option to control whether fmask is enabled

AMD-dwang:
Currently, the option "shadowDescriptorTable" not only controls whether fmask is enabled, but also is used as high dword of shadow descriptor table address. This doesn't meet the cases that the descriptors are not in shadow table, shadowDescriptorTable will be set ~0u, and fmask is required to be enabled.


lgc/builder/DescBuilder.cpp
```diff
@@ -425,9 +425,8 @@ Value *BuilderImpl::getDescPtr(ResourceNodeType concreteType, ResourceNodeType a
     // - value for high 32 bits of the pointer; HighAddrPc to use PC
     if (node || topNode || concreteType != ResourceNodeType::DescriptorFmask) {
       unsigned shadowDescriptorTable = m_pipelineState->getOptions().shadowDescriptorTable;
-      bool shadow =
-          concreteType == ResourceNodeType::DescriptorFmask && shadowDescriptorTable != ShadowDescriptorTableDisable;
-      Value *highHalf = getInt32(shadow ? shadowDescriptorTable : HighAddrPc);
+      bool enableFmask = concreteType == ResourceNodeType::DescriptorFmask && m_pipelineState->getOptions().enableFmask;
+      Value *highHalf = getInt32(enableFmask ? shadowDescriptorTable : HighAddrPc);
```
nhaehnle, line 429: This doesn't make sense. We should only get here if fmask is enabled, and the check is about where to load the fmask from, so it should remain unchanged.

lgc/elfLinker/RelocHandler.cpp
```diff
@@ -216,7 +216,7 @@ bool RelocHandler::getValue(StringRef name, uint64_t &value) {
     return true;
   }
   if (name == reloc::ShadowDescriptorTableEnabled) {
-    value = m_pipelineState->getOptions().shadowDescriptorTable != ShadowDescriptorTableDisable;
+    value = m_pipelineState->getOptions().enableFmask;
     return true;
   }
```
nhaehnle, line 221: Same here, this about where to find the fmask, not whether to use an fmask. Therefore, it shouldn't be changed.

lgc/interface/lgc/Pipeline.h
```diff
@@ -133,8 +133,8 @@ struct Options {
   bool fullSubgroups;                  // Use full subgroup lanes
   unsigned nggVertsPerSubgroup;        // How to determine NGG verts per subgroup
   unsigned nggPrimsPerSubgroup;        // How to determine NGG prims per subgroup
-  unsigned shadowDescriptorTable;      // High dword of shadow descriptor table address, or
-                                       //   ShadowDescriptorTableDisable to disable shadow descriptor tables
+  unsigned shadowDescriptorTable;      // High dword of shadow descriptor table address
+  bool enableFmask;                    // Enable Fmask
```
nhaehnle, line 137: A more descriptive comment may be helpful:
```suggestion
  bool enableFmask;                    // Whether to use Fmasks when loading from MSAA images
```

lgc/patch/SystemValues.cpp
```diff
@@ -560,5 +560,5 @@ unsigned ShaderSystemValues::findResourceNodeByDescSet(unsigned descSet) {
 // =====================================================================================================================
 // Test if shadow descriptor table is enabled
 bool ShaderSystemValues::isShadowDescTableEnabled() const {
-  return m_pipelineState->getOptions().shadowDescriptorTable != ShadowDescriptorTableDisable;
+  return m_pipelineState->getOptions().enableFmask;
```
nhaehnle, line 563: Again, should not be changed.

lgc/state/PipelineState.cpp
```diff
@@ -1003,8 +1003,7 @@ PipelineState::findResourceNode(ResourceNodeType nodeType, unsigned descSet, uns
       return {&node, &node};
   }
 
-  if (nodeType == ResourceNodeType::DescriptorFmask &&
-      getOptions().shadowDescriptorTable != ShadowDescriptorTableDisable) {
+  if (nodeType == ResourceNodeType::DescriptorFmask && getOptions().enableFmask) {
```
nhaehnle, line 1006: Pay attention to the comment inside of the `if` statement, please. This should not be changed, or the comment also needs changing.

nhaehnle (commented):
So is the plan to add a corresponding flag to the vkgc interface with the driver?

In general, I think you need to be a bit more careful about which places in the code talk about **whether** fmasks are loaded versus **where** they are loaded from.

AMD-dwang:
> So is the plan to add a corresponding flag to the vkgc interface with the driver?
> 
> In general, I think you need to be a bit more careful about which places in the code talk about **whether** fmasks are loaded versus **where** they are loaded from.

There have been already two options (`shadowDescriptorTableUsage and shadowDescriptorTablePtrHigh`) to control Fmask in vkgcDefs.h, only one of them was removed in compiler.

In vulkan driver, the `fmask` is from shadow table, but maybe not in other drivers. My thought is that if `fmask` is not from shadow table, `shadowDescriptorTablePtrHigh` will be set `~0u (That's the value of ShadowDescriptorTableDisable)`, so we will not care about the high address of `fmask`, if `fmask` is enabled, `shadowDescriptorTablePtrHigh` is used, otherwise `HighAddrPc` is used. Perhaps there should be a more meaningful name for two options, such as `enableFmask` and `highAddrOfFmask`.

amdvlk-admin:
### Test summary for commit dfc0ec6c2afe3f39f2086e5dd53f5cf5858a8b16
<details><summary>CTS tests (Failed: 0/208307)</summary><ul><li>Built with version 1.3.0.0</li>

<li>Passed: 36988/69449 (53.3%)</li>
<li>Failed: 0/69449 (0.0%)</li>
<li>Not Supported: 32461/69449 (46.7%)</li><li>Warnings: 0/69449 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle:
Okay, makes sense. The variable names are fine. I do think my earlier inline comments apply, though, at least in general. Consider what happens in `getDescPtr`, for example. It really shouldn't be checking `enableFmask`.

nhaehnle:
By the way: can Vulkan still use a shadow descriptor table for fmask when VK_EXT_descriptor_buffer is used?

AMD-dwang:
> Okay, makes sense. The variable names are fine. I do think my earlier inline comments apply, though, at least in general. Consider what happens in `getDescPtr`, for example. It really shouldn't be checking `enableFmask`.

I see.  I will address the comments according to semantics. Thanks.

> By the way: can Vulkan still use a shadow descriptor table for fmask when VK_EXT_descriptor_buffer is used?
No, we disable fmask.

amdvlk-admin:
### Test summary for commit 8c394068e01917cd3ae69d8b3a45d7a225ddb9ae

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/builder/DescBuilder.cpp
```diff
@@ -424,10 +424,9 @@ Value *BuilderImpl::getDescPtr(ResourceNodeType concreteType, ResourceNodeType a
     // - descriptor binding number
     // - value for high 32 bits of the pointer; HighAddrPc to use PC
     if (node || topNode || concreteType != ResourceNodeType::DescriptorFmask) {
-      unsigned shadowDescriptorTable = m_pipelineState->getOptions().shadowDescriptorTable;
-      bool shadow =
-          concreteType == ResourceNodeType::DescriptorFmask && shadowDescriptorTable != ShadowDescriptorTableDisable;
-      Value *highHalf = getInt32(shadow ? shadowDescriptorTable : HighAddrPc);
+      unsigned highAddrOfFmask = m_pipelineState->getOptions().highAddrOfFmask;
+      bool isFmask = concreteType == ResourceNodeType::DescriptorFmask;
+      Value *highHalf = getInt32(isFmask ? highAddrOfFmask : HighAddrPc);
```
nhaehnle, line 429: Please add an
```
assert(!(isFmask && highAddrOfFmask == ShadowDescriptorTableDisable) && "not implemented");
```
After all, in the longer term this change only makes sense if it is possible to use fmasks without a shadow descriptor table.

nhaehnle (commented):
Interesting, that seems like an avoidable pessimization. I suppose this change is one step towards avoiding the pessimization?

AMD-dwang:
> Interesting, that seems like an avoidable pessimization. I suppose this change is one step towards avoiding the pessimization?

Yeah, this change is preparation.

PR #2393, opened 2023-04-11
lgc: Accept num patch control points via shader mode

trenouf:


amdvlk-admin:
### Test summary for commit 5c911a3f57764cc17a872ea5b4ea3a2ba6e5e5c9

</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (commented):
Why support both the ShaderMode and the InputAssemblyState? Wouldn't it be better to only support one?

trenouf:
I changed it so LGC only supports ShaderMode, but it's a bit shonky with the hoops that the front-end has to jump through, especially with the need to include patchControlPoints in the part pipeline state hash. I think the original scheme is better where the front-end can use either ShaderMode or InputAssemblyState for patchControlPoints.

trenouf:
Ping @nhaehnle -- can I go back to my original scheme (modulo the two fixes I did for it)? Or do you still prefer the only-ShaderMode scheme?

llpc/context/llpcPipelineContext.h
```diff
@@ -146,6 +146,9 @@ class PipelineContext {
   // Set pipeline state in lgc::Pipeline object for middle-end, and (optionally) hash the state.
   virtual void setPipelineState(lgc::Pipeline *pipeline, Util::MetroHash64 *hasher, bool unlinked) const;
 
+  // For TCS, set inputVertices from patchControlPoints in the pipeline state.
+  virtual void setTcsInputVertices(llvm::Module *tcsModule) {}
+
```
nhaehnle, line 151: Maybe put an `llvm_unreachable` in there just in case?

nhaehnle (commented):
I do think it's best to have the option only in one place.

Yes, the `hasher` thing is a bit unfortunate but it's always been a bit strange in the first place to have a "setPipelineState" method optionally compute a hash. That's also only used in one place.

trenouf:
OK, I'll go with your way, as I need to get the change in. I still think it is the wrong way to do it as it spreads shonkiness around the front-end.

setPipelineState computing the hash is deliberate because, for the part pipeline scheme, we need to make sure the hash includes LGC pipeline state needed for that part pipeline, without risking someone adding some new state and forgetting to include it in the hash. It was less strange before this change.

nhaehnle (approved):
Thanks.

And yeah, the main part that makes me uncomfortable about the hashing is that it's *only* used for part pipelines. That feels a bit arbitrary. But it's also not the point here.

PR #2411, opened 2023-04-18
Fix tests to cope with upstream change to smem instructions used

dstutt:
Upstream change:
[AMDGPU] Don't select _SGPR forms of SMEM instructions on GFX9+

Update tests to allow for an optional "offset:0x0" which appears after the change.


jayfoad (approved):

trenouf (approved):
LGTM

PR #2432, opened 2023-04-28
Update tests to expect poisoned shufflevector masks instead of undef ones.

kosarev:
Fixes passing tests after:
[IR] Change shufflevector undef mask to poison
https://reviews.llvm.org/D149210


dstutt (approved):
LGTM

kosarev:
Do the checks need to be restarted or it's something else that is not quite right?

dstutt:
I think the Jenkins Pipeline failure is an internal infrastructure failure (I looked on the Jenkins server).

Flakebi:
retest this please

dstutt:
It will just fail again on Jenkins Pipeline - some internal infrastructure is down.
The change here is tested by the LLPC Docker CI build/test so we could just waive it through.

PR #2433, opened 2023-04-28
lgc: Add ModuleBunch infrastructure

trenouf:
This adds a pseudo-IR unit "ModuleBunch" (a bunch of modules), and the pass manager infrastructure to allow module bunch passes and analysis.

In the same way as a module pass manager can have an adaptor to run (e.g.) function passes, a module bunch pass manager can have an adaptor to run module passes.

Currently that adaptor runs multiple modules sequentially, but the design allows for a future change where modules can be run in parallel, as long as they are created in separate LLVMContexts. If that happens, a ModuleBunch pass that wants to move code from one module to another would need to be aware of that possibility.

To run in parallel, there needs to be a separate copy of the ModulePassManager inside the adaptor per thread. To enable that, there is a constructor that takes a PassMaker function instead of a pass, and the adaptor calls that once per distinct LLVMContext. There is no actual parallelism yet.

The normal PassBuilder does not know about module bunch passes, so you must instead use MbPassBuilder (added here).

Unlike for other real IR units, there is no way to directly get the parent ModuleBunch from a Module. A user of this feature could provide its own local way to do that for its passes, perhaps via a thread-local variable.

LLVM's standard instrumentations unfortunately blow up in some places when they see a novel IR unit they haven't heard of. This commit adds a new MbStandardInstrumentations that contains copies of a few of the standard instrumentations fixed up to cope with ModuleBunches. In particular, -print-after etc and -verify-ir work.

### DIFF
--- a/lgc/interface/lgc/ModuleBunch.h
+++ b/lgc/interface/lgc/ModuleBunch.h
@@ -1,0 +1,301 @@
+/*
+ ***********************************************************************************************************************
+ *
+ *  Copyright (c) 2023 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in all
+ *  copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


lgc/interface/lgc/ModuleBunch.h
```diff
@@ -0,0 +1,301 @@
...
+
+class ModuleBunch;
+
+using ModuleBunchPassManager = PassManager<ModuleBunch>;
+using ModuleBunchAnalysisManager = AnalysisManager<ModuleBunch>;
+using ModuleAnalysisManagerModuleBunchProxy = InnerAnalysisManagerProxy<ModuleAnalysisManager, ModuleBunch>;
+using ModuleBunchAnalysisManagerModuleProxy = OuterAnalysisManagerProxy<ModuleBunchAnalysisManager, Module>;
+
+/// A pointer encapsulated as random access iterator, so that it can be subclassed to handle dereferences differently.
+template <typename T> class PointerAsIterator : public std::iterator<std::random_access_iterator_tag, T> {
```
nhaehnle, line 44: Could/should this use LLVM's `iterator_adaptor_base` or even `pointee_iterator`?

nhaehnle (commented):
Mostly LGTM, though I haven't gone through all the copied code in detail.

lgc/interface/lgc/ModuleBunch.h
trenouf: Good idea, also suggested offline by Jannik.

amdvlk-admin:
### Test summary for commit 742cfa023935db096ef77fa1344336aa06ad2120
<details><summary>CTS tests (Failed: 0/144009)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40695/72001 (56.5%)</li>
<li>Failed: 0/72001 (0.0%)</li>
<li>Not Supported: 31306/72001 (43.5%)</li><li>Warnings: 0/72001 (0.0%)</li>
</ul></details>

<li>Timeout or crash</li>

</ul></details>

<li>Passed: 36765/72008 (51.1%)</li>
<li>Failed: 0/72008 (0.0%)</li>
<li>Not Supported: 35243/72008 (48.9%)</li><li>Warnings: 0/72008 (0.0%)</li>
</ul></details>


</ul></details>

trenouf:
The Jenkins failure looks unrelated to this change -- the RX6900 machine is just crashing without doing anything, and that happens in other surrounding builds.

nhaehnle (approved):
Thanks

PR #2438, opened 2023-05-04
Optimize fast link: Disable resource relocatable

AMD-dwang:
Currently, relocatable shader is unsupported if there is no 1-to-1 mapping

between descriptor sets and descriptor tables, in which case we have to go back

to full pipeline, it's slower.

According to the vulkan spec, when compiling graphics pipeline library, the

resources are identified, so we don't need to use relocatable resources.


amdvlk-admin:
### Test summary for commit 2a615aa157d5af840fb1f13e2d9eb48124da97dd
<details><summary>CTS tests (Failed: 0/215993)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40695/72001 (56.5%)</li>
<li>Failed: 0/72001 (0.0%)</li>
<li>Not Supported: 31306/72001 (43.5%)</li><li>Warnings: 0/72001 (0.0%)</li>
</ul></details>

<li>Passed: 40713/71984 (56.6%)</li>
<li>Failed: 0/71984 (0.0%)</li>
<li>Not Supported: 31271/71984 (43.4%)</li><li>Warnings: 0/71984 (0.0%)</li>
</ul></details>

<li>Passed: 36765/72008 (51.1%)</li>
<li>Failed: 0/72008 (0.0%)</li>
<li>Not Supported: 35243/72008 (48.9%)</li><li>Warnings: 0/72008 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit 2270cc754af3b29681dc17202be535caaceeaf03
<details><summary>CTS tests (Failed: 0/215993)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40695/72001 (56.5%)</li>
<li>Failed: 0/72001 (0.0%)</li>
<li>Not Supported: 31306/72001 (43.5%)</li><li>Warnings: 0/72001 (0.0%)</li>
</ul></details>

<li>Passed: 40713/71984 (56.6%)</li>
<li>Failed: 0/71984 (0.0%)</li>
<li>Not Supported: 31271/71984 (43.4%)</li><li>Warnings: 0/71984 (0.0%)</li>
</ul></details>

<li>Passed: 36765/72008 (51.1%)</li>
<li>Failed: 0/72008 (0.0%)</li>
<li>Not Supported: 35243/72008 (48.9%)</li><li>Warnings: 0/72008 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle:
+1 for avoiding relocations when they're not necessary, but let's make sure the design makes sense.

My understanding is that the default usage of Vulkan should never need relocations. Correct? In other words, the default should be to *not* use relocations. Which in turn means that we shouldn't add an option called `disableRelocResource`. It is generally better to have options that are "enable something". So if anything, it should be an option that must be set to true to enable relocations. But is that option truly required?

Perhaps you could share a pipeline dump / add a .pipe test somewhere in the llpc tests that illustrates the envisioned use case?

We should keep these two related but orthogonal concepts apart:

1. Whether the full graphics pipeline is available or not or whether only an individual shader / part pipeline is being compiled.
2. Whether all required pipeline layout info is available or not.

I'm also a bit confused: it looks like XGL uses `buildGraphicsPipelineWithElf`, which in turn uses `buildUnlinkedShaderInternal`. Why is that? What API use case does that support?

AMD-dwang:
> My understanding is that the default usage of Vulkan should never need relocations. Correct? 

Yes, according to the Vulkan spec, when compiling graphics pipeline library, the resources are identified, so we don't need to use relocatable resources.

> But is that option truly required?

The option is only for compatibility with currently existing lit tests.

> I'm also a bit confused: it looks like XGL uses `buildGraphicsPipelineWithElf`, which in turn uses `buildUnlinkedShaderInternal`. Why is that? What API use case does that support?
  
There is no such use case. `buildGraphicsPipelineWithElf` is just for link,  `buildUnlinkedShaderInternal` in `buildGraphicsPipelineWithElf` is never used.   Current design is for special case, using three libraries and an uncompiled shader to create a full pipeline.


We have a thread discussing this and will forward to you later. Thanks.

AMD-dwang:
V1: address the conflicts.

nhaehnle (commented):
Thanks for the reference.

There is already a `PipelineOptions::enableRelocatableShaderElf`, which should be `false` in all normal Vulkan uses. How about removing the new option and just checking the existing option in `PipelineContext::setPipelineState`. It seems like that should work?

AMD-dwang:
> Thanks for the reference.
> 
> There is already a `PipelineOptions::enableRelocatableShaderElf`, which should be `false` in all normal Vulkan uses. How about removing the new option and just checking the existing option in `PipelineContext::setPipelineState`. It seems like that should work?

The option `enableRelocatableShaderElf` is to enable relocatable shader, the newly added option is to handle resource relocatable.  Only if `enableRelocatableShaderElf` is true, the new option is valid.  If you don't want to add more options, can we consider removing the relevant code and the existing lit tests?

nhaehnle:
Well yes, I was thinking that "relocatable shader" means "relocations are used". And this change is about not using relocations :)

Are relocations still used for anything else here? Skimming the code in `RelocHandler`, I see three kinds of relocations that are perhaps unrelated to the pipeline layout:

* $numsamples
* $samplePatternIdx
* $deviceIdx

Do we still need those in the targeted Vulkan flow?

nhaehnle:
As to your other question: it may actually be a good idea to take out some of the relocation support again. It was a Stadia feature to solve exactly the kinds of problems that are now being worked on inside Khronos with graphics pipeline libraries etc., and it looks like the Khronos APIs are not going to require relocations. We should discuss this.

AMD-dwang:
Currently, relocations are only for VK_EXT_graphics_pipeline_library. In addition to relocations that are related to the resources, others are still needed.

I saw the discussions.  Are we sure to remove the relocation of the resource?

nhaehnle:
After internal discussions, I'd say we should just remove the relocation of resource descriptors.

AMD-dwang:
Address the comments:
When compiling a single shader with amdllpc, there is also no resource mapping, in these cases the relocation of resource descriptors is still needed, just don't link. so I only removed some code related to link.

amdvlk-admin:
### Test summary for commit d843f0eda1997d55035d7dfac75199b3961d0e94
<details><summary>CTS tests (Failed: 0/215993)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40695/72001 (56.5%)</li>
<li>Failed: 0/72001 (0.0%)</li>
<li>Not Supported: 31306/72001 (43.5%)</li><li>Warnings: 0/72001 (0.0%)</li>
</ul></details>

<li>Passed: 40713/71984 (56.6%)</li>
<li>Failed: 0/71984 (0.0%)</li>
<li>Not Supported: 31271/71984 (43.4%)</li><li>Warnings: 0/71984 (0.0%)</li>
</ul></details>

<li>Passed: 36765/72008 (51.1%)</li>
<li>Failed: 0/72008 (0.0%)</li>
<li>Not Supported: 35243/72008 (48.9%)</li><li>Warnings: 0/72008 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
Long-term, we can just make `-auto-layout-desc` the default. Anyway, this change seems reasonable to me.

PR #2452, opened 2023-05-10
RegisterMetadata: support unlinked cases

xuechen417:
Currently, the new elf metadata only works on the wholepipeline case while the unlinked case falls back to the old elf register mode. This change will add the support of new elf metadate for them.

### DIFF
--- a/lgc/patch/RegisterMetadataBuilder.cpp
+++ b/lgc/patch/RegisterMetadataBuilder.cpp
@@ -62,8 +62,10 @@ void RegisterMetadataBuilder::buildPalMetadata() {
     DenseMap<unsigned, unsigned> apiHwShaderMap;
     if (m_hasTask || m_hasMesh) {
       assert(m_pipelineState->getTargetInfo().getGfxIpVersion() >= GfxIpVersion({10, 3}));
-      apiHwShaderMap[ShaderStageMesh] = Util::Abi::HwShaderGs;
-      pipelineType = Util::Abi::PipelineType::Mesh;
+      if (m_hasMesh) {
+        apiHwShaderMap[ShaderStageMesh] = Util::Abi::HwShaderGs;
+        pipelineType = Util::Abi::PipelineType::Mesh;
+      }
       if (m_hasTask) {
         apiHwShaderMap[ShaderStageTask] = Util::Abi::HwShaderCs;
         pipelineType = Util::Abi::PipelineType::TaskMesh;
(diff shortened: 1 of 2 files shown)


amdvlk-admin:
### Test summary for commit c0816374c6f802070354ca0402455bad9701e0b2
<details><summary>CTS tests (Failed: 0/143991)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40695/72004 (56.5%)</li>
<li>Failed: 0/72004 (0.0%)</li>
<li>Not Supported: 31309/72004 (43.5%)</li><li>Warnings: 0/72004 (0.0%)</li>
</ul></details>

<li>Passed: 40713/71987 (56.6%)</li>
<li>Failed: 0/71987 (0.0%)</li>
<li>Not Supported: 31274/71987 (43.4%)</li><li>Warnings: 0/71987 (0.0%)</li>
</ul></details>

<li>Timeout or crash</li>

</ul></details>


</ul></details>

lgc/patch/RegisterMetadataBuilder.cpp
```diff
@@ -907,14 +920,21 @@ void RegisterMetadataBuilder::buildPsRegisters() {
...
-  auto entryPoint = m_pipelineShaders->getEntryPoint(shaderStage);
   if (shaderStage == ShaderStageCompute) {
+    Function *attribFunc = nullptr;
+    for (Function &func : *m_module) {
+      // Only entrypoint and amd_gfx functions may have the function attribute for workgroup id.
+      if (isShaderEntryPoint(&func) || (func.getCallingConv() == CallingConv::AMDGPU_Gfx)) {
+        attribFunc = &func;
+        break;
+      }
+    }
```
nhaehnle, line 931: Can you give a bit of background here? Is this related to raytracing indirect call support, where we may have a module that doesn't have an entry point?

lgc/state/PipelineState.cpp
```diff
@@ -434,11 +436,8 @@ void PipelineState::record(Module *module) {
   if (m_palMetadata)
     m_palMetadata->record(module);
 
-  if (UseRegisterFieldFormat) {
-    const bool isFieldSupported =
-        getTargetInfo().getGfxIpVersion().major >= 11 && (m_pipelineLink == PipelineLink::WholePipeline);
-    UseRegisterFieldFormat.setValue(isFieldSupported);
-  }
+  if (UseRegisterFieldFormat)
+    UseRegisterFieldFormat.setValue(getTargetInfo().getGfxIpVersion().major >= 11);
```
nhaehnle, line 440: This must have slipped through in an earlier review. Why are we changing the value of a cl::opt here? I think this code should just be removed... (also the setting of UseRegisterFieldFormat in the PipelineState constructor.

lgc/patch/RegisterMetadataBuilder.cpp
xuechen417: Actually, the test CsBPermuteWave64.lgc exposes a crash here which has a calling conversion of amdgpu_gfx. So I imitated the process in ShaderInputs::fixupUses to fix the issue.

lgc/state/PipelineState.cpp
xuechen417: If the option is turned on and the gfx ip is pre-gfx11, we should fallback to the old mode. 
The reason that we change cl::opt here is to make relocatable shaders work well. In `PalMetadata::record`, we can only check cl::opt `UseRegisterFieldFormat` instead of `m_pipelineState->useRegisterFieldFormat()` to set pal version metaddata since `m_piplineState` is null for relocatable shaders cases.

lgc/patch/RegisterMetadataBuilder.cpp
nhaehnle: Okay. My main worry was what happens when we have multiple functions in a module. But this is about whether we have workgroup IDs, and with multiple functions we need to be clear about the ABI anyway during PatchEntryPointMutate. So I think this code is okay.

lgc/state/PipelineState.cpp
nhaehnle: This is about the use from `GlueShader::compile`, right?

Could you please make it so that the `PalMetadata` constructor takes an explicit `bool useRegisterFieldFormat`, which can be programmed by GlueShader because all the cases where a (derived class of) GlueShader is instantiated do have a PipelineState object.

The cleanup could be in a follow-up change if you prefer.

nhaehnle (approved):

lgc/state/PipelineState.cpp
xuechen417: Thanks! I will make the cleanup in a follow-up change.

lgc/patch/RegisterMetadataBuilder.cpp
xuechen417: Thanks.

lgc/patch/RegisterMetadataBuilder.cpp
```diff
@@ -907,14 +920,21 @@ void RegisterMetadataBuilder::buildPsRegisters() {
 // Builds register configuration for compute/task shader.
 void RegisterMetadataBuilder::buildCsRegisters(ShaderStage shaderStage) {
   assert(shaderStage == ShaderStageCompute || shaderStage == ShaderStageTask);
-  auto entryPoint = m_pipelineShaders->getEntryPoint(shaderStage);
```
trenouf, line 910: clang now gives me an unused private field warning on m_pipelineShaders. Could you remove the field please? (Unless you have a forthcoming change that starts using it again.)

xuechen417: Thank you for pointing out. I will remove it.

PR #2474, opened 2023-05-19
A simple optimization of GS transform feedback on GFX11

amdrexu:
If the output primitive of a GS stream is point, we don't need primitive compaction to get rid of incomplete primitive caused by GS_CUT. This is because each vertex from GS_EMIT always forms a point primitive and no incomplete primitive will be generated. Therefore, the primitive index is consecutive in a NGG subgroup. We can safely skip primitive compaction when trying to export vertices to XFB buffer for each primitive.


nhaehnle (commented):
Maybe I'm missing something, but isn't compaction still required because each GS invocation can produce a variable number of vertex emits?

amdrexu:
> Maybe I'm missing something, but isn't compaction still required because each GS invocation can produce a variable number of vertex emits?

Yeah. I saw a test failure. I will look into the case.

PR #2491, opened 2023-05-31
[LLPC] Freeze undef branch conditions in SPIRVReader

tsymalla:
LLVM upstream commit 1fc425380e9860a6beb53fa68d02e7fb14969963 can cause some intermediate failures when we rely on a switch with an undef branch condition not being optimized after InstCombine. This patch aims at freezing the branch condition so we restore the previous behavior.

### DIFF
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -4745,6 +4745,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *bv, Function *f, Bas
 
     auto trueSuccessor = cast<BasicBlock>(transValue(br->getTrueLabel(), f, bb));
     auto falseSuccessor = cast<BasicBlock>(transValue(br->getFalseLabel(), f, bb));
+
+    // OpBranchConditional can branch with OpUndef as condition. The selected jump target is undefined.
+    // In LLVM IR, BranchInst with undef value is fully UB.
+    // LLVM will mark the successors as unreachable, so later SimplifyCFG passes will cause
+    // unwanted removal of branches. By using a freeze instruction as condition, we make sure that LLVM
+    // will always operate on a fixed value.
+    c = new FreezeInst(c, "cond.freeze", bb);
+
     auto bc = BranchInst::Create(trueSuccessor, falseSuccessor, c, bb);
     auto lm = static_cast<SPIRVLoopMerge *>(br->getPrevious());
     if (lm && lm->getOpCode() == OpLoopMerge)
@@ -4836,6 +4844,14 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *bv, Function *f, Bas
     auto select = transValue(bs->getSelect(), f, bb);
     auto defaultSuccessor = dyn_cast<BasicBlock>(transValue(bs->getDefault(), f, bb));
     recordBlockPredecessor(defaultSuccessor, bb);
+
+    // OpSwitch can branch with OpUndef as condition. The selected jump target is undefined.
+    // In LLVM IR, SwitchInst with undef value is fully UB.
+    // LLVM will mark the successors as unreachable, so later SimplifyCFG passes will cause
+    // unwanted removal of branches. By using a freeze instruction as condition, we make sure that LLVM
+    // will always operate on a fixed value.
+    select = new FreezeInst(select, "cond.freeze", bb);
+
     auto ls = SwitchInst::Create(select, defaultSuccessor, bs->getNumPairs(), bb);
     bs->foreachPair([&](SPIRVSwitch::LiteralTy literals, SPIRVBasicBlock *label) {
       assert(!literals.empty() && "Literals should not be empty");
--- a/llpc/test/shaderdb/core/OpAny_TestBvec2_lit.frag
+++ b/llpc/test/shaderdb/core/OpAny_TestBvec2_lit.frag
@@ -30,7 +30,8 @@ void main()
 // CHECK-NEXT:    [[TMP4:%.*]] = extractelement <2 x i1> [[TMP3]], i64 0
 // CHECK-NEXT:    [[TMP5:%.*]] = extractelement <2 x i1> [[TMP3]], i64 1
 // CHECK-NEXT:    [[TMP6:%.*]] = or i1 [[TMP4]], [[TMP5]]
-// CHECK-NEXT:    [[SPEC_SELECT:%.*]] = select i1 [[TMP6]], <4 x float> <float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00>, <4 x float> <float 5.000000e-01, float 5.000000e-01, float 5.000000e-01, float 5.000000e-01>
+// CHECK-NEXT:    [[TMP7:%.*]] = freeze i1 [[TMP6]]
+// CHECK-NEXT:    [[SPEC_SELECT:%.*]] = select i1 [[TMP7]], <4 x float> <float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00>, <4 x float> <float 5.000000e-01, float 5.000000e-01, float 5.000000e-01, float 5.000000e-01>
 // CHECK-NEXT:    call void (...) @lgc.create.write.generic.output(<4 x float> [[SPEC_SELECT]], i32 0, i32 0, i32 0, i32 0, i32 0, i32 undef)
 // CHECK-NEXT:    ret void
 //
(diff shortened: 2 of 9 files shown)


amdvlk-admin:

amdvlk-admin:

nhaehnle (commented):
Can you explain what the unwanted side effects are? Ideally in the comment :)

It would be nice to change this code to using the builder while you're at it, looks like the case just above does this as well.

nhaehnle:
ok to test <-- should make @amdvlk-admin happy

tsymalla:
I'll update the tests once somebody took a look at the unconditional insertion of the `FreezeInst`.

dstutt:
Is the change here that you've modified it from unconditional use of freeze to only when the condition is undef? (Presumably that's also why no test updates are required?)

tsymalla:
> Is the change here that you've modified it from unconditional use of freeze to only when the condition is undef? (Presumably that's also why no test updates are required?)

Yes. Unconditionally using the freeze requires some more fixup of the IR.

dstutt:
I think adding it conditionally (on undef) is the best plan (I was just wanting clarification rather than suggesting a change).
Not sure why the Jenkins Pipeline check isn't being triggered.

tsymalla:
Thinking more about it, it would indeed be possible to have the condition referenced multiple times even with checking on undef, so I'll go with adding additional checks for `hasOneUse()` which should fix the issues for now.

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4686,6 +4686,18 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *bv, Function *f, Bas
...
     auto c = transValue(br->getCondition(), f, bb);
 
+    // OpBranchConditional can branch with OpUndef as condition. The selected jump target is undefined.
+    // In LLVM IR, BranchInst with undef value is fully UB.
+    // LLVM will mark the successors as unreachable, so later SimplifyCFG passes will cause
+    // unwanted removal of branches. By using a freeze instruction as condition, we make sure that LLVM
+    // will operate on a fixed value.
+    if (isa<UndefValue>(c)) {
+      auto builder = getBuilder();
+      IRBuilder<>::InsertPointGuard guard(*builder);
```
jayfoad, line 4696: Is this guard needed? I would not expect `CreateFreeze` to change the insertion point.

tsymalla:
I did some more changes to the implementation. When we step through the OpUndefs, the undef Value * will only be recorded once. Since we don't see all users during lowering of OpSwitch / OpBranchConditional, the check for having more than one user needs to happen after the translation. For now, this should only do anything if we have a single user, which should fix the tests. Indeed I believe something would be broken if we had more than one branch operating on an undef value.
Note that this explicitly checks for undef values since otherwise this would also apply on true / false atomic values, possibly appearing in other functions.

jayfoad:
> The issue with inserting the change unconditionally is that in some cases (like with a plain true or false branch condition), the value map will have the freeze instruction instead of the true / false. If the true or false is used in another function, which can happen because we don't yet inline here, the SSA is violated, as the IR references the freeze instruction from another function. We could
> 
> a) Move the change to a later pass where everything is inlined b) Clone the underlying condition and create a freezeinst on that c) Don't do anything if the condition has multiple users

How about d) insert the freeze before every use that is a branch or switch, but don't insert the freeze into the value map?

tsymalla:
> > The issue with inserting the change unconditionally is that in some cases (like with a plain true or false branch condition), the value map will have the freeze instruction instead of the true / false. If the true or false is used in another function, which can happen because we don't yet inline here, the SSA is violated, as the IR references the freeze instruction from another function. We could
> > a) Move the change to a later pass where everything is inlined b) Clone the underlying condition and create a freezeinst on that c) Don't do anything if the condition has multiple users
> 
> How about d) insert the freeze before every use that is a branch or switch, but don't insert the freeze into the value map?

Indeed more elegant, thanks for the suggestion :)

dstutt:
I think the Jenkins Pipeline internal job is currently broken (for other reasons).

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4698,6 +4698,20 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *bv, Function *f, Bas
...
     auto falseSuccessor = cast<BasicBlock>(transValue(br->getFalseLabel(), f, bb));
     auto bc = BranchInst::Create(trueSuccessor, falseSuccessor, c, bb);
+
+    // OpBranchConditional can branch with OpUndef as condition. The selected jump target is undefined.
+    // In LLVM IR, BranchInst with undef value is fully UB.
+    // LLVM will mark the successors as unreachable, so later SimplifyCFG passes will cause
+    // unwanted removal of branches. By using a freeze instruction as condition, we make sure that LLVM
+    // will operate on a fixed value.
+    if (isa<UndefValue>(c)) {
+      auto builder = getBuilder();
```
jayfoad, line 4708: Nit: I know Nicolai suggested using IRBuilder, but I think (hope) he meant converting all of the "case OpBranchConditional" to use it. Using it just for inserting this one instruction makes the code more complicated for no benefit.

tsymalla: Doesn't it make sense to use the builder to get the earliest insertion point after the allocas?

jayfoad: It is simpler to insert the freeze immediately before the branch.

tsymalla:
Removed builder usage again

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -4697,6 +4697,15 @@ Value *SPIRVToLLVM::transValueWithoutDecoration(SPIRVValue *bv, Function *f, Bas
...
     auto trueSuccessor = cast<BasicBlock>(transValue(br->getTrueLabel(), f, bb));
     auto falseSuccessor = cast<BasicBlock>(transValue(br->getFalseLabel(), f, bb));
+
+    // OpBranchConditional can branch with OpUndef as condition. The selected jump target is undefined.
+    // In LLVM IR, BranchInst with undef value is fully UB.
+    // LLVM will mark the successors as unreachable, so later SimplifyCFG passes will cause
+    // unwanted removal of branches. By using a freeze instruction as condition, we make sure that LLVM
+    // will operate on a fixed value.
+    if (isa<UndefValue>(c))
+      c = new FreezeInst(c, Twine("undef.cond.freeze"), bb);
```
jayfoad, line 4707: Nit: don't need `Twine`, just use the string literal.

jayfoad (approved):
Much nicer!

tsymalla:
Rebased (altough I don't see anything related in the last few commits).

qiaojbao:
> Rebased (altough I don't see anything related in the last few commits).

The new build passed, let's waiting the test result.

The previous error seems too many lines to insert into the comment message.

amdvlk-admin:
### Test summary for commit a7208026c1cd276eb91b9f7c83b7594655e0e308
<details><summary>CTS tests (Failed: 0/142050)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40426/71030 (56.9%)</li>
<li>Failed: 0/71030 (0.0%)</li>
<li>Not Supported: 30604/71030 (43.1%)</li><li>Warnings: 0/71030 (0.0%)</li>
</ul></details>

<li>Passed: 40464/71020 (57.0%)</li>
<li>Failed: 0/71020 (0.0%)</li>
<li>Not Supported: 30556/71020 (43.0%)</li><li>Warnings: 0/71020 (0.0%)</li>
</ul></details>


</ul></details>

tsymalla:
Using the freeze unconditionally generally seems to cause some changes in optimizing of floating point conditionals. In some cases, the freeze causes the compiler to remove certain fast-math flags like NNAN.

tsymalla:
Insert the freeze unconditionally for `OpBranchConditional` and `OpSwitch`. Update lit tests to show the difference in behavior.

amdvlk-admin:
### Test summary for commit a26ccc2f7ae8db1bd47e34e820ebbf016854f844
<details><summary>CTS tests (Failed: 0/142050)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40426/71030 (56.9%)</li>
<li>Failed: 0/71030 (0.0%)</li>
<li>Not Supported: 30604/71030 (43.1%)</li><li>Warnings: 0/71030 (0.0%)</li>
</ul></details>

<li>Passed: 40464/71020 (57.0%)</li>
<li>Failed: 0/71020 (0.0%)</li>
<li>Not Supported: 30556/71020 (43.0%)</li><li>Warnings: 0/71020 (0.0%)</li>
</ul></details>


</ul></details>

jayfoad (approved):
I suggest committing this patch as-is since it is blocking some of our internal merge jobs and the changes in fast math flags look pretty minor to me. Then do an experiment (or file a ticket so we remember to do it): compile all of shadertest with and without the `freeze`s and see if there is any consistent missed optimization in the final ISA. If there is, then there is probably some part of the LLVM middle or back end that needs to be taught how to look through `freeze`s.

tsymalla:
> I suggest committing this patch as-is since it is blocking some of our internal merge jobs and the changes in fast math flags look pretty minor to me. Then do an experiment (or file a ticket so we remember to do it): compile all of shadertest with and without the `freeze`s and see if there is any consistent missed optimization in the final ISA. If there is, then there is probably some part of the LLVM middle or back end that needs to be taught how to look through `freeze`s.

I agree, let us just merge this for now.

dstutt (approved):
Based on the last comments - approving

nhaehnle (approved):

tsymalla:
Thanks! Can somebody take care of merging please?

PR #2500, opened 2023-06-05
Stop using llvm.fmuladd

jayfoad:
The llvm.fmuladd instrinsic represents an fmul and an fadd that may be fused with each other but may not be fused with other instructions. This is overly restrictive for SPIR-V. Using separate fmul and fadd instructions and marking them both as contractable gives LLVM more flexibility to optimize.


jayfoad:
In fact I think an `fmuladd` that is already marked contractable is completely pointless and should be canonicalized to separate instructions anyway. I have an upstream LLVM patch for that: https://reviews.llvm.org/D152166

nhaehnle (approved):
Yeah, that does sound plausible.

PR #2505, opened 2023-06-08
Ignore vgprLimit setting if smaller than entrypoint return type.

perlfu:
Specifically if the size of return type and callee saved registers of an RT shader is bigger than the VGPR limit then we need to drop the limit entirely (or expand it).
If we don't do this then backend compilation is impossible.

### DIFF
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -64,6 +64,7 @@
 #include "lgc/state/TargetInfo.h"
 #include "lgc/util/AddressExtender.h"
 #include "lgc/util/BuilderBase.h"
+#include "lgc/util/Debug.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Analysis/AliasAnalysis.h" // for MemoryEffects
 #include "llvm/Support/CommandLine.h"
@@ -894,6 +895,16 @@ void PatchEntryPointMutate::setFuncAttrs(Function *entryPoint) {
 
   unsigned vgprLimit = shaderOptions->vgprLimit;
   unsigned sgprLimit = shaderOptions->sgprLimit;
+
+  if (vgprLimit != 0 && entryPoint->getReturnType()->isSized()) {
+    const unsigned CalleeSaveRegsMax = 56; // FIXME: can we compute this?
+    const DataLayout &dataLayout = entryPoint->getParent()->getDataLayout();
+    const uint64_t returnSize = dataLayout.getTypeAllocSize(entryPoint->getReturnType()) / 4;
+    if ((vgprLimit - CalleeSaveRegsMax) < returnSize) {
+      LLPC_OUTS("Ignoring vgprLimit setting due to return size of " << returnSize << " dwords\n");
+      vgprLimit = 0;
+    }
+  }
 
   if (vgprLimit != 0) {
     builder.addAttribute("amdgpu-num-vgpr", std::to_string(vgprLimit));


amdvlk-admin:
### Test summary for commit 2c77f118126bf9474560381fd11cbc0c6f23e4dd
<details><summary>CTS tests (Failed: 0/141320)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40065/70667 (56.7%)</li>
<li>Failed: 0/70667 (0.0%)</li>
<li>Not Supported: 30602/70667 (43.3%)</li><li>Warnings: 0/70667 (0.0%)</li>
</ul></details>

<li>Passed: 40106/70653 (56.8%)</li>
<li>Failed: 0/70653 (0.0%)</li>
<li>Not Supported: 30547/70653 (43.2%)</li><li>Warnings: 0/70653 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit 2c77f118126bf9474560381fd11cbc0c6f23e4dd
<details><summary>CTS tests (Failed: 0/141308)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 40067/70655 (56.7%)</li>
<li>Failed: 0/70655 (0.0%)</li>
<li>Not Supported: 30588/70655 (43.3%)</li><li>Warnings: 0/70655 (0.0%)</li>
</ul></details>

<li>Passed: 40106/70653 (56.8%)</li>
<li>Failed: 0/70653 (0.0%)</li>
<li>Not Supported: 30547/70653 (43.2%)</li><li>Warnings: 0/70653 (0.0%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -895,6 +896,16 @@ void PatchEntryPointMutate::setFuncAttrs(Function *entryPoint) {
   unsigned vgprLimit = shaderOptions->vgprLimit;
   unsigned sgprLimit = shaderOptions->sgprLimit;
 
+  if (vgprLimit != 0 && entryPoint->getReturnType()->isSized()) {
+    const unsigned CalleeSaveRegsMax = 56; // FIXME: can we compute this?
+    const DataLayout &dataLayout = entryPoint->getParent()->getDataLayout();
+    const uint64_t returnSize = dataLayout.getTypeAllocSize(entryPoint->getReturnType()) / 4;
+    if ((vgprLimit - CalleeSaveRegsMax) < returnSize) {
```
Flakebi, line 903: What is the logic behind this and where does the 56 come from?

Doesn’t it work to put the return value into return-VGPRs and then restore callee-saves?
If that worked an `if (vgprLimit < returnSize)` should be enough.

perlfu: 56 is large enough value to solve the problem I am seeing based on observation (with some buffer), which is why I want to replace it with some kind of computed estimate.
The shader I am looking at returns 120 VGPRs, and has 47 CSRs at the point of return.
So under our typical 128 VGPR limit on Navi2 there is no way to compile this regardless of ordering.

If we cannot estimate this in the front-end then I'll build something in the backend to ignore the limit instead, perhaps that is going to be more reliable -- although it breaks current convention that set vgpr limits are hard.

nhaehnle (commented):
I assume this comes from an RT pipeline with huge payload sizes? We'll need a real fix for that in the long run, but we're implicitly working on that already, and this may be a reasonable interim fix.

The one thing that makes me slightly nervous about this is that it's not clear to me where/how the actual VGPR allocation programmed on the hardware actually comes from. Have you tested that whatever application you're looking at actually runs correctly with this change?

lgc/patch/PatchEntryPointMutate.cpp
Flakebi: I tried compiling such a function and the generated code is indeed incorrect. But removing the VGPR limit doesn’t make it correct either.

Callee-save registers apparently take priority over return values, so part of the return value is overwritten with callee-saves.
So, independent of the VGPR limit, the return value is incorrect.

I think the correct solution is to make the return value VGPRs match the argument VGPRs, i.e. use only `v[0:31]` for return values. That way they won’t collide with callee-saves, which start at v40 (and are striped until v255).

<details>
<summary>Test shader</summary>
```llvm
; Define a compute compute library function with a large payload and a tight VGPR limit.

; RUN: lgc -mcpu=gfx1010 -o - - <%s | FileCheck %s

; ModuleID = 'lgcPipeline'
target datalayout = "e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-ni:7"
target triple = "amdgcn--amdpal"

define spir_func <120 x i32> @compute_library_func(<120 x i32> %arg) #2 !lgc.shaderstage !7 {
  ret <120 x i32> %arg
}

; Function Attrs: nounwind readonly
declare i8 addrspace(7)* @lgc.create.load.buffer.desc.p7i8(...) local_unnamed_addr #1

attributes #0 = { nounwind }
attributes #1 = { nounwind readonly }
attributes #2 = { "amdgpu-num-vgpr"="128" }

!llpc.compute.mode = !{!0}
!lgc.options = !{!1}
!lgc.options.CS = !{!2}
!lgc.user.data.nodes = !{!3, !4, !5, !6}

!0 = !{i32 2, i32 3, i32 1}
!1 = !{i32 2113342239, i32 1385488414, i32 -1007072744, i32 -815526592, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 2}
!2 = !{i32 1792639877, i32 1348715323, i32 0, i32 0, i32 0, i32 100, i32 0, i32 0, i32 0, i32 0, i32 0, i32 64, i32 0, i32 0, i32 3}
!3 = !{!"DescriptorBuffer", i32 6, i32 0, i32 4, i32 0, i32 0, i32 4}
!4 = !{!"DescriptorBuffer", i32 6, i32 4, i32 16, i32 0, i32 1, i32 4}
!5 = !{!"DescriptorTableVaPtr", i32 0, i32 20, i32 1, i32 1}
!6 = !{!"DescriptorBuffer", i32 6, i32 0, i32 4, i32 0, i32 2, i32 4}
!7 = !{i32 7}
```
</details>

perlfu: I can see the failure in your example shader, and this might account for issues I am seeing in an end application.
However I'm still trying to make sense of your explanation, partly due to my incomplete knowledge of our current ABI for this.

What you're saying is that under our current scheme, any return larger than 8 VGPRs (32:39), is clobbered by CSRs which start at v40?
At best our maximum return size is 40 VGPRs if we moved the range to start at v0?

perlfu: After some poking I can see the _current_ limit is 40 VGPRs for a return (without clobbering occuring)?

By the following:
> I think the correct solution is to make the return value VGPRs match the argument VGPRs, i.e. use only v[0:31] for return values.

Did you mean that like call arguments, return values above v31 should be passed via the stack?

nhaehnle: Yes, that would lead to a consistent ABI and is probably the easiest fix.

Longer term, I think it would make sense to have return values take precedence over callee-saves status. That is, if vN is usually callee-saves but there are many return values, then vN may be used as a return value instead of callee-saves. But that feels like it may be harder to do? (I'm not that familiar with the calling convention lowering code in the backend.)

perlfu: https://reviews.llvm.org/D152892

perlfu:
This has been address by:
- https://reviews.llvm.org/D152892 (Preventing the clobber of large returns by callee saves)
- https://reviews.llvm.org/D152912 (Making return lowering respect VGPR limit)

PR #2554, opened 2023-06-25
Dialect-ify SPIR-V RT Ops

LLJJDD:
Map SPIR-V ray tracing related ops to lgc.rt dialect.

This commit does not include OpTraceRay, as we will need some extra handling on it.

### DIFF
--- a/llpc/context/llpcContext.cpp
+++ b/llpc/context/llpcContext.cpp
@@ -30,6 +30,7 @@
  */
 #include "llpcContext.h"
 #include "SPIRVInternal.h"
+#include "lgcrt/LgcRtDialect.h"
 #include "llpcCompiler.h"
 #include "llpcDebug.h"
 #include "llpcPipelineContext.h"
@@ -66,7 +67,7 @@ namespace Llpc {
 //
 // @param gfxIp : Graphics IP version info
 Context::Context(GfxIpVersion gfxIp) : LLVMContext(), m_gfxIp(gfxIp) {
-  m_dialectContext = llvm_dialects::DialectContext::make<LgcDialect, GpurtDialect>(*this);
+  m_dialectContext = llvm_dialects::DialectContext::make<LgcDialect, GpurtDialect, rt::LgcRtDialect>(*this);
 
   reset();
 }
--- a/llpc/lower/llpcSpirvLowerRayTracing.cpp
+++ b/llpc/lower/llpcSpirvLowerRayTracing.cpp
@@ -32,11 +32,13 @@
 #include "llpcSpirvLowerRayTracing.h"
 #include "SPIRVInternal.h"
 #include "gpurt-compiler.h"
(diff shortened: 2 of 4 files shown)


amdvlk-admin:
### Test summary for commit 36f152a13efd15356cbeadb3b57da66bda7b0362
<details><summary>CTS tests (Failed: 0/211243)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 39023/70408 (55.4%)</li>
<li>Failed: 0/70408 (0.0%)</li>
<li>Not Supported: 31385/70408 (44.6%)</li><li>Warnings: 0/70408 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70410 (55.5%)</li>
<li>Failed: 0/70410 (0.0%)</li>
<li>Not Supported: 31352/70410 (44.5%)</li><li>Warnings: 0/70410 (0.0%)</li>
</ul></details>

<li>Passed: 35147/70425 (49.9%)</li>
<li>Failed: 0/70425 (0.0%)</li>
<li>Not Supported: 35278/70425 (50.1%)</li><li>Warnings: 0/70425 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
LGTM for the refactoring

Looking through the code, I suspect that there's a lurking bug in callable shader handling. The args to the callable shader are:

1. Callable payload (passed by value, using a per-pipeline type)
2. Shader record index

The bug is around what happens when there are different callable payload types in play. Consider:

* Pipeline library #1 contains a callable shader with payload type T1
* Pipeline library #2 contains a callable shader with payload type T2
* Final pipeline uses both libraries and has rgen shader which calls both callable shaders, using the same per-pipeline payload type U

I really don't think this can work if T1 and T2 are of a different size -- U will be compatible with the larger one.

Since it's a pre-existing problem it shouldn't block this PR. I think the solution would be:

* Put the shader record index first
* Always use a payload size that matches the `paramDataSizeInBytes` of the current call in question
* (Maybe) Add an optional workaround for games that use callable shaders incorrectly with mismatching payload types

jasilvanus:
> * Pipeline library 1 contains a callable shader with payload type T1
> * Pipeline library 2 contains a callable shader with payload type T2
> * Final pipeline uses both libraries and has rgen shader which calls both callable shaders, using the same per-pipeline payload type U

Can you elaborate a bit more? Do you mean that the RayGen contains two `CallShader(idx, payload)` calls with a payload of type `U`, where `U` is the same for both CallShader calls, and `U` is *different* from either one of (or both) `T1` and `T2`?

If so, isn't this UB?

nhaehnle:
No, the application shader has two calls, one with type T1 and the other with type T2. The problem is that the current implementation uses type U.

LLJJDD:
I see, this is indeed something need to be addressed.

PR #2561, opened 2023-06-27
Fix compiling errors on windows

jiaolu:
Windows reports "Value* var" should be "Value& var" error

### DIFF
--- a/lgc/patch/LowerDebugPrintf.cpp
+++ b/lgc/patch/LowerDebugPrintf.cpp
@@ -90,10 +90,9 @@ void LowerDebugPrintf::visitDebugPrintf(DebugPrintfOp &op) {
   SmallVector<Value *> printArgs;
   // Records printf output variables are 64bit or not
   SmallBitVector bit64Vector;
-  for (Value *var : op.getArgs()) {
-    getDwordValues(var, printArgs, bit64Vector, builder);
+  for (auto var = op.arg_begin() + 2; var != op.arg_end(); ++var) {
+    getDwordValues(*var, printArgs, bit64Vector, builder);
   }
-
   GlobalVariable *globalStr = cast<GlobalVariable>(op.getFormat());
   StringRef strDebugStr = (cast<ConstantDataSequential>(globalStr->getInitializer()))->getAsString();
 


lgc/patch/LowerDebugPrintf.cpp
```diff
@@ -90,10 +90,9 @@ void LowerDebugPrintf::visitDebugPrintf(DebugPrintfOp &op) {
   SmallVector<Value *> printArgs;
   // Records printf output variables are 64bit or not
   SmallBitVector bit64Vector;
-  for (Value *var : op.getArgs()) {
```
Flakebi, line 93: The windows compilation error is probably because the iterator type on Linux is a pointer, but on Windows it’s a different type.
Can you use `auto var : op.getArgs()`?

jiaolu: auto is Value& on windows. Value* on linux

we still need to get the different var to the 
getDwordValues

Flakebi: If it’s really such different types (which makes `getArgs` unusable), this needs to be fixed in llvm-dialects instead of working around it here.

jiaolu: > If it’s really such different types (which makes `getArgs` unusable), this needs to be fixed in llvm-dialects instead of working around it here.

right. i am not familiar with llvm-dialects.   from LgcDialect.cpp.inc
i think 
llvm::iterator_range<llvm::Value*> should be llvm::iterator_range<User::op_iterator>

but i don't know how  to correct this in llvm-dialect.

so i figure having a fix is better than no fix.

jiaolu: llvm::iterator_range is very simple implementation with begin and end().  so it must use with op_iterator.
So even the linux can compile and run. i guess something could be wrong with many var aguments.

amdvlk-admin:
### Test summary for commit 05b060be45149a3cf06635993fc80dfea0ecec28
<details><summary>CTS tests (Failed: 0/211258)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 39024/70423 (55.4%)</li>
<li>Failed: 0/70423 (0.0%)</li>
<li>Not Supported: 31399/70423 (44.6%)</li><li>Warnings: 0/70423 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70410 (55.5%)</li>
<li>Failed: 0/70410 (0.0%)</li>
<li>Not Supported: 31352/70410 (44.5%)</li><li>Warnings: 0/70410 (0.0%)</li>
</ul></details>

<li>Passed: 35147/70425 (49.9%)</li>
<li>Failed: 0/70425 (0.0%)</li>
<li>Not Supported: 35278/70425 (50.1%)</li><li>Warnings: 0/70425 (0.0%)</li>
</ul></details>


</ul></details>

lgc/patch/LowerDebugPrintf.cpp
nhaehnle: Has the local checkout of the llvm-dialects submodule been updated? Maybe it's still using an old version, as this problem was fixed recently...

If the submodule *has* been updated, maybe there's a missing dependency in the cmake?

dstutt: Presumably the change went in to the GPUOpen llvm-dialects (since that's the definitive version)?

PR #2572, opened 2023-07-04
Unify cps function exits

ruiling:


amdvlk-admin:
### Test summary for commit d66c03d2b96e5a41599b829e6043e3c74b6b0281
<details><summary>CTS tests (Failed: 0/211353)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70456 (49.9%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 35309/70456 (50.1%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70456 (55.4%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 31430/70456 (44.6%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70441 (55.4%)</li>
<li>Failed: 0/70441 (0.0%)</li>
<li>Not Supported: 31383/70441 (44.6%)</li><li>Warnings: 0/70441 (0.0%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -265,6 +266,13 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
   if (!isCpsFunc && cpsOps.empty())
     return false;
 
+  // Get all the return instructions.
+  SmallVector<ReturnInst *> retInstrs;
+  for (BasicBlock &block : *func)
+    if (ReturnInst *ret = dyn_cast<ReturnInst>(block.getTerminator()))
+      retInstrs.push_back(ret);
```
nhaehnle, line 273: Instead of first collecting all `ret`s and then using the list later, just iterate over the `ret`s directly in the loop below.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+
+  unsigned vgprNum = 0;
+  for (const auto &exit : exitInfos)
+    vgprNum = exit.vgpr.size() > vgprNum ? exit.vgpr.size() : vgprNum;
+
+  SmallVector<Value *> newVgpr;
+  builder.SetInsertPoint(tailBlock);
+
+  if (exitInfos.size() == 1) {
+    newVgpr.append(exitInfos[0].vgpr);
```
nhaehnle, line 314: This could just be a direct assignment, right?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+  }
+  // Packing VGPR arguments.
+  Value *vgprArg = mergeIntoStruct(builder, newVgpr);
+
+  // Packing SGPR arguments(user data) into vector of dwords.
+  SmallVector<Value *> userData;
+  for (unsigned idx = 0; idx != numUserdata; ++idx)
+    userData.push_back(func->getArg(idx));
+
+  SmallVector<Value *> userDataDwords;
```
nhaehnle, line 337: Suggestion: call it `userDataI32` for consistency?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+  Value *vcrWwm;
+  if (isCpsFunc) {
+    // FIXME: Switch to LLVM intrinsic
+    FunctionType *setInactiveChainTy = FunctionType::get(vcrTy, {vcrTy, vcrTy}, false);
+    auto setInactiveChain = Function::Create(setInactiveChainTy, GlobalValue::ExternalLinkage,
+                                             "llvm.amdgcn.setinactive.chain.arg", func->getParent());
+    vcrWwm = builder.CreateCall(setInactiveChain, {vcr, func->getArg(numUserdata)});
+  } else {
+    vcrWwm = builder.CreateIntrinsic(Intrinsic::amdgcn_set_inactive, vcrTy, {vcr, builder.getInt32(0)});
+  }
```
nhaehnle, line 364: The only case where we're not in a CPS function is the entry shader function. That one doesn't need any of the WWM logic.

Ideally, we would even recognize if `vcr` is uniform, but that's more difficult to pull off. We can postpone addressing that to a later time.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -404,41 +530,15 @@ void PatchEntryPointMutate::lowerCpsJump(Function *parent, cps::JumpOp *jumpOp,
...
-#endif
+  splitIntoI32(layout, builder, remainingArgs, vgprArgs);
+
+  // Fill exit information.
+  exitInfos.push_back(CpsExitInfo(jumpOp->getParent(), std::move(vgprArgs)));
+  // Branch to tailBlock.
+  auto *oldTerm = jumpOp->getParent()->getTerminator();
+  assert(isa<UnreachableInst>(oldTerm));
+  oldTerm->eraseFromParent();
+  builder.CreateBr(tailBlock);
```
nhaehnle, line 541: This is probably fine in practice, but a more robust solution for this sort of thing (without relying on the `unreachable` being there) is:

* Split the basic block after the `jumpOp`
* Insert the branch instruction

The tail of the old basic block remains as an unreachable basic block, to be cleaned up by a later pass. An unreachable basic block is dominated by everything, so even if there were other instructions in the tail that use some values, they would always be in SSA form.

nhaehnle (commented):
Very nice. I have some inline comments but this mostly LGTM

lgc/patch/PatchEntryPointMutate.cpp
ruiling: This is used to handle the case that the entry point shader was launched with inactive threads in a wave. I think this is possible in Vulkan compute shader? In this case, the inactive lanes of vcr should be cleared to zero to make scheduler work correctly. Sounds reasonable?

lgc/patch/PatchEntryPointMutate.cpp
ruiling: I am assuming the passes before should generate a valid form of cps.jump followed immediately by unreachable. The assertion will check against this. If there is any meaningful instruction happen after the cps.jump other than unreachable. Then I would assume there might be something wrong happened. So current code might help catch possible errors. Do you have any possible shape of valid IR but breaking the assumption here? I agree your suggestion could work in more cases. But it may be more useful to catch possible broken cases for now.

nhaehnle: No, I can't think of any plausible reason why we would end up with valid IR that doesn't fit the pattern. So yeah, I'm fine with that.

lgc/patch/PatchEntryPointMutate.cpp
nhaehnle: Yes, compute shaders can be launched with inactive threads in a wave, but overall this still wouldn't work:

* The entry point shader would jump/chain to the correct place
* However, `llvm.amdgcn.cs.chain` would still leave the values in inactive lanes of VGPRs unchanged. So `vcr` in the next CPS function would contain garbage in inactive lanes

I don't see an easy way to make this work, and it's not necessary in practice for what we want to do. And just not using any form of `set.inactive` should make the assembly code in the entry point shader a bit cheaper.

ruiling: >     * However, `llvm.amdgcn.cs.chain` would still leave the values in inactive lanes of VGPRs unchanged. So `vcr` in the next CPS function would contain garbage in inactive lanes

I get the point that `llvm.amdgcn.cs.chain` call will leave the inactive lanes unchanged, but how the vcr in the next cps function contain garbage if we clear lanes (by doing amdgcn.set.inactive) that were inactive at shader dispatch? All the lanes would have been set with meaningful data(either zero or target cps vcr).

lgc/include/lgc/patch/PatchEntryPointMutate.h
```diff
@@ -139,10 +139,18 @@ class PatchEntryPointMutate : public Patch, public llvm::PassInfoMixin<PatchEntr
 
   uint64_t pushFixedShaderArgTys(llvm::SmallVectorImpl<llvm::Type *> &argTys) const;
 
+  // Information of each cps exit(return or cps.jump) used for exit unification.
```
tsymalla-AMD, line 142: Insert whitespace after "exit"

tsymalla-AMD: "of" => "about" ?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
     numUserdata = m_cpsShaderInputCache.getTypes().size();
   }
 
+  SmallVector<CpsExitInfo> exitInfos;
+  const DataLayout &layout = func->getParent()->getDataLayout();
```
tsymalla-AMD, line 287: Can move that line a bit to the bottom, since `layout` is first used ~50 lines later if I'm not mistaken

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+    exitInfos.push_back(
+        CpsExitInfo(ret->getParent(), {builder.getInt32(0), PoisonValue::get(vspTy)}));
+    builder.SetInsertPoint(ret);
+    builder.CreateBr(tailBlock);
+    ret->eraseFromParent();
+  }
+
+  unsigned vgprNum = 0;
+  for (const auto &exit : exitInfos)
+    vgprNum = exit.vgpr.size() > vgprNum ? exit.vgpr.size() : vgprNum;
```
tsymalla-AMD, line 308: Use `std::max`?

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +283,126 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+        else
+          phi->addIncoming(PoisonValue::get(builder.getInt32Ty()), exitInfos[exitIdx].pred);
+      }
+      newVgpr.push_back(phi);
+    }
+  }
+  // Packing VGPR arguments.
+  Value *vgprArg = mergeIntoStruct(builder, newVgpr);
+
+  // Packing SGPR arguments(user data) into vector of dwords.
```
tsymalla-AMD, line 332: Missing whitespace after "arguments"

tsymalla-AMD: Did you mean to use "i32s" here?

tsymalla-AMD (changes requested):

lgc/patch/PatchEntryPointMutate.cpp
ruiling: I find that iterating over the ret's later will visit the newly created tail block (have no terminator) which would need explicitly by-passing null terminator. So I just keep it as now. Please let me know if you still prefer to merge the two loops together.

amdvlk-admin:
### Test summary for commit 9d5dc3ffab149115a85db28aa157f5b3db3b331a
<details><summary>CTS tests (Failed: 0/211353)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70456 (49.9%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 35309/70456 (50.1%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70456 (55.4%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 31430/70456 (44.6%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70441 (55.4%)</li>
<li>Failed: 0/70441 (0.0%)</li>
<li>Not Supported: 31383/70441 (44.6%)</li><li>Warnings: 0/70441 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit dd0169d3fab2bcc70404c63d2945a9622b60c187
<details><summary>CTS tests (Failed: 0/211353)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70456 (49.9%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 35309/70456 (50.1%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70456 (55.4%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 31430/70456 (44.6%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70441 (55.4%)</li>
<li>Failed: 0/70441 (0.0%)</li>
<li>Not Supported: 31383/70441 (44.6%)</li><li>Warnings: 0/70441 (0.0%)</li>
</ul></details>


</ul></details>

lgc/patch/PatchEntryPointMutate.cpp
nhaehnle: What `amdgcn.set.inactive` does conceptually only affects the SSA value that is defined by the `amdgcn.set.inactive`. It doesn't affect what's in the "physical" VGPR holding the vcr argument.

That's why the calculation in the entry point shader itself would be correct, but then the values are again undefined in the following CPS function.

lgc/patch/PatchEntryPointMutate.cpp
nhaehnle: Okay, makes sense.

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -275,9 +277,134 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func) {
...
+                         builder.getInt32Ty()};
+  FunctionType *chainFuncTy = FunctionType::get(builder.getVoidTy(), chainArgTys, true);
+  auto chainFunc =
+      Function::Create(chainFuncTy, GlobalValue::ExternalLinkage, "llvm.amdgcn.cs.chain", func->getParent());
+  builder.CreateCall(chainFunc, chainArgs);
+#else
+  // New version of the code (also handles unknown version, which we treat as
+  // latest)
+  Type *chainTys[] = {builder.getPtrTy(), builder.getIntNTy(waveSize), userDataVec->getType(), vgprArg->getType()};
+  builder.CreateIntrinsic(Intrinsic::amdgcn_cs_chain, chainTys, chainArgs);
```
nhaehnle, line 405: Use the overload of `CreateIntrinsic` that doesn't require explicitly passing the overload set. It's the one where you explicitly pass `void` as return type, and then the rest should be inferred from `chainArgs`.

nhaehnle (commented):
Thanks, this looks good to me now. Only one nit I just noticed inline.

lgc/patch/PatchEntryPointMutate.cpp
ruiling: The code was contributed by @jayfoad . Will that version mangle or not mangle the intrinsic name? I would leave the change later when the LLVM backend change was propagated to github. So I can replace it and update the test check as well.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
```diff
@@ -0,0 +1,202 @@
...
+  call void (...) @lgc.cps.jump(i32 %cr.then, i32 2, {i32} %state.then, i32 %then.arg)
+  unreachable
+
+else:
+  ret void
+}
+
+!0 = !{i32 1} ; level 1
+; CHECK-LABEL: define {{[^@]+}}@unify_jumps.1
+; CHECK-SAME: (i32 inreg [[GLOBALTABLE:%.*]], i32 inreg [[PERSHADERTABLE:%.*]], ptr addrspace(4) inreg [[NUMWORKGROUPSPTR:%.*]], i32 inreg [[TMP0:%.*]], i32 inreg [[TMP1:%.*]], i32 inreg [[TMP2:%.*]], i32 inreg [[TMP3:%.*]], i32 inreg [[TMP4:%.*]], i32 inreg [[TMP5:%.*]], i32 inreg [[TMP6:%.*]], i32 inreg [[TMP7:%.*]], i32 inreg [[TMP8:%.*]], i32 inreg [[TMP9:%.*]], i32 inreg [[TMP10:%.*]], i32 inreg [[SPILLTABLE:%.*]], i32 [[VCR:%.*]], ptr addrspace(5) [[VSP:%.*]], i32 [[ARG:%.*]], ptr [[TABLE:%.*]]) #[[ATTR1:[0-9]+]] align 64 !lgc.cps !2 {
```
ruiling, line 50: The CI test seems failed because the metadata number changed from `!2` to `!3` on the CI machine. And I even cannot reproduce the failure locally... I have experienced such issue when I ported the change from internal, at that time the number changed from `!0` to `!2`. So anybody know a good way to handle such case. It is annoying that the number keep changing.

ruiling: The more strange thing is the test only fails for clang, but passed for gcc...

tsymalla-AMD: You can try to capture the number with ![[name:[0-9]+]] and then check in the same test run that the captured node is equivalent to some value with

CHECK: ![[name]] = !...

This requires to run different tests with other names and a bit of tinkering.

jasilvanus: I think in this particular case, we are not interested in the metadata contents, it's rather a UTC-generated test that includes MD node ids, which seem to change for some reason.

We need to understand why they are changing. If it cannot be avoid, we could add a UTC flag to ignore MD node ids, but clang and gcc already disagree this sounds more like a bug somewhere.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
```diff
@@ -0,0 +1,202 @@
...
+  %state.then = insertvalue {i32} poison, i32 %v.then, 0
+  call void (...) @lgc.cps.jump(i32 %cr.then, i32 2, {i32} %state.then, i32 %then.arg)
+  unreachable
+
+else:
+  ret void
+}
+
+!0 = !{i32 1} ; level 1
+; CHECK-LABEL: define {{[^@]+}}@unify_jumps.1
```
Flakebi, line 49: Not part of this patch but the cps lowering should call `takeName` and others (especially `insertAfter` for function order) that are also called in `lgc::addFuncitonArgs`.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
Flakebi: I only see 3 metadata entries, so I wonder where another one is coming from.
```llvm
!0 = !{!""}
!1 = !{!"\82\B0amdpal.pipelines\91\83\AA.registers\83\CD.@\CE\10\00\00\00\CD.B\CE\10\00\00\06\CD.O\CE\10\00\00\02\B0.spill_threshold\CE\FF\FF\FF\FF\B0.user_data_limit\00\AEamdpal.version\92\02\06"}
!2 = !{i32 1}
```

We really want deterministic compilation (and an internal CI job checks that at least the outcoming ELF is consistent) :)

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
ruiling: Yes, agree the takeName should be there. Do we need function order for particular reason or just make reading print-after log easier? Others should be covered by `copyMetadata()` and `copyAttributesFrom()`.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
ruiling: The weird extra metadata in clang build comes from the `m_options` in PipelineState (https://github.com/GPUOpen-Drivers/llpc/blob/dev/lgc/include/lgc/state/PipelineState.h#L555). The sizeof(Options) is 128 bytes, actually the last field is at offset 122, so there are 6 tail padding bytes. The brace initialization sounds like does not initialize the padding 6bytes in clang build. But it was read inside `setNamedMetadataToArrayOfInt32`. That's where `i32 -16777216, i32 -1` comes from.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
Flakebi: I think it’s helpful to preserve function order when writing tests with multiple functions (e.g. it allows auto-generating function-CHECKs).
So if it doesn’t hurt, I would add it :)

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
nhaehnle: Yes, I think the solution is to add explicit padding everywhere in these structures.

ruiling: I think maintaining explicitly padded struct is annoying. I have just made a change #2584 following the way ShaderOptions being zero initialized.

lgc/patch/PatchEntryPointMutate.cpp
ruiling: I tried, but it does not work as the other version of `CreateIntrinsic` has limitation in supporting vararg intrinsic. I would leave this as future work.

lgc/test/Transforms/CpsLowering/cps-unify-exits.lgc
ruiling: Fixed in new commit. Thanks!

lgc/patch/PatchEntryPointMutate.cpp
jayfoad: > the other version of CreateIntrinsic has limitation in supporting vararg intrinsic

Sorry, I did try but I failed: https://reviews.llvm.org/D130776#3691034
It would be great if someone else could implement that.

nhaehnle (approved):

amdvlk-admin:
### Test summary for commit b38bf32a178e11f5eaaa335b1f07441b0ddf9638
<details><summary>CTS tests (Failed: 0/211353)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70456 (49.9%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 35309/70456 (50.1%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70456 (55.4%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 31430/70456 (44.6%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70441 (55.4%)</li>
<li>Failed: 0/70441 (0.0%)</li>
<li>Not Supported: 31383/70441 (44.6%)</li><li>Warnings: 0/70441 (0.0%)</li>
</ul></details>


</ul></details>

PR #2584, opened 2023-07-11
lgc: memset Options to avoid garbage in padding bits

ruiling:
Options will be written into LLVM metadata as i32. Explicitly memset it will help avoid unexpected padding bits being written into metadata.

### DIFF
--- a/lgc/include/lgc/state/PipelineState.h
+++ b/lgc/include/lgc/state/PipelineState.h
@@ -552,7 +552,7 @@ private:
   bool m_preRasterHasGs = false;                        // Whether pre-rasterization part has a geometry shader
   bool m_computeLibrary = false;                        // Whether pipeline is in fact a compute library
   std::string m_client;                                 // Client name for PAL metadata
-  Options m_options = {};                               // Per-pipeline options
+  Options m_options;                                    // Per-pipeline options
   std::vector<ShaderOptions> m_shaderOptions;           // Per-shader options
   std::unique_ptr<ResourceNode[]> m_allocUserDataNodes; // Allocated buffer for user data
   llvm::ArrayRef<ResourceNode> m_userDataNodes;         // Top-level user data node table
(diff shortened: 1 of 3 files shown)


amdvlk-admin:
### Test summary for commit d08768735947c9b446e093672962bd18c62991de
<details><summary>CTS tests (Failed: 0/211353)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70456 (49.9%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 35309/70456 (50.1%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70456 (55.4%)</li>
<li>Failed: 0/70456 (0.0%)</li>
<li>Not Supported: 31430/70456 (44.6%)</li><li>Warnings: 0/70456 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70441 (55.4%)</li>
<li>Failed: 0/70441 (0.0%)</li>
<li>Not Supported: 31383/70441 (44.6%)</li><li>Warnings: 0/70441 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
Hmm yes. This entire thing makes me slightly uncomfortable, and I would have added padding member variables, but I can see how the explicit memset/memcpy is a solution that probably works just as robustly and actually has less risk of regression if somebody adds new fields to the struct.

lgc/include/lgc/state/PipelineState.h
```diff
@@ -552,7 +552,7 @@ class PipelineState final : public Pipeline {
   bool m_preRasterHasGs = false;                        // Whether pre-rasterization part has a geometry shader
   bool m_computeLibrary = false;                        // Whether pipeline is in fact a compute library
   std::string m_client;                                 // Client name for PAL metadata
-  Options m_options = {};                               // Per-pipeline options
```
jasilvanus, line 555: `Options m_options = Options()` should also work, because this value-initializes instead of aggregate initialization [1]. In my local test, this worked with both gcc and clang, while `= {}` did not initialize padding with clang.

However, even if this works now, I feel the memset variant is more robust.

[1]: https://stackoverflow.com/questions/70979077/does-c-standard-guarantee-the-initialization-of-padding-bytes-to-zero-for-non

ruiling: But the stackoverflow link says the `Foo()` does not work for g++, right?

jasilvanus: Seems to be a bug, so yeah, let's not depend on this.

PR #2597, opened 2023-07-19
Freeze shift instructions

jayfoad:
In SPIR-V the result of a shift with an out-of-range shift amount is an undefined value. In LLVM IR it is poison, which is more undefined. To avoid correctness problems, freeze the result of every shift just in case it is poison. LLVM middle end optimizations should be able to remove many of these freeze instructions, for example if it can prove that the shift amount is in-range.

This is prompted by some recent LLVM patches which optimize poison values more aggressively.

### DIFF
--- a/llpc/test/shaderdb/gfx9/ExtShaderInt8_TestShiftOp_lit.comp
+++ b/llpc/test/shaderdb/gfx9/ExtShaderInt8_TestShiftOp_lit.comp
@@ -23,9 +23,12 @@ void main()
 ; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIRV-to-LLVM translation results
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIR-V lowering results
-; SHADERTEST: %{{[0-9]*}} = ashr i8 %{{[0-9]*}}, 3
-; SHADERTEST: %{{[0-9]*}} = lshr i8 %{{[0-9]*}}, 3
-; SHADERTEST: %{{[0-9]*}} = shl i8 %{{[0-9]*}}, 3
+; SHADERTEST: %.fr = freeze i8 %{{[0-9]*}}
+; SHADERTEST: %{{[0-9]*}} = ashr i8 %.fr, 3
+; SHADERTEST: %.fr3 = freeze i8 %{{[0-9]*}}
+; SHADERTEST: %{{[0-9]*}} = lshr i8 %.fr3, 3
+; SHADERTEST: %.fr4 = freeze i8 %{{[0-9]*}}
+; SHADERTEST: %{{[0-9]*}} = shl i8 %.fr4, 3
 ; SHADERTEST-LABEL: {{^// LLPC}} pipeline patching results
 ; SHADERTEST: AMDLLPC SUCCESS
 */
--- a/llpc/translator/lib/SPIRV/SPIRVReader.cpp
+++ b/llpc/translator/lib/SPIRV/SPIRVReader.cpp
@@ -1272,8 +1272,11 @@ Value *SPIRVToLLVM::transShiftLogicalBitwiseInst(SPIRVValue *bv, BasicBlock *bb,
   if (shift->getType()->isIntOrIntVectorTy())
     shift = getBuilder()->CreateZExtOrTrunc(shift, base->getType());
 
-  auto inst = BinaryOperator::Create(bo, base, shift, bv->getName(), bb);
+  Instruction *inst = BinaryOperator::Create(bo, base, shift, bv->getName(), bb);
   setFastMathFlags(inst);
+
+  if (isShiftOpCode(op))
+    inst = new FreezeInst(inst, "shift.freeze", bb);
 
   return inst;
 }


jayfoad:
Without this patch, upstream LLVM commit cde681c865302f689f0937e5ec2e127a089a4db3 "[InstCombine] Replace phi operands in successors of unreachable block" causes a failure in Vulkan CTS test `dEQP-VK.graphicsfuzz.cov-fold-shift-gte32`.

jayfoad:
> but do we already know if this has negative impact on Codegen?

I have not analyzed that yet.

llpc/test/shaderdb/gfx9/ExtShaderInt8_TestShiftOp_lit.comp
```diff
@@ -23,9 +23,9 @@ void main()
 ; RUN: amdllpc -v %gfxip %s | FileCheck -check-prefix=SHADERTEST %s
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIRV-to-LLVM translation results
 ; SHADERTEST-LABEL: {{^// LLPC}} SPIR-V lowering results
-; SHADERTEST: %{{[0-9]*}} = ashr i8 %{{[0-9]*}}, 3
-; SHADERTEST: %{{[0-9]*}} = lshr i8 %{{[0-9]*}}, 3
-; SHADERTEST: %{{[0-9]*}} = shl i8 %{{[0-9]*}}, 3
+; SHADERTEST: %{{[0-9]*}} = ashr i8 %.fr{{[0-9]*}}, 3
```
tsymalla-AMD, line 26: Since it is not immediately obvious what the `.fr` operand is, could you include the freeze instructions in the test please?

tsymalla-AMD (changes requested):

llpc/test/shaderdb/gfx9/ExtShaderInt8_TestShiftOp_lit.comp
jayfoad: Done. This shows an interesting effect. LLPC inserts the `freeze` after the shift, but InstCombine moves it before the shift if it can see that the shift amount operand is in-range.

dstutt:
https://github.com/GPUOpen-Drivers/xgl/pull/172 is required before this will pass the actions (although maybe that doesn't block merge)

amdvlk-admin:
### Test summary for commit ac9a7fe0fec60c55b13f1f06579a7848036b1667
<details><summary>CTS tests (Failed: 0/211362)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35147/70459 (49.9%)</li>
<li>Failed: 0/70459 (0.0%)</li>
<li>Not Supported: 35312/70459 (50.1%)</li><li>Warnings: 0/70459 (0.0%)</li>
</ul></details>

<li>Passed: 39026/70459 (55.4%)</li>
<li>Failed: 0/70459 (0.0%)</li>
<li>Not Supported: 31433/70459 (44.6%)</li><li>Warnings: 0/70459 (0.0%)</li>
</ul></details>

<li>Passed: 39058/70444 (55.4%)</li>
<li>Failed: 0/70444 (0.0%)</li>
<li>Not Supported: 31386/70444 (44.6%)</li><li>Warnings: 0/70444 (0.0%)</li>
</ul></details>


</ul></details>

llpc/translator/lib/SPIRV/SPIRVReader.cpp
```diff
@@ -1272,9 +1272,12 @@ Value *SPIRVToLLVM::transShiftLogicalBitwiseInst(SPIRVValue *bv, BasicBlock *bb,
   if (shift->getType()->isIntOrIntVectorTy())
     shift = getBuilder()->CreateZExtOrTrunc(shift, base->getType());
 
-  auto inst = BinaryOperator::Create(bo, base, shift, bv->getName(), bb);
+  Instruction *inst = BinaryOperator::Create(bo, base, shift, bv->getName(), bb);
   setFastMathFlags(inst);
 
+  if (isShiftOpCode(op))
+    inst = new FreezeInst(inst, "shift.freeze", bb);
```
nhaehnle, line 1279: Some of the code above already uses the builder, can we just align with that here as well?

nhaehnle (approved):
This feels fairly likely to introduce code quality regressions because of how it adds a freeze on the result of the shift even when we really only care about one of the operands.

Perhaps a pragmatic mitigation would be to avoid the freeze if the shift amount is a constant that's in range.

jayfoad:
> This feels fairly likely to introduce code quality regressions because of how it adds a freeze on the result of the shift even when we really only care about one of the operands.
> 
> Perhaps a pragmatic mitigation would be to avoid the freeze if the shift amount is a constant that's in range.

Agreed on all counts. I will do some static analysis of the impact on codegen.

jayfoad:
> This feels fairly likely to introduce code quality regressions

Yes, this patch is fairly disruptive. The first problem I noticed is that InstCombine can no longer combine:
```
  %s = shl i32 1, %b
  %f = freeze i32 %s
  %m = mul i32 %a, %f
->
  %m = shl i32 %a, %b
```
Even if you add a `freeze` after the new `shl` the transformation is not valid, because `mul 0, undef` is `0` but `shl 0, undef` is `poison`.

>  because of how it adds a freeze on the result of the shift even when we really only care about one of the operands.

Right. In practice it turns out to be 20x less disruptive to add a `urem` before every shift:
```
  %s = shl i32 %a, %b
->
  %urem = urem i32 %b, 32
  %s = shl i32 %a, %urem
```
If the shift amount is a constant these can be folded away in IR, otherwise they should get folded away during instruction selection (but in practice that does not _always_ happen).

nhaehnle:
Instead of a `urem` you could use an `and` since certainly that's what InstCombine will canonicalize it to anyway?

But yeah, I think you're right about the `a * (1 << b)` pattern. Alive also agrees that combine can't be salvaged: https://alive2.llvm.org/ce/z/au3gRL.

Masking the shift amount seems like a reasonable compromise. And we should push for SPIR-V to eventually align with LLVM in this regard...

jayfoad:
> Instead of a `urem` you could use an `and` since certainly that's what InstCombine will canonicalize it to anyway?

Does SPIR-V support non-power-of-two sized integer types?

jayfoad:
Since the code is completely different I've opened a new PR for the masking approach: #2599

nhaehnle:
> Does SPIR-V support non-power-of-two sized integer types?

Syntactically, SPIR-V itself would support non-POT integer types, but the actual set of supported types is restricted. There are explicit extensions for 16-bit and 8-bit types, for example.

PR #2594, opened 2023-07-17
Cps stack lowering

ruiling:
Lower cps stack operations.

PR #2610, opened 2023-07-31
lgc: make irLink() more robust

ruiling:
Just use the datalayout from one of the modules, the modules to be linked should already get the same datalayout. This make the code more robust to possible datalayout modifications.

### DIFF
--- a/lgc/state/Compiler.cpp
+++ b/lgc/state/Compiler.cpp
@@ -141,7 +141,7 @@ Module *PipelineState::irLink(ArrayRef<Module *> modules, PipelineLink pipelineL
     pipelineModule = new Module("lgcPipeline", getContext());
     TargetMachine *targetMachine = getLgcContext()->getTargetMachine();
     pipelineModule->setTargetTriple(targetMachine->getTargetTriple().getTriple());
-    pipelineModule->setDataLayout(targetMachine->createDataLayout());
+    pipelineModule->setDataLayout(modules.front()->getDataLayout());
 
     Linker linker(*pipelineModule);
     for (Module *module : modules) {


amdvlk-admin:
### Test summary for commit b01d71321d2da8dfc572e0ffc310d12415b2de39
<details><summary>CTS tests (Failed: 0/211876)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35146/70631 (49.8%)</li>
<li>Failed: 0/70631 (0.0%)</li>
<li>Not Supported: 35485/70631 (50.2%)</li><li>Warnings: 0/70631 (0.0%)</li>
</ul></details>

<li>Passed: 39023/70629 (55.3%)</li>
<li>Failed: 0/70629 (0.0%)</li>
<li>Not Supported: 31606/70629 (44.7%)</li><li>Warnings: 0/70629 (0.0%)</li>
</ul></details>

<li>Passed: 39057/70616 (55.3%)</li>
<li>Failed: 0/70616 (0.0%)</li>
<li>Not Supported: 31559/70616 (44.7%)</li><li>Warnings: 0/70616 (0.0%)</li>
</ul></details>


</ul></details>

jasilvanus:
This is only to work around differences in address spaces that are unknown in some parts of the compiler, right?

Maybe we can add a utility function that checks that the two DLs agree up to some address spaces that are only contained in one of them (e.g. by first checking that pointers are consistent, and then after overriding pointers, that the resulting DLs are the same), and add an assertion using that?

jayfoad:
> Maybe we can add a utility function that checks that the two DLs agree up to some address spaces that are only contained in one of them (e.g. by first checking that pointers are consistent, and then after overriding pointers, that the resulting DLs are the same), and add an assertion using that?

Can you just assert that all modules in `modules` have identical datalayout?

jasilvanus:
That is what the linker does, and my understanding with #2954 is that in some cases the DL is not identical due to address space differences?

jayfoad (approved):
This fixes the problem for me, so I am in favour.

ruiling:
> This is only to work around differences in address spaces that are unknown in some parts of the compiler, right?
I think this is the right way to implement irLink() function. The caller should already make sure the data layout are compatible with each other between the modules. We just use the layout from the input module.

> That is what the linker does, and my understanding with #2954 is that in some cases the DL is not identical due to address space differences?
It's not, the issue in #2594 is we were creating a new empty module in irLink() and set a different layout from the modules being linked.

jasilvanus (approved):
I misunderstood this, this is now fine for me.

The linker will still complain if the passed modules have inconsistent DLs.

linqun (approved):

PR #2638, opened 2023-08-15
Fix an issue of setting function attribute 'amdgpu-no-workgroup-id'

amdrexu:
We try to find the function that might use the attribute 'amdgpu-no-workgroup-id'. This is true when such entry-point function exists. However, in the case that shader cache is enabled, such functions might be absent in current module because we deliberately remove the duplicated shader entry-points by reusing cached ELFs. Therefore, we must check such case when trying to find the entry-point function.

The failure is found in the CTS group: VK.binding_model.descriptorset_random. sets*.*.noubo.nosbo.nosampledimg.outimgonly.noiub.*uab.task.noia.0

### DIFF
--- a/lgc/patch/Gfx9ConfigBuilder.cpp
+++ b/lgc/patch/Gfx9ConfigBuilder.cpp
@@ -1975,9 +1975,16 @@ void ConfigBuilder::buildCsRegConfig(ShaderStage shaderStage, CsRegConfig *confi
       break;
     }
   }
-  SET_REG_FIELD(config, COMPUTE_PGM_RSRC2, TGID_X_EN, !attribFunc->hasFnAttribute("amdgpu-no-workgroup-id-x"));
-  SET_REG_FIELD(config, COMPUTE_PGM_RSRC2, TGID_Y_EN, !attribFunc->hasFnAttribute("amdgpu-no-workgroup-id-y"));
-  SET_REG_FIELD(config, COMPUTE_PGM_RSRC2, TGID_Z_EN, !attribFunc->hasFnAttribute("amdgpu-no-workgroup-id-z"));
+  // NOTE: We enable TGID if we don't find any function using the attributes 'amdgpu-no-workgroup-id'. If shader
+  // cache is enabled, the cached ELFs will be reused. In such case, we might not find any entry-point functions.
+  // Since the cached ELFs will be reused, it is still safe to enable TGID because the register metadata will not
+  // be actually used.
+  const bool tgidXEn = !attribFunc || !attribFunc->hasFnAttribute("amdgpu-no-workgroup-id-x");
+  const bool tgidYEn = !attribFunc || !attribFunc->hasFnAttribute("amdgpu-no-workgroup-id-y");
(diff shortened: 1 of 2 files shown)


amdvlk-admin:
### Test summary for commit 2f503ea131f08662ed5e782456890f390cbd3b00
<details><summary>CTS tests (Failed: 0/211633)</summary><ul><li>Built with version 1.3.3.1</li>

<li>Passed: 35496/70549 (50.3%)</li>
<li>Failed: 0/70549 (0.0%)</li>
<li>Not Supported: 35053/70549 (49.7%)</li><li>Warnings: 0/70549 (0.0%)</li>
</ul></details>

<li>Passed: 39381/70550 (55.8%)</li>
<li>Failed: 0/70550 (0.0%)</li>
<li>Not Supported: 31169/70550 (44.2%)</li><li>Warnings: 0/70550 (0.0%)</li>
</ul></details>

<li>Passed: 39413/70534 (55.9%)</li>
<li>Failed: 0/70534 (0.0%)</li>
<li>Not Supported: 31121/70534 (44.1%)</li><li>Warnings: 0/70534 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (commented):
I believe this fix may be incomplete or incorrect. Consider the following scenario:

* The compiled shader uses workgroup ID X and also the "tg_size" register (for wave ID) and stored in the cache
* It is found in the cache, and we try to re-establish metadata here
* The function doesn't exist, so we assume that it uses all of X/Y/Z, but this puts the tg_size register in the wrong location

Can't we somehow re-use the PAL metadata from the ELF in the cache?

amdrexu:
Thanks. Yes, this is for task shader. 

For task shader, we always need X/Y/Z to calculate the flat workgroup ID so maybe the check of existence of `amdgpu-no-workgroup-id` is meaningless here. Do you think we can simplify it?

If not, my original plan is to add the existence of `amdgpu-no-workgroup-id` to `pipelineState` so it can be retrieved here to correctly set the `TGID_EN`. But Xue told me you objected to this in her previous review. I know it is clean we only keep metadata between passes by avoiding the introduction of member data in a global class. Do you still stick to this criteria?

Getting `TGID_EN` from cached ELF is feasible but current we don't have such utilities in LGC. The ELF merging and modification are mostly done in LLPC front-end `llpcElfWriter`. And I am thinking of if we can skip setting PAL ABI metadata when we find this shader stage is reused from cached ELF. This is because the whole stage will not be compiled by LLVM backend and all GPU codes are reused.

Which option you believe is reasonable and practical?

nhaehnle:
Can we simply skip the code that sets in `buildCsRegisters`

nhaehnle:
Can we simply skip the code that sets COMPUTE_PGM_RSRC2 fields in the case where the task shader is taken from the cache?

Other fields like VGPR_SIZE must be taken from the cached ELF already, right? So if we just don't set that register at all here, then a later step that pulls VGPR_SIZE and others from the cached task shader ELF should be able to do the same for COMPUTE_PGM_RSRC2.

PR #2670, opened 2023-09-04
llpc: Add Intrinsics implementation for FloatOpWithRoundMode

KOTOKORURU:
add the missed floatOp with roundMode support for RT

HLSL:

static float FloatOpWithRoundMode(uint roundMode, uint operation, float src0, float src1)
{
    return AmdExtD3DShaderIntrinsics_FloatOpWithRoundMode(roundMode, operation, src0, src1);
}

SPV:
               OpStore %479 %uint_1
               OpStore %480 %uint_1
               OpStore %481 %4548
               OpStore %482 %4551
       %4589 = OpFunctionCall %v3float %AmdExtD3DShaderIntrinsics_FloatOpWithRoundMode %479 %480 %481 %482

### DIFF
--- a/llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
+++ b/llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
@@ -88,6 +88,8 @@ SpirvProcessGpuRtLibrary::LibraryFunctionTable::LibraryFunctionTable() {
 #else
   m_libFuncPtrs["AmdExtD3DShaderIntrinsics_IntersectInternal"] = &SpirvProcessGpuRtLibrary::createIntersectBvh;
 #endif
+  m_libFuncPtrs["AmdExtD3DShaderIntrinsics_FloatOpWithRoundMode"] =
+      &SpirvProcessGpuRtLibrary::createFloatOpWithRoundMode;
   m_libFuncPtrs["AmdTraceRaySampleGpuTimer"] = &SpirvProcessGpuRtLibrary::createSampleGpuTimer;
   m_libFuncPtrs["AmdTraceRayGetFlattenedGroupThreadId"] = &SpirvProcessGpuRtLibrary::createGetFlattenedGroupThreadId;
   m_libFuncPtrs["AmdTraceRayGetHitAttributes"] = &SpirvProcessGpuRtLibrary::createGetHitAttributes;
@@ -104,8 +106,6 @@ SpirvProcessGpuRtLibrary::LibraryFunctionTable::LibraryFunctionTable() {
   m_libFuncPtrs["AmdTraceRaySetParentId"] = &SpirvProcessGpuRtLibrary::createSetParentId;
   m_libFuncPtrs["AmdTraceRayDispatchRaysIndex"] = &SpirvProcessGpuRtLibrary::createDispatchRayIndex;
   m_libFuncPtrs["AmdTraceRayGetStaticId"] = &SpirvProcessGpuRtLibrary::createGetStaticId;
-  m_libFuncPtrs["AmdExtD3DShaderIntrinsics_FloatOpWithRoundMode"] =
-      &SpirvProcessGpuRtLibrary::createFloatOpWithRoundMode;
 }
 
 // =====================================================================================================================
@@ -379,17 +379,6 @@ void SpirvProcessGpuRtLibrary::createFloatOpWithRoundMode(Function *func) {
   //     }
   // }
   enum OperationType : uint32_t { Add = 0, Sub, Mul };
-
-  // Use setReg to set SQ_WAVE_MODE.
-  // hwRegId : SQ related register index.
-  // Offset : register field offset.
-  // Width  : field width.
-  // hwReg : (hwRegId | (Offset << 6) | ((Width - 1) << 11)
-  constexpr uint32_t sqHwRegMode = 1;
-  constexpr uint32_t witdth = 2;
-  constexpr uint32_t offset = 0;
-  uint32_t hwRegField = ((sqHwRegMode) | (offset << 6) | ((witdth - 1) << 11));
-
   auto int32Ty = m_builder->getInt32Ty();
   auto retType = cast<FixedVectorType>(func->getReturnType());
   auto argIt = func->arg_begin();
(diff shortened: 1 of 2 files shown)


amdvlk-admin:

llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
```diff
@@ -104,6 +104,8 @@ SpirvProcessGpuRtLibrary::LibraryFunctionTable::LibraryFunctionTable() {
   m_libFuncPtrs["AmdTraceRaySetParentId"] = &SpirvProcessGpuRtLibrary::createSetParentId;
   m_libFuncPtrs["AmdTraceRayDispatchRaysIndex"] = &SpirvProcessGpuRtLibrary::createDispatchRayIndex;
   m_libFuncPtrs["AmdTraceRayGetStaticId"] = &SpirvProcessGpuRtLibrary::createGetStaticId;
+  m_libFuncPtrs["AmdExtD3DShaderIntrinsics_FloatOpWithRoundMode"] =
+      &SpirvProcessGpuRtLibrary::createFloatOpWithRoundMode;
```
nhaehnle, line 108: Please move this up so it's listed next to the other AmdExtD3D functions.

llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
```diff
@@ -349,6 +351,107 @@ void SpirvProcessGpuRtLibrary::createConvertF32toF16WithRoundingMode(Function *f
...
+      // Restore the roundMode.
+      m_builder->CreateNamedCall("llvm.amdgcn.s.setreg", Type::getVoidTy(m_builder->getContext()),
+                                 {m_builder->getInt32(hwRegField), tmpRoundMode}, {});
+      m_builder->CreateRet(result);
+    } else {
+      result = m_builder->CreateFMul(src0, src1);
+      // Restore the roundMode.
+      m_builder->CreateNamedCall("llvm.amdgcn.s.setreg", Type::getVoidTy(m_builder->getContext()),
+                                 {m_builder->getInt32(hwRegField), tmpRoundMode}, {});
+      m_builder->CreateRet(result);
```
nhaehnle, line 448: The setreg + return suffix should be moved out of the if/else chain since it's logically redundant.

llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
```diff
@@ -349,6 +351,107 @@ void SpirvProcessGpuRtLibrary::createConvertF32toF16WithRoundingMode(Function *f
...
+    }
+  }
+  m_builder->SetInsertPoint(opDefault);
+  m_builder->CreateRet(PoisonValue::get(retType));
```
nhaehnle, line 452: Use an `unreachable` instead.

nhaehnle:
I don't think that this approach using getreg/setreg can work correctly. The issue is that `fadd` and `fmul` are defined as pure instructions that don't have an interaction with their environment. Intermediate passes may well move these instructions around in ways that don't consider those getreg/setreg calls.

Please try to use the constrained FP intrinsics, which allow explicitly setting a rounding mode: https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics

KOTOKORURU:
> I don't think that this approach using getreg/setreg can work correctly. The issue is that `fadd` and `fmul` are defined as pure instructions that don't have an interaction with their environment. Intermediate passes may well move these instructions around in ways that don't consider those getreg/setreg calls.
> 
> Please try to use the constrained FP intrinsics, which allow explicitly setting a rounding mode: https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics


Thanks Nicolai, that's really helpful information.
And  i guess  there would be 3*4 intrinsic calls, can you take look into my first commit,
if we can load the op and roundmode directly from all the calls? because HLSL now only use static param with this function in our driver(like gpurt..etc), so in SPV they are all constant. but not sure they will change the usage in future, such that it will decide the op or roundmode in runtime.

anyway i still send the implemention of the switch-case with your advice.

amdvlk-admin:
### Test summary for commit 4ace5b8b0a993c8f617e7f948217e1b201d6f0f2
<details open><summary>Rhel 8.9, Gfx10</summary><ul>

<li>Passed: 34736/69788 (49.8%)</li>
<li>Not Supported: 35052/69788 (50.2%)</li><li>Warnings: 0/69788 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
```diff
@@ -349,6 +351,85 @@ void SpirvProcessGpuRtLibrary::createConvertF32toF16WithRoundingMode(Function *f
...
+  enum OperationType : uint32_t { Add = 0, Sub, Mul };
+  auto int32Ty = m_builder->getInt32Ty();
+  auto retType = cast<FixedVectorType>(func->getReturnType());
+  auto argIt = func->arg_begin();
+  Value *roundMode = m_builder->CreateLoad(int32Ty, argIt++);
+  Value *operation = m_builder->CreateLoad(int32Ty, argIt++);
+  Value *src0 = m_builder->CreateLoad(retType, argIt++);
+  Value *src1 = m_builder->CreateLoad(retType, argIt);
+
+  Value *fpExcpt = MetadataAsValue::get(*m_context, MDString::get(*m_context, "fpexcept.strict"));
```
nhaehnle, line 390: Do we really have shader code that reads the floating point exception flags?

nhaehnle (approved):
Thanks, this basically LGTM, though I do wonder about something (see inline).

llpc/lower/llpcSpirvProcessGpuRtLibrary.cpp
KOTOKORURU: I think there is no excption flags for shader

nhaehnle: The shader HW does have exception flags. But if we don't read them, this should be "fpexcept.ignore" instead of "fpexcept.strict" to allow for a little more freedom in codegen.

KOTOKORURU: another question is constrained.fp seems fail to generate SP3 to set roundmode.

amdvlk-admin:
### Test summary for commit adcd81327ab60561d45f1015ea3a5ca6d23e444f
<details><summary>CTS tests (Failed: 0/208496)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 36978/69503 (53.2%)</li>
<li>Not Supported: 32525/69503 (46.8%)</li><li>Warnings: 0/69503 (0.0%)</li>
</ul></details>

</ul></details>

<li>Passed: 41178/69490 (59.3%)</li>
<li>Failed: 0/69490 (0.0%)</li>
<li>Not Supported: 28312/69490 (40.7%)</li><li>Warnings: 0/69490 (0.0%)</li>
</ul></details>


</ul></details>

PR #2673, opened 2023-09-04
Add input/output semantic metadata in partial pipeline

xuechen417:
Currently, we fail to link relocatable shaders if FS has builtIn inputs. 
To support this kind of case, we add two ArrayDocNodes in ELF: 

`.ps_input_semantic` has one member - `.semantic`
`.preraster_output_semantic` has two members - `.semantic` and `.index`. 

Where `.semantic` encodes builtIn Id and location to perform matching operation and '.index` is used to update the SPI_PS_INPUT_CNTL_*.OFFSET.

### DIFF
--- a/lgc/patch/RegisterMetadataBuilder.cpp
+++ b/lgc/patch/RegisterMetadataBuilder.cpp
@@ -58,6 +58,7 @@ void RegisterMetadataBuilder::buildPalMetadata() {
       m_isNggMode = m_pipelineState->getNggControl()->enableNgg;
 
     Util::Abi::PipelineType pipelineType = Util::Abi::PipelineType::VsPs;
+    auto lastVertexProcessingStage = m_pipelineState->getLastVertexProcessingStage();
 
     DenseMap<unsigned, unsigned> apiHwShaderMap;
     if (m_hasTask || m_hasMesh) {
@@ -81,7 +82,7 @@ void RegisterMetadataBuilder::buildPalMetadata() {
         if (m_hasVs)
           apiHwShaderMap[ShaderStageVertex] = Util::Abi::HwShaderHs;
       }
-      auto lastVertexProcessingStage = m_pipelineState->getLastVertexProcessingStage();
+
       if (lastVertexProcessingStage != ShaderStageInvalid) {
         if (lastVertexProcessingStage == ShaderStageCopyShader)
           lastVertexProcessingStage = ShaderStageGeometry;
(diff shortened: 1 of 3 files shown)


amdvlk-admin:
### Test summary for commit 1b753740066daa2fda393b0f52fa1d655ecb26e0
<details open><summary>Rhel 8.9, Gfx10</summary><ul>

<li>Passed: 34736/69788 (49.8%)</li>
<li>Not Supported: 35052/69788 (50.2%)</li><li>Warnings: 0/69788 (0.0%)</li>
</ul></details>

</ul></details>

</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit 5b3912bd21c73ab2f8b5490c912ec53bdd034cb0

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/RegisterMetadataBuilder.cpp
```diff
@@ -172,6 +173,36 @@ void RegisterMetadataBuilder::buildPalMetadata() {
...
       buildPaSpecificRegisters();
 
+    if (lastVertexProcessingStage != ShaderStageInvalid) {
+      // Fill ".preraster_output_semantic"
+      auto resUsage = m_pipelineState->getShaderResourceUsage(lastVertexProcessingStage);
+      auto &outputLocInfoMap = resUsage->inOutUsage.outputLocInfoMap;
+      msgpack::ArrayDocNode *preRasterOutputSemanticNode =
+          outputLocInfoMap.empty()
+              ? nullptr
+              : &getPipelineNode()[Util::Abi::PipelineMetadataKey::PrerasterOutputSemantic].getArray(true);
```
nhaehnle, line 183: In this case, I think it's worth to just add another `if` to guard all of this by `!outputLocInfoMap.empty() || !builtInOutputLocMap.empty()`. Then the slightly strange ternary operator here can be avoided.

lgc/patch/RegisterMetadataBuilder.cpp
```diff
@@ -931,6 +962,27 @@ void RegisterMetadataBuilder::buildPsRegisters() {
   cbShaderMaskNode[Util::Abi::CbShaderMaskMetadataKey::Output5Enable] = (cbShaderMask >> 20) & 0xF;
   cbShaderMaskNode[Util::Abi::CbShaderMaskMetadataKey::Output6Enable] = (cbShaderMask >> 24) & 0xF;
   cbShaderMaskNode[Util::Abi::CbShaderMaskMetadataKey::Output7Enable] = (cbShaderMask >> 28) & 0xF;
+
+  // Fill .ps_input_semantic for partial pipeline
+  auto &inputLocInfoMap = resUsage->inOutUsage.inputLocInfoMap;
+  msgpack::ArrayDocNode *psInputSemanticNode =
+      inputLocInfoMap.empty() ? nullptr
+                              : &getPipelineNode()[Util::Abi::PipelineMetadataKey::PsInputSemantic].getArray(true);
```
nhaehnle, line 970: Same thing here: guard with `if (!inputLocInfoMap.empty() || !builtInInputLocMap.empty())`

lgc/state/PalMetadata.cpp
```diff
@@ -1263,6 +1263,45 @@ bool PalMetadata::fragmentShaderUsesMappedBuiltInInputs() {
...
+    for (auto outputSemantic : outputSemanticArray) {
+      const unsigned outSemantic =
+          outputSemantic.getMap(true)[Util::Abi::PrerasterOutputSemanticMetadataKey::Semantic].getUInt();
+      if (inSemantic == outSemantic) {
+        const unsigned mappedLoc =
+            outputSemantic.getMap(true)[Util::Abi::PrerasterOutputSemanticMetadataKey::Index].getUInt();
+        spiPsInputCntElem[Util::Abi::SpiPsInputCntlMetadataKey::Offset] = mappedLoc;
+        ++matchedElemCount;
+      }
+    }
```
nhaehnle, line 1298: Presumably this loop should:

* Early-exit when the pre-raster output has been found
* Have an assertion that is triggered when no pre-raster output is found

lgc/state/PalMetadata.cpp
```diff
@@ -1263,6 +1263,45 @@ bool PalMetadata::fragmentShaderUsesMappedBuiltInInputs() {
...
+    ++elemIdx;
+  }
+
+  return (elemIdx == matchedElemCount);
```
nhaehnle, line 1302: Drop the parentheses in LLVM style.

nhaehnle (commented):
Why is one an array and the other a map? That seems inconsistent, it would be better to make both a map or both an array.

Can we perhaps also avoid the numeric encoding of the "semantic ID" in favor of just using a dictionary with `builtin` and `location` key/value pairs?

xuechen417:
> Why is one an array and the other a map? That seems inconsistent, it would be better to make both a map or both an array.
> 
It follows the PAL's definitions of .ps_input_semantic (an array) and .preraster_output_semantic (a map).

lgc/patch/RegisterMetadataBuilder.cpp
xuechen417: Fixed.

lgc/patch/RegisterMetadataBuilder.cpp
xuechen417: Fixed.

lgc/state/PalMetadata.cpp
xuechen417: This function is removed as the match for semantic is handled by PAL. In the follow-up patch, we just need to loose the restriction, that is the check for `fragmentShaderUsesMappedBuiltInInputs()` is only required for relocatable shaders.

lgc/state/PalMetadata.cpp
xuechen417: Code is remvoed.

xuechen417 (commented):
V1: Remove the match relavant code. Refine the code of adding metadata based on the review.

xuechen417:
> Can we perhaps also avoid the numeric encoding of the "semantic ID" in favor of just using a dictionary with `builtin` and `location` key/value pairs?

I think a dictionary is not easy to share between two partial pipelines. I changed to assign builtIn Id and `location + MaxBuiltIn` to `.semantic`, respectively.

nhaehnle (approved):
Okay, thanks for the added explanations. LGTM.

PR #2677, opened 2023-09-04
Add Continufy pass

ruiling:


lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,194 @@
...
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ *
+ **********************************************************************************************************************/
+/**
+ ***********************************************************************************************************************
+ * @file  Continufy.cpp
+ * @brief LLPC source file: contains implementation of class lgc::Continufy.
+ * This pass translate indirect call into cps.await call, which will be lowered into continuation call.
```
Flakebi, line 29: typo: translate**s**

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,194 @@
...
+  LLVMContext &context = module.getContext();
+
+  llvm_dialects::Builder builder(context);
+  SmallVector<Instruction *> tobeErased;
+  SmallVector<Function *> functions;
+
+  for (auto &fn : make_early_inc_range(module.functions())) {
+    MDNode *continufyStage = fn.getMetadata("continufy.stage");
+    Function *fnPtr = &fn;
+    int32_t currentRtStage;
```
jasilvanus, line 134: Maybe make this an `optional` to catch unintialized cases? Or initialize with an obviously invalid value.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,194 @@
...
+  rt::RayTracingShaderStage rtStage = static_cast<rt::RayTracingShaderStage>(stage);
+  switch(rtStage){
+    case rt::RayTracingShaderStage::ShaderStageRayGeneration:
+      llvm_unreachable("Raygen shader should arrive here.\n");
+    case rt::RayTracingShaderStage::ShaderStageClosestHit:
+    case rt::RayTracingShaderStage::ShaderStageMiss:
+      // Traversal | RGS
+      return (1 << 2 | 1);
+    case rt::RayTracingShaderStage::ShaderStageCallable:
+      // CHS/Miss/Callable | Traversal | RGS
```
Flakebi: Traversal does not call a callable shader, so I think it is only CHS/Miss/Callable | RGS

lgc/include/lgc/util/Internal.h
```diff
@@ -96,4 +96,8 @@ bool isDontCareValue(llvm::Value *value);
 // type in a return value struct, ensuring it gets into VGPRs.
 llvm::Type *getVgprTy(llvm::Type *ty);
 
+// Modify the function argument types, and return the new function.
+llvm::Function *mutateFunctionArguments(llvm::Function &fn, llvm::Type *retTy,
+                                        const llvm::ArrayRef<llvm::Type *> argTys, llvm::AttributeList attributes);
+
```
jasilvanus, line 102: Maybe integrate the argument fixup into this utility? We could pass an additional list of indices, mapping between old and new arguments.

Otherwise, we should clearly mention here that the user needs to fix up arguments.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+/// Return the levels of the caller ray-tracing stages, which will be executed after the function return.
+static unsigned getReturnedLevels(int32_t stage) {
+  // Traversal will return to RGS.
+  if (stage == -1)
+    return 1;
+  rt::RayTracingShaderStage rtStage = static_cast<rt::RayTracingShaderStage>(stage);
+  switch (rtStage) {
+  case rt::RayTracingShaderStage::ShaderStageRayGeneration:
+    llvm_unreachable("Raygen shader should arrive here.\n");
```
jasilvanus, line 78: "should not"?

I think we can remove the the trailing newline in the message (same for the other `llvm_unreachable`s.).

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+  case rt::RayTracingShaderStage::ShaderStageCallable:
+    return 1;
+  case rt::RayTracingShaderStage::ShaderStageAnyHit:
+    return 3;
+  case rt::RayTracingShaderStage::ShaderStageIntersection:
+    return 4;
+  default:
+    llvm_unreachable("Unknown raytracing shader type.\n");
+  }
+}
```
jasilvanus, line 114: This is not specific to Continufy. Maybe add to `LgcRtDialect.h`?

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+  case rt::RayTracingShaderStage::ShaderStageClosestHit:
+  case rt::RayTracingShaderStage::ShaderStageMiss:
+    // Traversal | RGS
+    return (1 << 2 | 1);
+  case rt::RayTracingShaderStage::ShaderStageCallable:
+    // CHS/Miss/Callable | Traversal | RGS
+    return (1 << 2 | 1 << 1 | 1);
+  case rt::RayTracingShaderStage::ShaderStageAnyHit:
+    // IS | Traversal | RGS
+    return (1 << 4 | 1 << 2 | 1);
```
jasilvanus, line 88: We should avoid these redundant constants.
We should instead implement this in terms of `getLevelOfRtStage`.
With `using Stage = rt::RayTracingShaderStage`, the names are a bit shorter.

Also, maybe we can improve the `rt::RayTracingShaderStage` enum to avoid the redundant `ShaderStage` prefix? It's already an enum class, so there is no danger of name conflicts.

Flakebi (commented):
Code looks mostly good to me (see some inline comments).

Who inserts the `continuify.stage` metadata?
As I understood it, the intention is to run this pass on normal/unmodified IR and this pass fetches any information from the existing IR or sideband data. Can’t we extract the needed information from the existing shaderstage metadata?

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+  default:
+    llvm_unreachable("Unknown raytracing shader type.\n");
+  }
+}
+
+/// Return level of the ray-tracing stage.
+static unsigned getLevelOfRtStage(int32_t stage) {
```
jasilvanus, line 95: Nit: Maybe return `uint32_t` instead of `unsigned`, given you also specify a bit width on the argument?

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+// @param [in/out] analysisManager : Analysis manager to use for this transformation
+// @returns : The preserved analyses (The analyses that are still valid after this pass)
+PreservedAnalyses Continufy::run(Module &module, ModuleAnalysisManager &analysisManager) {
+  LLVM_DEBUG(dbgs() << "Run the Continufy pass \n");
+  m_pipelineState = analysisManager.getResult<PipelineStateWrapper>(module).getPipelineState();
+  LLVMContext &context = module.getContext();
+
+  llvm_dialects::Builder builder(context);
+  SmallVector<Instruction *> tobeErased;
+  SmallVector<Function *> functions;
```
jasilvanus, line 129: This is unused.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+// =====================================================================================================================
+// Executes this LLVM patching pass on the specified LLVM module.
+//
+// @param [in/out] module : LLVM module to be run on
+// @param [in/out] analysisManager : Analysis manager to use for this transformation
+// @returns : The preserved analyses (The analyses that are still valid after this pass)
+PreservedAnalyses Continufy::run(Module &module, ModuleAnalysisManager &analysisManager) {
+  LLVM_DEBUG(dbgs() << "Run the Continufy pass \n");
+  m_pipelineState = analysisManager.getResult<PipelineStateWrapper>(module).getPipelineState();
```
jasilvanus, line 124: Maybe reset `m_pipelineState` at the end of `run`? Otherwise we might have a dangling pointer later on. 

Or even better: Move the implementation of `Continufy` into a separate class `ContinufyPassImpl`, and replace this function by

```c++
ContinufyPassHelper(analysisManager.getResult<PipelineStateWrapper>(module).getPipelineState()).run();
```

Then there is no state that can leak after `run`.

amdvlk-admin:
### Test summary for commit 3c527806066fbac7e45c228d4d4403cbddd0ea26

</ul></details>

</ul></details>

</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit e290b3bcb48ae0e4ec6d03c76ce17c1e20cc4c1d

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+  for (auto &fn : make_early_inc_range(module.functions())) {
+    MDNode *continufyStage = fn.getMetadata("continufy.stage");
+    Function *fnPtr = &fn;
+    int32_t currentRtStage;
+    if (continufyStage) {
+      fnPtr = insertCpsArguments(fn);
+      currentRtStage = mdconst::extract<ConstantInt>(continufyStage->getOperand(0))->getSExtValue();
+      unsigned level = getLevelOfRtStage(currentRtStage);
+      fnPtr->setMetadata("lgc.cps", MDNode::get(context, ConstantAsMetadata::get(builder.getInt32(level))));
```
tsymalla, line 139: Isn’t there a cps helper for that?

nhaehnle:
> Who inserts the `continuify.stage` metadata?
> As I understood it, the intention is to run this pass on normal/unmodified IR and this pass fetches any information from the existing IR or sideband data. Can’t we extract the needed information from the existing shaderstage metadata?

SpirvLowerRaytracing should insert it. Unfortunately, the normal IR flow only has `!lgc.shaderstage`, which just indicates "compute shader" for all raytracing shaders. Hence the idea to preserve the "real" shader stage in one-off metadata.

lgc/patch/Continufy.cpp
nhaehnle: Please also note that there is `lgc::cps::CpsLevel`, which should be used here.

lgc/patch/Continufy.cpp
nhaehnle: There's `lgc::cps::CpsLevel`, and lgcrt currently does not depend on lgccps... I suppose it would make sense in shared/continuations though.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,196 @@
...
+    // Translate call instruction with %continufy.stage into lgc.cps.await() with continuation reference.
+    for (auto &block : *fnPtr) {
+      for (auto &inst : block) {
+        if (!isa<CallInst>(inst))
+          continue;
+        auto *calleeStage = inst.getMetadata("continufy.stage");
+        if (!calleeStage)
+          continue;
+
+        auto &call = cast<CallInst>(inst);
```
nhaehnle, line 151: assert that call.getCallingConv is SPIR_FUNC.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,194 @@
...
+  // Traversal will return to RGS.
+  if (stage == -1) return 1;
+  rt::RayTracingShaderStage rtStage = static_cast<rt::RayTracingShaderStage>(stage);
+  switch(rtStage){
+    case rt::RayTracingShaderStage::ShaderStageRayGeneration:
+      llvm_unreachable("Raygen shader should arrive here.\n");
+    case rt::RayTracingShaderStage::ShaderStageClosestHit:
+    case rt::RayTracingShaderStage::ShaderStageMiss:
+      // Traversal | RGS
+      return (1 << 2 | 1);
```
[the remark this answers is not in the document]
nhaehnle: The code is correct but the comment is wrong, because CHS or miss can return to CHS or miss.

lgc/patch/Continufy.cpp
ruiling: I have re-implemented the code using cps::CpsLevel, please double check.

lgc/patch/Continufy.cpp
ruiling: The `shared/continuation` is in the repo now, so I just leave it here for now.

lgc/patch/Continufy.cpp
ruiling: `optional` sounds a good idea here. fixed.

lgc/patch/Continufy.cpp
ruiling: I did not realize it. Thanks for letting me know.

lgc/patch/Continufy.cpp
ruiling: I think the pointer member will not be accessed once the `run` end. So I don't think this is a serious issue. Anyway, I have reset it to `nullptr` at the end.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+using namespace llvm;
+using namespace lgc;
+using namespace lgc::cps;
+
+namespace lgc {
+using RtStage = rt::RayTracingShaderStage;
+
+static Function *insertCpsArguments(Function &fn) {
+  // convert rtstage to levels
```
tsymalla, line 51: This is not happening here, right now? Maybe remove that comment

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+    Value *newArg = newFn->getArg(idx + 2);
+    newArg->setName(oldArg->getName());
+    oldArg->replaceAllUsesWith(newArg);
+  }
+  newFn->getArg(0)->setName("state");
+  newFn->getArg(1)->setName("rcr");
+  return newFn;
+}
+
+/// Return the levels of the caller ray-tracing stages, which will be executed after the function return.
```
tsymalla, line 72:
```suggestion
/// For a given ray-tracing stage, return the CPS level.
```

lgc/patch/Continufy.cpp
ruiling: I have misunderstood the meaning of `levels` here. I have updated the function completely. `Continufy` is meant to be used with SpirvLowerRayTracing. And translate old indirect-call based ray-tracing into cps-style call.

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+  SmallVector<Instruction *> tobeErased;
+
+  for (auto &fn : make_early_inc_range(module.functions())) {
+    MDNode *continufyStage = fn.getMetadata("continufy.stage");
+    Function *fnPtr = &fn;
+    std::optional<int32_t> currentRtStage;
+    if (continufyStage) {
+      fnPtr = insertCpsArguments(fn);
+      currentRtStage = mdconst::extract<ConstantInt>(continufyStage->getOperand(0))->getSExtValue();
+      CpsLevel level = getLevelOfRtStage(currentRtStage.value());
```
tsymalla, line 144: Pass the `MDNode` directly into the `getLevelOfRtStage` function?

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+        builder.SetInsertPoint(&call);
+        auto *continuationRef = builder.CreatePtrToInt(called, IntegerType::get(context, 32));
+        CpsLevel calleeLevel =
+            getLevelOfRtStage(mdconst::extract<ConstantInt>(calleeStage->getOperand(0))->getSExtValue());
+        // RayGen level is zero, so it does not need a logic OR here.
+        if (calleeLevel != CpsLevel::RayGen)
+          continuationRef = builder.CreateOr(continuationRef, builder.getInt32((uint32_t)calleeLevel));
+
+        SmallVector<Value *> callArgs(call.args());
+        auto *newCall = builder.create<AwaitOp>(call.getType(), continuationRef, 1u << (unsigned)calleeLevel, callArgs);
```
tsymalla, line 170: Why not let `getLevelOfRtStage` compute the shifted level?

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+  case RtStage::ShaderStageIntersection:
+    // Traversal
+    return 1u << (unsigned)CpsLevel::Traversal;
+  default:
+    llvm_unreachable("Unknown raytracing shader type.");
+  }
+}
+
+/// Return level of the ray-tracing stage.
+static CpsLevel getLevelOfRtStage(int stage) {
```
tsymalla, line 101:
```suggestion
static CpsLevel getCpsLevelFromRtStage(int stage) {
```

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return (1u << (unsigned)CpsLevel::Intersection | 1u << (unsigned)CpsLevel::Traversal);
+  case RtStage::ShaderStageIntersection:
+    // Traversal
+    return 1u << (unsigned)CpsLevel::Traversal;
+  default:
+    llvm_unreachable("Unknown raytracing shader type.");
+  }
+}
+
+/// Return level of the ray-tracing stage.
```
tsymalla, line 100:
```suggestion
/// Return CPS level of the ray-tracing stage.
```

tsymalla (changes requested):

lgc/include/lgc/util/Internal.h
ruiling: I would rather keep the uses replacement out of this helper function for simplicity. Caller usually need to do some more work on the arguments. So they can just handle the replacement with other work in one loop.

ruiling: Btw, I have updated the comments per your suggestion.

lgc/patch/Continufy.cpp
ruiling: The `getLevelOfRtStage()` will be later moved to some other common place like `shared/continuation`, the `MDNode` handling is quite specific to this pass. I think passing integer is more proper.

lgc/patch/Continufy.cpp
ruiling: I don't see why we want to let `getLevelOfRtStage` calculate shifted level. We only need to shift the level when we want to generate the mask. We also need the un-shifted level.

lgc/patch/Continufy.cpp
ruiling: I think the suggestion is for `getLevelOfRtStage`?

ruiling: I have slightly updated the comment, please check whether it sounds reasonable.

lgc/patch/Continufy.cpp
jasilvanus: Thanks. I agree it's not an issue now. Still I think it is bad style (being aware this pattern is quite common though), and with more complicated passes this becomes more difficult to keep track of.

If we consistently adopted the pattern of using implementation classes for passes, and keeping only state in the pass class that is intended to stay alive, this issue would just go away.

amdvlk-admin:
### Test summary for commit 85fc60adb682de1a8ba84adcceb2f609623f1fd7
<details><summary>CTS tests (Failed: 0/209067)</summary><ul><li>Built with version 1.3.3.1</li>

</ul></details>

<li>Passed: 38620/69504 (55.6%)</li>
<li>Not Supported: 30884/69504 (44.4%)</li><li>Warnings: 0/69504 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,204 @@
...
+  switch (rtStage) {
+  case RtStage::ShaderStageRayGeneration:
+    llvm_unreachable("Raygen shader should not arrive here.");
+  case RtStage::ShaderStageClosestHit:
+  case RtStage::ShaderStageMiss:
+    // Traversal
+    return (1u << (unsigned)CpsLevel::Traversal);
+  case RtStage::ShaderStageCallable:
+    // CHS/Miss/Callable
+    return (1u << (unsigned)CpsLevel::ClosestHit_Miss_Callable);
```
Flakebi, line 87: RayGen can also call a callable shader.

Flakebi (commented):
Thanks for all the fixes!

lgc/patch/Continufy.cpp
ruiling: I agree resetting member variables at the end of run() is a good idea we should follow, but not sure about having implementation class for each pass. I just realized I don't need the PipelineState. I have removed it.

amdvlk-admin:
### Test summary for commit 3e21d6c04259f7494a90938bb0a752d078a9e058

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,202 @@
...
+  newFn->getArg(0)->setName("state");
+  newFn->getArg(1)->setName("rcr");
+  return newFn;
+}
+
+/// Return the CPS levels mask of the ray-tracing stages that the input stage will return to.
+static unsigned getReturnedLevels(int stage) {
+  // Traversal will return to RGS.
+  if (stage == -1)
+    return 1u << (unsigned)CpsLevel::RayGen;
```
nhaehnle, line 75: Traversal can also return to CHS / miss.

lgc/test/Transforms/Continufy/simple.ll
```diff
@@ -0,0 +1,92 @@
...
+;
+;
+; CHECK-LABEL: define {{[^@]+}}@chs
+; CHECK-SAME: ({} [[STATE:%.*]], i32 [[RCR:%.*]], i32 [[X:%.*]]) !lgc.shaderstage !2 !continufy.stage !4 !lgc.cps !5 {
+; CHECK-NEXT:    [[PUSHCONST:%.*]] = call ptr addrspace(4) @lgc.user.data(i32 24)
+; CHECK-NEXT:    [[FN:%.*]] = load ptr, ptr addrspace(4) [[PUSHCONST]], align 8
+; CHECK-NEXT:    [[TMP1:%.*]] = ptrtoint ptr [[FN]] to i32
+; CHECK-NEXT:    [[TMP2:%.*]] = or i32 [[TMP1]], 1
+; CHECK-NEXT:    [[TMP3:%.*]] = call i32 (...) @lgc.cps.await.i32(i32 [[TMP2]], i32 2, i32 [[X]])
+; CHECK-NEXT:    call void (...) @lgc.cps.jump(i32 [[RCR]], i32 5, {} poison, i32 poison, i32 [[TMP3]])
```
nhaehnle, line 64: Why is `levels = 5` here? Shouldn't it be 4 according to getReturnedLevels?

nhaehnle (commented):
I'm still confused by something... see inline.

lgc/test/Transforms/Continufy/simple.lgc
```diff
@@ -0,0 +1,92 @@
...
+; CHECK-NEXT:    br label [[EXIT]]
+; CHECK:       exit:
+; CHECK-NEXT:    ret void
+;
+;.
+; CHECK: attributes #[[ATTR0:[0-9]+]] = { nounwind willreturn }
+; CHECK: attributes #[[ATTR1:[0-9]+]] = { noreturn }
+;.
+; CHECK: [[META0:![0-9]+]] = !{!""}
+; CHECK: [[META1:![0-9]+]] = !{!"\82\B0amdpal.pipelines\91\82\B0.spill_threshold\CD\FF\FF\B0.user_data_limit\00\AEamdpal.version\92\03\00"}
```
ruiling, line 87: I noticed there is metadata difference when generating the test checks against internal/external llpc. It has extra '\AA.registers\80' when using the internal llpc. Does anybody know why?

lgc/test/Transforms/Continufy/simple.ll
ruiling: I made some mistake to use `.ll` postfix as file name, and we don't include such test in "lgc/test/lit.cfg.py", I did not notice that the test was not run through `ninja check-lgc`. I am not sure whether it sounds reasonable to also include `.ll` test since we are using LLVM IR file for the test?

amdvlk-admin:
### Test summary for commit b937bc02d4c4f74e0137d228b8344c80c6332d15

</ul></details>

</ul></details>

</ul></details>


</ul></details>

lgc/patch/Continufy.cpp
```diff
@@ -0,0 +1,202 @@
...
+    return 1u << (unsigned)CpsLevel::RayGen | 1u << (unsigned)CpsLevel::ClosestHit_Miss_Callable;
+
+  RtStage rtStage = static_cast<RtStage>(stage);
+  switch (rtStage) {
+  case RtStage::ShaderStageRayGeneration:
+    llvm_unreachable("Raygen shader should not arrive here.");
+  case RtStage::ShaderStageClosestHit:
+  case RtStage::ShaderStageMiss:
+    // Traversal
+    return (1u << (unsigned)CpsLevel::Traversal);
```
jasilvanus, line 84: Aren't we directly jumping back to RGS/CHS/Miss from CHS and Miss?

When calling CHS/Miss, Traversal obtains the RGS/CHS/Miss return address from system data, and passes it to CHS/Miss.

Flakebi: In the real continuations workflow, yes. With the continufy pass, no. It will keep the “real” function stack of
RayGen → TraceRay → CHS/Miss
and return the same way.

jasilvanus: Oh, I see. That warrants a comment though :)

ruiling: I am not sure what kind of comment would be more helpful here. I just added some comment to help people be more aware of the usage of Continufy pass.

amdvlk-admin:
### Test summary for commit afd86b207b806fe5a9005a1ae70814cec5a6fe15

</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (approved):

PR #2678, opened 2023-09-05
llpc: Remove manipulation of FAdd fast math flags

perlfu:
Remove code which attempts to clear contract flag from FAdd operations if any operand (recursively) has a NoContraction decoration.

Removal on the basis of:
1. code is quite old and has no lit tests or clear motivation.
2. it is applied too late to effect all relevant manipulations.
3. implementation of isOperandNoContract was incorrect such that it would not properly recurse other than the first operand. Which means any manipulation of flags was partial at best.

Analysis suggests this change only affects one or two older game titles.  If this is required then it should be added back as tuning enable workaround.
I plan to add support for this in a follow up patch.

### DIFF
--- a/llpc/lower/llpcSpirvLowerMath.cpp
+++ b/llpc/lower/llpcSpirvLowerMath.cpp
@@ -101,27 +101,6 @@ void SpirvLowerMath::flushDenormIfNeeded(Instruction *inst) {
 }
 
 // =====================================================================================================================
-// Recursively finds backward if the FPMathOperator operand does not specify "contract" flag.
-//
-// @param operand : Operand to check
-bool SpirvLowerMath::isOperandNoContract(Value *operand) {
-  if (isa<BinaryOperator>(operand)) {
-    auto inst = dyn_cast<BinaryOperator>(operand);
-
-    if (isa<FPMathOperator>(operand)) {
-      auto fastMathFlags = inst->getFastMathFlags();
-      bool allowContract = fastMathFlags.allowContract();
-      if (fastMathFlags.any() && !allowContract)
-        return true;
-    }
-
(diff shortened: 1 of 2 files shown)


amdvlk-admin:
### Test summary for commit a37107e67951cf136142919361c0e3b15428094c
<details><summary>CTS tests (Failed: 0/209067)</summary><ul><li>Built with version 1.3.3.1</li>

</ul></details>

<li>Passed: 38620/69504 (55.6%)</li>
<li>Not Supported: 30884/69504 (44.4%)</li><li>Warnings: 0/69504 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

nhaehnle:
I believe the motivation for this may have been that SPIR-V is defined such that if we have a sequence of OpFMul + OpFAdd with a NoContract only on the OpFMul, we're not supposed to contract this to a fused multiply-add; but LLVM has (had?) different opinions. Given that, I'm a bit hesitant to just outright remove this without a replacement.

nhaehnle:
Yes, agreed. The code as-is definitely looks suspicious so I agree with you some change makes sense. I just doubt we can remove it wholesale.

perlfu:
Having looked at this in detail I am now strongly convinced this change is valid.

LLVM will never combine fadd (e.g. to fma) if an operand does not have contract flag.
From testing, the original code only has any impact if recursion is enabled (exactly because of the above).
In all the examples effected by this change, the resulting ISA is still accurate to the original SPIRV/GLSL intent.

The common pattern the original code is acting on this is the following:
```
precise float r0 = ...
output0 = r0
float r1 = (r0 * x) + c
output1 = r1
```

Where NoContraction is only requested on the first operation (r0), the existing code extends it to the second (r1).

perlfu:
Supersede this with #2687

PR #2687, opened 2023-09-08
llpc: Add support for propagating NoContract fast math flags

perlfu:
Replace current forward propagation of NoContract flags to FAdd operations with a more complete and efficient implementation.

Add support to backward propagation of NoContract flags. This is required for some applications where SPIRV NoContract decorations are incomplete.

Control both of the above from PipelineShaderOptions.


amdvlk-admin:
### Test summary for commit ef8a19fc716dc204debc10931a6fbe723a11ee7b

</ul></details>

</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerMath.cpp
```diff
@@ -295,16 +283,154 @@ bool SpirvLowerMathPrecision::runImpl(Module &module) {
...
+        // Only propagate through instructions
+        if (auto userInst = dyn_cast<Instruction>(user)) {
+          if (visited.count(userInst))
+            continue;
+          // Only update FAdd instructions
+          if (userInst->getOpcode() == Instruction::FAdd) {
+            if (clearContractFlag(userInst))
+              changed = true;
+          }
+          worklist.push_back(userInst);
```
nhaehnle, line 365: Do we really need to propagate further than the first `fadd`?

llpc/lower/llpcSpirvLowerMath.cpp
```diff
@@ -295,16 +283,154 @@ bool SpirvLowerMathPrecision::runImpl(Module &module) {
...
+  // Backward propagate via operands
+  if (backward) {
+    LLVM_DEBUG(dbgs() << "backward propagate no contract\n");
+    worklist = roots;
+    while (!worklist.empty()) {
+      auto inst = worklist.pop_back_val();
+      LLVM_DEBUG(dbgs() << "visit: " << *inst << "\n");
+      for (Value *operand : inst->operands()) {
+        if (auto opInst = dyn_cast<Instruction>(operand)) {
+          if (visited.count(opInst))
```
nhaehnle, line 337: Silly little trick: use `!visited.insert(opInst).second` -- this avoids a redundant set lookup (also applies below)

llpc/lower/llpcSpirvLowerMath.cpp
perlfu: Yes, I believe we do if our goal is to replicate the previous code.

The previous code ran in an instruction visitor and as such it could strip _contract_ from one FAdd (a), then find that FAdd (a) as a input to a later FAdd (b).
With the caveat that the instruction visitor progresses in program order, so it is possible that (b) could be reached before (a), yielding a different result depending on program structure.
This implementation works independent of program order.

That said, I am not against changing this to first FAdd only -- if we are happy in the change in behaviour.

llpc/lower/llpcSpirvLowerMath.cpp
perlfu: Thanks! Implemented.

amdvlk-admin:
### Test summary for commit aca633901b52cc56ccd08a2ded22a1281b39d906

</ul></details>

</ul></details>

</ul></details>


</ul></details>

nhaehnle (approved):
Okay, it makes sense to be conservative with this change, but I think it's worth exploring a change to the forward propagation -- precisely because at least AFAIU there's an in-spec reason to do *some* forward propagation, but only by one step. Getting closer to what the spec actually requires is worthwhile.

perlfu:
@nhaehnle Thank you for the approval.

Just to confirm before I merge this -- the forward propagation in this patch is disabled by default.
My intention is to make a driver (XGL) change to enable it by default -- I am not sure if this detail was obvious in the review.
Are you OK with that change?

nhaehnle:
I missed that and it should technically be another version major bump. On the other hand, the major version was just bumped. Bit dodgy, but I think still okay.

PR #2694, opened 2023-09-11
test: Allow .ll test file

ruiling:
As we used .ll file format, it is reasonable to include .ll test in lgc.


Flakebi (approved):
Thank you!

nhaehnle (approved):

PR #2712, opened 2023-09-19
Fix tests to pass after LLVM update.

kosarev:


kosarev:
@dstutt Hi David, it seems I don't have commit access for this repo (likely because of long inactivity?), so can't even add reviewers.

dstutt (approved):
I think this is ok in this case since there's no auto-update being used.
I suspect there will at some point, but I guess by then the llvm version will have propagated.

dstutt:
> @dstutt Hi David, it seems I don't have commit access for this repo (likely because of long inactivity?), so can't even add reviewers.

Ok - I'll merge this for you when it passes.

dstutt:
test this please

PR #2739, opened 2023-10-02
Lower OpFNegate directly to fneg

dstutt:
Update SPIRVReader to lower OpFNegate directly to fneg. By doing this the code in SpirvLowerMath can be updated to remove the change to add a canonicalize for any fneg.

The motivation for this is that InstCombine can introduce fneg instructions (after some recent upstream llvm changes) and in this case we don't want to blindly add a canonicalize as this invalidates e.g. xor 0x8000000 val -> fneg val.

It makes more sense to handle the requirement to treat OpFNegate as a floating point instruction at the point it is lowered (llvm fneg is not, hence the reason we may need the canonicalize intrinsic).

### DIFF
--- a/llpc/lower/llpcSpirvLowerMath.cpp
+++ b/llpc/lower/llpcSpirvLowerMath.cpp
@@ -476,15 +476,6 @@ bool SpirvLowerMathFloatOp::runImpl(Module &module) {
 }
 
 // =====================================================================================================================
-// Visits unary operator instruction.
-//
-// @param unaryOp : Unary operator instruction
-void SpirvLowerMathFloatOp::visitUnaryOperator(UnaryOperator &unaryOp) {
-  if (unaryOp.getOpcode() == Instruction::FNeg)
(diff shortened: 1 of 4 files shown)


amdvlk-admin:
### Test summary for commit 8aaf5a4c7ab49d89ad9379d613dcbf660e79e204
<details><summary>CTS tests (Failed: 0/139021)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 41144/69517 (59.2%)</li>
<li>Failed: 0/69517 (0.0%)</li>
<li>Not Supported: 28373/69517 (40.8%)</li><li>Warnings: 0/69517 (0.0%)</li>
</ul></details>

<li>Passed: 41178/69504 (59.2%)</li>
<li>Not Supported: 28326/69504 (40.8%)</li><li>Warnings: 0/69504 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (approved):
LGTM

PR #2758, opened 2023-10-12
lgc: Optimally insert return block during cps lowering

ruiling:
If the original function does not have `return` instruction, we don't need the runtime check to exit, thus no need to insert return block.

A handy command to fixup the test check:
sed -i 's/llvm.amdgcn.set.inactive.i32/llvm.amdgcn.set.inactive.{{(chain.arg.)?}}i32/g' lgc/test/Transforms/CpsLowering/*

### DIFF
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -402,8 +402,6 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func, ShaderInputs *shaderInpu
   //      ret void
   unsigned waveSize = m_pipelineState->getShaderWaveSize(m_shaderStage);
   Type *waveMaskTy = builder.getIntNTy(waveSize);
-  auto *chainBlock = BasicBlock::Create(func->getContext(), "chain.block", func);
-  auto *retBlock = BasicBlock::Create(func->getContext(), "ret.block", func);
   // For continufy based continuation, the vgpr list: LocalInvocationId, vcr, vsp, ...
   unsigned vcrIndexInVgpr = shaderInputs ? 1 : 0;
   auto *vcr = builder.CreateExtractValue(vgprArg, vcrIndexInVgpr);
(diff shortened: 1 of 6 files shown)


amdvlk-admin:
### Test summary for commit 7a995d48fe533693707d2a1414613fe64ac9b52d
<details><summary>CTS tests (Failed: 0/139021)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 41144/69517 (59.2%)</li>
<li>Failed: 0/69517 (0.0%)</li>
<li>Not Supported: 28373/69517 (40.8%)</li><li>Warnings: 0/69517 (0.0%)</li>
</ul></details>

<li>Passed: 41177/69504 (59.2%)</li>
<li>Not Supported: 28327/69504 (40.8%)</li><li>Warnings: 0/69504 (0.0%)</li>
</ul></details>


</ul></details>

Flakebi (approved):

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -437,13 +435,21 @@ bool PatchEntryPointMutate::lowerCpsOps(Function *func, ShaderInputs *shaderInpu
...
+    auto *isNullTarget = builder.CreateICmpEQ(targetVcr, builder.getInt32(0));
+    builder.CreateCondBr(isNullTarget, retBlock, chainBlock);
+
+    builder.SetInsertPoint(retBlock);
+    builder.CreateRetVoid();
+  }
 
-  builder.SetInsertPoint(chainBlock);
+  if (chainBlock)
+    builder.SetInsertPoint(chainBlock);
```
nhaehnle, line 452: This could be moved in the `îf`above, right?

nhaehnle (commented):
One nit, but LGTM otherwise.

lgc/patch/PatchEntryPointMutate.cpp
ruiling: Yes, it can, I just move this out of the above `if` for the sake of clear code (compiler should be able to simplify the branch here), this line should better stay with the code below not the above.

PR #2775, opened 2023-10-19
lgc: Workaround crash issue with stop-after option

ruiling:
The issue was first observed by running:
`amdllpc -gfxip=10.3 -stop-after=lgc-patch-resource-collect llpc/test/shaderdb/general/PipelineGsTess_TestInOutPacking.pipe -o -`.

Based on my limited experiment, there are three key factors to trigger this issue:
1.) There is a `stop-after` option
2.) the stopped after name points to a pass before optimization 3.) The input IR to lgc stage have a loop

Using `lgc` command can also reproduce the issue by generating the input `.lgc` through amdllpc first.

Based on my observation, the issue happens mainly because the `InnerAnalysisManagerProxy::Result::~Result()` try to clear() the things in the corresponding InnerAnalysisManager(aka `InnerAM`).

For our specific case, the LoopAnalysisManager (in derived class) was destructed before the ModuleAnalysisManager (in base class).  After the LoopAnalysisManager was destroyed, there is still an `AnalysisResultModel` object in `AnalysisResults` of FunctionAnalysisManager holding pointer to the LoopAnalysisManger (stale now!), and which was transitively in the ModuleAnalysisManager. When ModuleAnalysisManager is destroyed, it will try to clear() the `AnalysisResults` in LoopAnalysisManager because of the outdated `InnerAM` pointer to the LoopAnalysisManager, but the LoopAnalysisManager itself and all its memembers are freed. The problem should have more deeper reason, but it is hard for me to figure out why it was designed like this.

I also did some interesting experiment and it shows the problem is not specific to LoopAnalysisManager, but also applies to FunctionAnalysisManager. If we change the order of FunctionAnalysisManger and ModuleAnalysisManager (that is change their destruction order) in lgc::PassManager, we can also see the crash running: `lgc -stop-after=lgc-patch-buffer-op -o - lgc/test/ScalarizeInputWithDynamicIndexUser.lgc`.

Another experiment I did to change the declaration order of the anlysis managers in llvm/tools/opt/NewPMDriver.cpp. Crash can also be observed running `check-llvm`.

I think there should be more robust solution to fix the issue properly. But I am completely unsure what it would be look like. So I post this change to unblock the usage of `stop-after` on certain early patch passes for our lit-testing.

The `CGSCCAnalysisManager` order change is not mandatory for the issue I see, but I just change it to be consistent.

### DIFF
--- a/lgc/interface/lgc/PassManager.h
+++ b/lgc/interface/lgc/PassManager.h
@@ -76,7 +76,11 @@ public:
   virtual llvm::PassInstrumentationCallbacks &getInstrumentationCallbacks() = 0;
 
 protected:
+  // NOTE: The analysis managers are intentionally placed in inner->outer order so that the outmost analysis is
+  // destructed first, this help workaround crash when stop-after is used.
+  llvm::LoopAnalysisManager m_loopAnalysisManager;
   llvm::FunctionAnalysisManager m_functionAnalysisManager;
+  llvm::CGSCCAnalysisManager m_cgsccAnalysisManager;
   llvm::ModuleAnalysisManager m_moduleAnalysisManager;
 };
 
--- a/lgc/util/PassManager.cpp
+++ b/lgc/util/PassManager.cpp
@@ -115,8 +115,6 @@ private:
 
   // -----------------------------------------------------------------------------------------------------------------
 
-  LoopAnalysisManager m_loopAnalysisManager;               // Loop analysis manager used when running the passes.
-  CGSCCAnalysisManager m_cgsccAnalysisManager;             // CGSCC analysis manager used when running the passes.
   PassInstrumentationCallbacks m_instrumentationCallbacks; // Instrumentation callbacks ran when running the passes.
   StandardInstrumentations m_instrumentationStandard;      // LLVM's Standard instrumentations
   unsigned *m_passIndex = nullptr;                         // Pass Index.


amdvlk-admin:
### Test summary for commit a36164f088bdf68b9d14daded4962f7e871fb5a9
<details><summary>CTS tests (Failed: 0/138994)</summary><ul><li>Built with version 1.3.5.2</li>

</ul></details>

<li>Passed: 41178/69491 (59.3%)</li>
<li>Failed: 0/69491 (0.0%)</li>
<li>Not Supported: 28313/69491 (40.7%)</li><li>Warnings: 0/69491 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit d1c369754694c427e8e206b00bc8d73ed2d5ef56
<details><summary>CTS tests (Failed: 0/138995)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 41144/69504 (59.2%)</li>
<li>Not Supported: 28360/69504 (40.8%)</li><li>Warnings: 0/69504 (0.0%)</li>
</ul></details>

<li>Passed: 41178/69491 (59.3%)</li>
<li>Failed: 0/69491 (0.0%)</li>
<li>Not Supported: 28313/69491 (40.7%)</li><li>Warnings: 0/69491 (0.0%)</li>
</ul></details>


</ul></details>

trenouf (approved):
I don't know if we are doing something wrong here, or if the LLVM pass manager is doing something wrong. But this seems like a reasonable workaround, so LGTM.

GinShio:
Hi, I think this is a coding error on our.

We can found documents for [InnerAnalysisManagerProxy](https://llvm.org/doxygen/classllvm_1_1InnerAnalysisManagerProxy.html#details) and [OuterAnalysisManagerProxy](https://llvm.org/doxygen/classllvm_1_1OuterAnalysisManagerProxy.html#details).

> Note that the proxy's result is a move-only RAII object. The validity of the analyses in the inner analysis manager is tied to its lifetime. 

Looks like it was intentionally designed like this, we have to guarantee these manager's lifetime. That is, inner manager must be live the longer than outer.

In C++, initialization order is the order they appear in the class definition. [[class.base.init-13](https://timsong-cpp.github.io/cppwp/n4659/class.init#class.base.init-13)]

>The declaration order is mandated to ensure that base and member subobjects are destroyed in the reverse order of initialization.

In LLVM, the dependencies of these managers are defined.
```C++
typedef InnerAnalysisManagerProxy<LoopAnalysisManager, Function> LoopAnalysisManagerFunctionProxy;
using CGSCCAnalysisManagerModuleProxy = InnerAnalysisManagerProxy<CGSCCAnalysisManager, Module>;
using CGSCCAnalysisManagerFunctionProxy = OuterAnalysisManagerProxy<CGSCCAnalysisManager, Function>;
using FunctionAnalysisManagerModuleProxy = InnerAnalysisManagerProxy<FunctionAnalysisManager, Module>;
```

Does it mean the correct order is loop, function, cgscc, and module?

nhaehnle (approved):
Yeah, this is a bit of a strange / dangerous design. But the comment you added helps. Let's do this.

PR #2782, opened 2023-10-25
Fix signed zero issues of FRem

amdrexu:
We use IR builder CreateFRem to translate OpFRem to LLVM native frem instruction. The instruction is further lowered by backend following such formula:

`  frem(x, y) = x - y * trunc(x/y)`

There is a latent issue when x=-0.0. According to SPIR-V spec, the sign of the result of FRem is the same as the sign of the dividend. But when we input x=-0.0 to above formula, we finally get the addition of (-0.0) + 0.0. The result is +0.0 returned by HW. Hence, we have to manually check this special case when nsz fast math flag is not specified.

### DIFF
--- a/lgc/builder/ArithBuilder.cpp
+++ b/lgc/builder/ArithBuilder.cpp
@@ -236,8 +236,7 @@ Value *BuilderImpl::CreateSMod(Value *dividend, Value *divisor, const Twine &ins
 }
 
 // =====================================================================================================================
-// Create FP modulo operation, where the sign of the result (if not zero) is the same as the sign
-// of the divisor.
+// Create FP modulo operation, where the sign of the result is the same as the sign of the divisor.
 //
 // @param dividend : Dividend value
 // @param divisor : Divisor value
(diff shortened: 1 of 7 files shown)


amdvlk-admin:
### Test summary for commit fc12f6aac9968961b3f0e43118a1b9ab7c4f0fc6
<details><summary>CTS tests (Failed: 0/138994)</summary><ul><li>Built with version 1.3.5.2</li>

</ul></details>

<li>Passed: 41177/69491 (59.3%)</li>
<li>Failed: 0/69491 (0.0%)</li>
<li>Not Supported: 28314/69491 (40.7%)</li><li>Warnings: 0/69491 (0.0%)</li>
</ul></details>


</ul></details>

amdvlk-admin:
### Test summary for commit efec18ecee29c7c9ae493eb04a0124ab0a78d0b9
<details><summary>CTS tests (Failed: 0/137949)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 35352/68947 (51.3%)</li>
<li>Failed: 0/68947 (0.0%)</li>
<li>Not Supported: 33595/68947 (48.7%)</li><li>Warnings: 0/68947 (0.0%)</li>
</ul></details>

<li>Passed: 35423/69002 (51.3%)</li>
<li>Failed: 0/69002 (0.0%)</li>
<li>Not Supported: 33579/69002 (48.7%)</li><li>Warnings: 0/69002 (0.0%)</li>
</ul></details>


</ul></details>

nhaehnle (commented):
I just looked into LLVM LangRef.rst. It says this about `frem`:

> The value produced is the floating-point remainder of the two operands. This is the same output as a libm '``fmod``' function, but without any possibility of setting ``errno``. **The remainder has the same sign as the dividend.**

It seems like we should probably consider this an LLVM backend bug that must be fixed there instead of in LLPC?

cc @kezhaoAMD This might affect some of your stuff as well.

amdrexu:
> I just looked into LLVM LangRef.rst. It says this about `frem`:
> 
> > The value produced is the floating-point remainder of the two operands. This is the same output as a libm '`fmod`' function, but without any possibility of setting `errno`. **The remainder has the same sign as the dividend.**
> 
> It seems like we should probably consider this an LLVM backend bug that must be fixed there instead of in LLPC?
> 
> cc @kezhaoAMD This might affect some of your stuff as well.

I created an issue ticket to LLVM backend team. If this could be resolved quickly, I will drop this change. Otherwise, I will keep it as a temporary workaround and remove it once backend fix lands.

amdrexu:
Thanks @jayfoad. I close this PR since a better fix is in LLVM itself: https://github.com/llvm/llvm-project/pull/70448

jayfoad:
> According to SPIR-V spec, the sign of the result of FRem is the same as the sign of the dividend. But when we input x=-0.0 to above formula, we finally get the addition of (-0.0) + 0.0. The result is +0.0 returned by HW.

The SPIR-V spec (https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpFRem) only specifies the sign of the result when the result is not zero, so this case is OK.

PR #2789, opened 2023-10-28
Fix caching and PAL metadata for color export shaders

nhaehnle:
The second commit fixes a regression from a recent cleanup.

The core underlying design problem is that the PipelineState from the outer pipeline compile (the one that requests the glue shader) leaks into the generation of glue shaders. This is a problem because glue shaders can be cached.

The more immediate problem in the regression was that instead of writing color export-related PAL metadata into the glue shader ELF, it was written into the metadata for3 the overall pipeline compile, which meant that it wasn't available in the cache. This previously worked because the PAL metadata was recomputed each time, but that required duplicating the logic and/or keeping it separate from where the actual exports instructions are created, which makes for hard-to-maintain code (basically, two very separate pieces of code would need to be kept in sync).

The solution is to write the PAL metadata into the glue shader ELF.

The first commit is a precaution against likely future regressions with PRs to optimize exports from the color export shader based on pieces of pipeline state.

### DIFF
--- a/lgc/elfLinker/ColorExportShader.cpp
+++ b/lgc/elfLinker/ColorExportShader.cpp
@@ -116,14 +116,19 @@ Module *ColorExportShader::generate() {
     values[m_exports[idx].hwColorTarget] = colorExportFunc->getArg(idx);
   }
 
+  PalMetadata palMetadata{m_pipelineState, m_pipelineState->useRegisterFieldFormat()};
+
   bool dummyExport = m_lgcContext->getTargetInfo().getGfxIpVersion().major < 10 || m_killEnabled;
-  fragColorExport.generateExportInstructions(m_exports, values, dummyExport, builder);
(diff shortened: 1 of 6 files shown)


amdvlk-admin:
### Test summary for commit a34a24206e6881f4ef3ce29ee5d7273c31f2dcc7
<details><summary>CTS tests (Failed: 0/138020)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 35401/69018 (51.3%)</li>
<li>Failed: 0/69018 (0.0%)</li>
<li>Not Supported: 33617/69018 (48.7%)</li><li>Warnings: 0/69018 (0.0%)</li>
</ul></details>

<li>Passed: 35424/69002 (51.3%)</li>
<li>Failed: 0/69002 (0.0%)</li>
<li>Not Supported: 33578/69002 (48.7%)</li><li>Warnings: 0/69002 (0.0%)</li>
</ul></details>


</ul></details>

amdrexu (approved):

PR #2802, opened 2023-11-07
lgc: add dialect GroupMemcpyOp

xazhangAMD:


amdvlk-admin:
### Test summary for commit da88fdcb1d14e651850d831e51f8ba9ff4d117e5
<details><summary>CTS tests (Failed: 0/138184)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 35162/69058 (50.9%)</li>
<li>Failed: 0/69058 (0.0%)</li>
<li>Not Supported: 33896/69058 (49.1%)</li><li>Warnings: 0/69058 (0.0%)</li>
</ul></details>

<li>Passed: 35242/69126 (51.0%)</li>
<li>Failed: 0/69126 (0.0%)</li>
<li>Not Supported: 33884/69126 (49.0%)</li><li>Warnings: 0/69126 (0.0%)</li>
</ul></details>


</ul></details>

linqun (approved):

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -254,6 +259,203 @@ static Value *mergeDwordsIntoVector(IRBuilder<> &builder, ArrayRef<Value *> inpu
   return vec;
 }
 
+// =====================================================================================================================
+// Lower GroupMemcpyOp
+void PatchEntryPointMutate::processGroupMemcpy(Module &module) {
+  SmallVector<CallInst *> tobeErased;
```
amdrexu, line 265: tobeErased -> toBeErased

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -254,6 +259,203 @@ static Value *mergeDwordsIntoVector(IRBuilder<> &builder, ArrayRef<Value *> inpu
...
+    // LocalInvocationId is a function argument now and CreateReadBuiltInInput cannot retrieve it.
+    unsigned argIndex = 0xFFFFFFFF;
+    switch (stage) {
+    case ShaderStageTask: {
+      auto &entryArgIdxs = m_pipelineState->getShaderInterfaceData(ShaderStageTask)->entryArgIdxs.task;
+      argIndex = entryArgIdxs.localInvocationId;
+      break;
+    }
+    case ShaderStageMesh: {
+      auto &entryArgIdxs = m_pipelineState->getShaderInterfaceData(ShaderStageTask)->entryArgIdxs.mesh;
```
amdrexu, line 328: -> getShaderInterfaceData(ShaderStageMesh)

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -254,6 +259,203 @@ static Value *mergeDwordsIntoVector(IRBuilder<> &builder, ArrayRef<Value *> inpu
...
+      argIndex = entryArgIdxs.localInvocationId;
+      break;
+    }
+    case ShaderStageCompute: {
+      auto &entryArgIdxs = m_pipelineState->getShaderInterfaceData(ShaderStageTask)->entryArgIdxs.cs;
```
amdrexu, line 333: -> getShaderInterfaceData(ShaderStageCompute)

lgc/patch/PatchEntryPointMutate.cpp
```diff
@@ -254,6 +259,203 @@ static Value *mergeDwordsIntoVector(IRBuilder<> &builder, ArrayRef<Value *> inpu
...
+                            })
+                            .build();
+  visitor.visit(payload, module);
+  for (auto call : payload.tobeErased)
+    call->eraseFromParent();
+}
+
+// =====================================================================================================================
+// Lower GroupMemcpyOp - Copy memory using threads in a workgroup (scope=2) or subgroup (scope=3).
+void PatchEntryPointMutate::lowerGroupMemcpy(GroupMemcpyOp &groupMemcpyOp) {
```
amdrexu, line 286: I don't think it is a good idea to place the handling of group memcpy here because the pass is aimed to handle entry-point mutation. Other responsibilities should be moved to other passes or even creating a new pass according to LLVM design philosophy. 

The handling of task/mesh shader could be put in MeshTaskShader.cpp. For CS, many operations are straightforward, if possible, maybe we can place it on InOutBuilder since readCsBuiltIn() can read back any CS built-in so you can do anything you want. This is true for task shader as well and share the handling. If that is impossible, we can move the handling of CS to PatchInOutImportExport.

xazhangAMD: I had a similar thought in the internal review - split the op into 2: one for task/mesh and another for compute.

amdrexu: You can check shader stage when lowering this op so we can differentiate its usages in task/mesh shader or in compute shader. If you decide to make two dedicated ops, that is fine as well. 

I discussed this with Ruiling, we both believe the lowering of this op is better to be moved to other appropriate passes other than this pass. Also, if you can share us with a LGC file (.lgc generated by frontend with the option --emit-lgc) showing the usage of this op we can better evaluate your future refactoring change in the review.

xazhangAMD: It is natural to handle task/mesh shader in MeshTaskShader.cpp but for CS, it is also weird to place the code in InOutBuilder or PatchInOutImportExport. Actually my only intent is to use this for task shader only. The llpcfe standalone tool doesn't seem to support -emit-lgc but the dump should have all the information you need.

I modified code in a commit [here](https://github.com/GPUOpen-Drivers/llpc/commit/21d89cb2c2c4872773ca8b0aceb3e843d476e468).

And a pipeline dump attached.
[PipelineTaskMesh_0xB812ED624A368A8F.txt](https://github.com/GPUOpen-Drivers/llpc/files/13310830/PipelineTaskMesh_0xB812ED624A368A8F.txt)

amdrexu: Thank you. I see the usage. Your requirement is similar to the usage of PatchInitializeWorkgroupMemory::initializeWithZero. We add a new pass to handle this. Anyway, it is fine to keep CS handling in entry-point mutation as a temporary solution. I plan to rework this pass and try to enable the usage for your case in the near future.

PR #2813, opened 2023-11-10
Fix signed zero issues of FRem

amdrexu:
We use IR builder CreateFRem to translate OpFRem to LLVM native frem instruction. The instruction is further lowered by backend following such formula on our HW:

`  frem(x, y) = x - y * trunc(x/y)`

There is a latent issue when x=-0.0. Although SPIR-V spec says nothing about the case when x = 0.0, C fmod function does specify the sign of the result is the same as that of the dividend even if it is signed zero. But when we input x=-0.0 to above formula, we finally get the addition of (-0.0) + 0.0. The result is +0.0 returned by HW. Hence, we have to manually check this special case when nsz fast math flag is not specified.

### DIFF
--- a/lgc/builder/ArithBuilder.cpp
+++ b/lgc/builder/ArithBuilder.cpp
@@ -236,8 +236,7 @@ Value *BuilderImpl::CreateSMod(Value *dividend, Value *divisor, const Twine &ins
 }
 
 // =====================================================================================================================
-// Create FP modulo operation, where the sign of the result (if not zero) is the same as the sign
-// of the divisor.
+// Create FP modulo operation, where the sign of the result is the same as the sign of the divisor.
 //
 // @param dividend : Dividend value
 // @param divisor : Divisor value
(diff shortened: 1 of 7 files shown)


jasilvanus:
I didn't understand the part of why the behavior of C's `fmod` is relevant here. Could you explain that please?

amdvlk-admin:
### Test summary for commit 5bb8d9317c6d01ed084ee008e1196d041506c536

</ul></details>

</ul></details>


</ul></details>

amdrexu:
> I didn't understand the part of why the behavior of C's `fmod` is relevant here. Could you explain that please?

The SPIR-V spec doesn't specify the sign of the result when handle frem(-0.0, y). But C fmod does say that the sign of the result is the same as that of the dividend. We encounter a test which expects this. So we decide to fix the problem of FRem following C fmod and propose a SPIR-V spec fix to align the behavior with C fmod.

jasilvanus:
What kind of test is that? Does it use SPIR-V, in which case the test case would be broken?

I'm not saying we shouldn't do this, just trying to understand the setting.

amdrexu:
> What kind of test is that? Does it use SPIR-V, in which case the test case would be broken?
> 
It is a new CTS in review. The CTS is hand written SPIR-V assembly test. It tests OpFRem with the input x=-0.0 and y=0.5. According the computation of frem(x,y), the result should be -0.0. The test expects the result to be 0x80000000(-0.0) raw bits but we return 0x00000000(+0.0) so it reports test failure.

Of course, it is controversial because SPIR-V spec doesn't say anything about this when x=0.0. The test author admits this and would like SPIR-V spec to be aligned with C fmod (IEEE) with spec fix. It is reasonable to have SPIR-V consistent with fmod because DirectX has fmod as well.

nhaehnle:
I know there've been internal discussions back and forth on this but I'm still confused about the resolution. LangRef says this about `frem`:

> The remainder has the same sign as the dividend.

... so this really ought to be unnecessary.

amdrexu:
@jayfoad I think we still need a LLVM fix. What's your opinion?

jayfoad:
> @jayfoad I think we still need a LLVM fix. What's your opinion?

Yes, fair enough. I have reopened https://github.com/llvm/llvm-project/pull/70448

amdrexu:
After SPIR-V workgroup discussed this issue. They would like to respect original SPIR-V spec statement, leaving the sign of the result undefined when the dividend is zero. Therefore, we don't have to do anything. Current implementation is fine enough. I will close this PR. Thanks.

PR #2815, opened 2023-11-13
llpc: disable spilling of SGPR spill VGPRs

perlfu:
Add function attribute amdgpu-prealloc-sgpr-spill-vgprs to prevent spilling of SGPR spill VGPRs.
This occurs in shaders with extremely high VGPR pressure, e.g. large RT shaders, and can yield unpredictable results.

This is an interim workaround and should be removed when backend support for these spills is fully implemented.

### DIFF
--- a/lgc/patch/PatchSetupTargetFeatures.cpp
+++ b/lgc/patch/PatchSetupTargetFeatures.cpp
@@ -187,6 +187,10 @@ void PatchSetupTargetFeatures::setupTargetFeatures(Module *module) {
 
     builder.addAttribute("target-features", targetFeatures);
 
+    // Prevent spilling of VGPRs holding SGPR spills as this can have undefined behaviour in callee functions.
+    // Note: this is an intermediate workaround and should be removed when backend support is complete.
+    builder.addAttribute("amdgpu-prealloc-sgpr-spill-vgprs");
+
     func->addFnAttrs(builder);
   }
 }
--- a/lgc/patch/PatchEntryPointMutate.cpp
+++ b/lgc/patch/PatchEntryPointMutate.cpp
@@ -1449,10 +1449,6 @@ void PatchEntryPointMutate::setFuncAttrs(Function *entryPoint) {
   builder.addAttribute("amdgpu-memory-bound", shaderOptions->favorLatencyHiding ? "true" : "false");
   builder.addAttribute("amdgpu-wave-limiter", "false");
 
-  // Prevent spilling of VGPRs holding SGPR spills as this can have undefined behaviour in callee functions.
-  // Note: this is an intermediate workaround and should be removed when backend support is complete.
-  builder.addAttribute("amdgpu-prealloc-sgpr-spill-vgprs");
-
   entryPoint->addFnAttrs(builder);
 
   // NOTE: Remove "readnone" attribute for entry-point. If GS is empty, this attribute will allow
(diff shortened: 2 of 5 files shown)


perlfu:
Use of function attribute to control this means this is safe to merge regardless of backend support.

I have not added any explicit tests as existing tests show change in attributes, but can add a dedicated test if reviewers would prefer.

amdvlk-admin:
### Test summary for commit bfd2d2c215665c54d74017084f6f0dea5922e498

</ul></details>

</ul></details>


</ul></details>

nhaehnle (dismissed):
If the number of SGPR spills becomes truly excessive, does this just fail or does it fallback to some other mechanism of spilling SGPRs?

FWIW, this should probably rather go into PatchSetupTargetFeatures

perlfu:
> If the number of SGPR spills becomes truly excessive, does this just fail or does it fallback to some other mechanism of spilling SGPRs?

As this reserves VGPRs, SGPR spills will reduce the pool of VGPRs and increase VGPR pressure -- eventually VGPR allocation will fail.
In _graphics_ we do not yet have any test cases where this happens, but we do have test cases where the spill/restore of spill VGPRs yield unpredictable results.
Work is progressing to resolve the underlying issues, but there is no clear timeframe yet.

> FWIW, this should probably rather go into PatchSetupTargetFeatures

There does seem to be some odd split between PatchEntryPointMutate and PatchSetupTargetFeatures.
I wouldn't describe this as a "target feature", but then neither are some of the other things in PatchSetupTargetFeatures.
Perhaps there needs to be some rationalisation between the two?

I am indifferent to where this lives, so will move it to PatchSetupTargetFeatures.

amdvlk-admin:
### Test summary for commit ba8f9d8c81b363df80f87669a7741977a6b52d4e
<details><summary>CTS tests (Failed: 0/138443)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 35211/69228 (50.9%)</li>
<li>Failed: 0/69228 (0.0%)</li>
<li>Not Supported: 34017/69228 (49.1%)</li><li>Warnings: 0/69228 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

lgc/patch/PatchSetupTargetFeatures.cpp
```diff
@@ -187,6 +187,10 @@ void PatchSetupTargetFeatures::setupTargetFeatures(Module *module) {
 
     builder.addAttribute("target-features", targetFeatures);
 
+    // Prevent spilling of VGPRs holding SGPR spills as this can have undefined behaviour in callee functions.
```
ruiling, line 190: Thanks for fixing this. Personally I think it maybe helpful to better comment the problem. The problem is the current way to allocate WWM-VGPR used for SGPR spilling during RA is broken when the WWM-VGPR is either spilled or its live-range was split. So we choose to pre-allocate those WWM-VGPRs for SGPR spills. It maybe more proper to say we will re-visit the problem when the backend issue is fixed. I agree SetupTargetFeature is the reasonable place for this attribute.

perlfu: I choose to not describe the exact failure mode of the backend in this comment for brevity.
I felt "undefined behaviour" a sufficiently accurate a description as on a good day it'll all work fine, but on a bad day you get lane clobbering and unexpected behaviour.
The comment on the next line already basically says "remove this later".
If this wasn't clear then I'll reword the comment.

nhaehnle (approved):
Thanks for moving this. I fully agree with your general assessment of PatchEntryPointMutate vs. PatchSetupTargetFeatures. Long-term, I hope we can unify all this into a single pass, but it's a non-trivial refactoring effort to do so.

PR #2817, opened 2023-11-14
llpc: GFX11 flat/custom parameter loads need to be strict WQM

perlfu:
These already use WQM annotation to ensure fetch lanes are active, but this does not work if there are discards or divergence as the WQM exec mask will not truly WQM (some helper invocations killed).

Hence:
- Generate strict.wqm call for flat/custom interpolation loads
- Add test coverage for this code path

Ideally this could be resolved by using the FI bit on the mov_dpp, but there is no backend support for this currently.

### DIFF
--- a/lgc/patch/PatchInOutImportExport.cpp
+++ b/lgc/patch/PatchInOutImportExport.cpp
@@ -1754,9 +1754,10 @@ Value *PatchInOutImportExport::performFsParameterLoad(BuilderBase &builder, Valu
     compValue = builder.CreateIntrinsic(Intrinsic::amdgcn_mov_dpp, builder.getInt32Ty(),
                                         {compValue, builder.getInt32(static_cast<unsigned>(dppCtrl)),
                                          builder.getInt32(15), builder.getInt32(15), builder.getTrue()});
-    // NOTE: Make mov_dpp and its source instructions run in WQM to make sure the mov_dpp could fetch
-    // correct data from possible inactive lanes.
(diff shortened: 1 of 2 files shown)


perlfu:
As noted in the description this could also be solved by adding support for FI bit on the mov_dpp to the backend.
But that will take much longer and not necessarily yield any better results.

The risk with this solution is that it effectively asks for another strict WQM register to be allocated -- potentially wasting a physical VGPR.
However in practice the VGPR allocated for the lds_param_load will only be live until the mov_dpp, so this will always reuse an existing strict WQM register allocation.
This solution will only add register pressure if:
1. the result of the mov_dpp (the parameter) lives for a long time,
2. another lds_param_load or strict WQM could be using reuse the reserved physical register.

amdvlk-admin:
### Test summary for commit bcfbd1c92c9858b03ccdcc727838ea64ddf7fe7a

</ul></details>

</ul></details>


</ul></details>

nhaehnle (approved):
Yeah, the problem with supporting FI is that it's hard to model in IR what that even means...

The assessment of register re-use depends on how instructions are scheduled when there are multiple param loads. But I can't think of a better correctness check either. Ultimately we may have to figure out how to do better register allocation for WWM/WQM registers.

perlfu:
Thanks!

> Ultimately we may have to figure out how to do better register allocation for WWM/WQM registers.

Yes, I am actively working on incremental improvements on this specifically to deal with param loads.

PR #2821, opened 2023-11-16
Mesh shader: Enable row export on GFX11 and fix issues

amdrexu:
- The preparation of attribute ring access should be moved to entry block.
- The function attribute 'amdgpu-flat-work-group-size' is incorrect, which leads to unexpected removal of s_barrier.
- Use primOrVertexIndex to do attribute ring access as the VGPR index of buffer_store. Don't use threadIdInSubgroup because when row export is enabled, threadIdInSubgroup is not always equal to primOrVertexIndex.
- After fixing those issues, enable row export by default on GFX11. This is because mesh query will have to check msInvocations. On GFX11, HW obtains this value from the register field SPI_SHADER_GS_MESHLET_DIM.MESHLET_THREADGROUP_SIZE.

### DIFF
--- a/lgc/patch/MeshTaskShader.cpp
+++ b/lgc/patch/MeshTaskShader.cpp
@@ -479,7 +479,7 @@ void MeshTaskShader::processMeshShader(Function *entryPoint) {
   const unsigned flatWorkgroupSize =
       alignTo(m_pipelineState->enableMeshRowExport() ? numMeshThreads : primAmpFactor, waveSize);
   entryPoint->addFnAttr("amdgpu-flat-work-group-size",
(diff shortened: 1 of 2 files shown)


nhaehnle (approved):
Whoops, I thought that had been done ages ago... I'm surprised this was never caught, are there no CTS tests for mesh shader invocations?

amdrexu:
> I'm surprised this was never caught, are there no CTS tests for mesh shader invocations?

Some issues were not caught because we didn't enable row export for GFX11 and at that time mesh shader CTS were under implementation, changed from time to time. Now, the tests are finalized so we can have a try and enable row export by default.

PR #2854, opened 2023-11-29
Add tests for Deduplicate constant data tables

mariusz-sikora-at-amd:

PR #2858, opened 2023-11-30
Refine the commit f9da39114d

AMD-dwang:
A better way to use Map.

### DIFF
--- a/llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
+++ b/llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
@@ -223,16 +223,12 @@ bool SpirvLowerConstImmediateStore::tryProcessAlloca(AllocaInst *allocaInst) {
   }
 
   // Step 3: Create the global variable and replace the alloca
-  GlobalVariable *global = nullptr;
-  auto iter = m_allocToGlobals.find(initializer);
-  if (iter == m_allocToGlobals.end()) {
+  GlobalVariable *&global = m_allocToGlobals[initializer];
+  if (!global) {
     global = new GlobalVariable(*m_module, allocatedTy,
                                 true, // isConstant
                                 GlobalValue::InternalLinkage, initializer, "", nullptr, GlobalValue::NotThreadLocal,
                                 SPIRAS_Constant);
-    m_allocToGlobals[initializer] = global;
-  } else {
-    global = iter->second;
   }
   global->takeName(allocaInst);
 


amdvlk-admin:
### Test summary for commit 4f0e99224e87e533385eef78f9cd55e2e0d6fca1
<details><summary>CTS tests (Failed: 0/138378)</summary><ul><li>Built with version 1.3.5.2</li>

<li>Passed: 35163/69163 (50.8%)</li>
<li>Failed: 0/69163 (0.0%)</li>
<li>Not Supported: 34000/69163 (49.2%)</li><li>Warnings: 0/69163 (0.0%)</li>
</ul></details>

</ul></details>


</ul></details>

llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
```diff
@@ -223,16 +223,12 @@ bool SpirvLowerConstImmediateStore::tryProcessAlloca(AllocaInst *allocaInst) {
...
-  GlobalVariable *global = nullptr;
-  auto iter = m_allocToGlobals.find(initializer);
-  if (iter == m_allocToGlobals.end()) {
+  GlobalVariable *&global = m_allocToGlobals[initializer];
+  if (!global) {
     global = new GlobalVariable(*m_module, allocatedTy,
                                 true, // isConstant
                                 GlobalValue::InternalLinkage, initializer, "", nullptr, GlobalValue::NotThreadLocal,
                                 SPIRAS_Constant);
-    m_allocToGlobals[initializer] = global;
```
mariusz-sikora-at-amd, line 233: Where is new global added to the map now ?

mariusz-sikora-at-amd: You may run these lit tests (https://github.com/GPUOpen-Drivers/llpc/pull/2854) to verify the change.

jasilvanus: The new code is taking a *reference* to the global pointer stored in the map, so the assignment `global = new...` updates the map.

jayfoad (approved):
Thanks!

llpc/lower/llpcSpirvLowerConstImmediateStore.cpp
```diff
@@ -223,16 +223,12 @@ bool SpirvLowerConstImmediateStore::tryProcessAlloca(AllocaInst *allocaInst) {
   }
 
   // Step 3: Create the global variable and replace the alloca
-  GlobalVariable *global = nullptr;
-  auto iter = m_allocToGlobals.find(initializer);
-  if (iter == m_allocToGlobals.end()) {
+  GlobalVariable *&global = m_allocToGlobals[initializer];
+  if (!global) {
```
Flakebi, line 227: This assumes that if the key `initializer` was not found in `m_allocToGlobals`, a new entry is created and the value of the new entry is `nullptr`, right?

`DensedMap::InsertIntoBucket` creates the new value with `::new (&TheBucket->getSecond()) ValueT(std::forward<ValueArgs>(Values)...);` (where `Values` is empty), so it calls the default constructor of `GlobalVariable*`.
If I am not mistaken, the default constructor of a pointer **does not initialize memory**, so checking for `nullptr` afterwards only works when we are lucky.

`try_emplace(initializer, nullptr)` should do the correct thing.

jayfoad: > But as DenseMap doesn't explicitly document any of this, using it conservatively won't hurt.

My feeling is that DenseMap behaves exactly like std::map in this regard, so it's fine to rely on the value initialization.

Flakebi: That value initialization is a good point. This code looks like it does the correct thing then :)

tsymalla-AMD:
Thanks, but I'd rather edit the commit message (title + message) at least, since it is not really informative without looking up the explicit commit.

Flakebi (approved):

Flakebi:
You can edit the commit title and message when you ‘squash and merge’.

PR #2862, opened 2023-12-04
Fix Doom Eternal RT regression

LLJJDD:
Root cause:
When building ELF with helper thread, a new lgc::Pipeline object is created, which does not set useGpurt.

Solution:
Set useGpurt for that lgc::Pipeline.


nhaehnle (approved):
LGTM

PR #2870, opened 2023-12-07
PatchBufferOp: fix the check condition for PreservedAnalyses

xuechen417:
`impl.run(function)` returns true means we should preserve nothing.


nhaehnle (approved):
Whoops :) Thanks!

PR #2861, opened 2023-12-03
Remove the shader input IsOffChip

amdrexu:
This SGPR was originally for pre-GFX9 when tessellation on-chip mode is available. Now, this SGPR is necessary when merged shader doesn't provide it since tessellation is always in off-chip mode.

PR #2882, opened 2023-12-14
Fix a regression of removing isOffChip input

amdrexu:
This is to fix a regression of
https://github.com/GPUOpen-Drivers/llpc/pull/2861. When TES is mapped to HW VS (no NGG), this SGPR must be present.


chuang13:
@amdrexu

nhaehnle (approved):

PR #2889, opened 2023-12-22
[lgc/llpc][NFC] - Use `starts_with` and `ends_with`

OutOfCache:
`StringRef::startswith` and `StringRef::endswith` are deprecated. See [llvm-project#75491](https://github.com/llvm/llvm-project/pull/75491)


amdvlk-admin:

amdrexu:
I think we have already made such changes in our internal repo. For GPUOpen, just request a code promotion to make codes sync-up. A separate PR for porting could be avoided :)

PR #2917, opened 2025-07-31
Navi44

LloydNicholson:
This is my first time I have written c or c++, so please let me know what else I can do to make this code more robust or any refactors that are required.

Please let me know how I can improve the code. I am under no illusion that this not everything needed for this to work across different 9060 series (gfx1200/navi44) cards but I have it working on mine and that was my goal from the start anyway.

I honestly just wanted to have a better experience with a game I was enjoying and stop enjoying due to an update, so I fixed it to the best of my ability and maybe it can help to get this change into dev.

I did it to the point where I can run Indiana Jones and the Great Circle because it was bugging me that I couldn't and there we some serious performance issues around the save points and I knew AMDVLK could fix the issue.

I haven't done any other meaningful testing other than the fact that my ASRock Challenger 9060 XT 16GB works perfectly so far with AMDVLK in Indiana Jones instead of throwing a VK_ERROR_INCOMPATIBLE_DRIVER.

I added a PR to pal repo as well  -> https://github.com/GPUOpen-Drivers/pal/pull/109

### DIFF
--- a/lgc/state/TargetInfo.cpp
+++ b/lgc/state/TargetInfo.cpp
@@ -408,6 +408,20 @@ static void setGfx1201Info(TargetInfo *targetInfo) {
 
 #endif
 
+#if LLPC_BUILD_NAVI44
+// gfx1200
+//
+// @param [in/out] targetInfo : Target info
+static void setGfx1200Info(TargetInfo *targetInfo) {
+  setGfx12Info(targetInfo);
+
+  targetInfo->getGpuProperty().numShaderEngines = 4;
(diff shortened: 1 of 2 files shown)


jinjianrong:
@LloydNicholson Thanks for your contribution. However, the AMDVLK open source project is discontinued as announced [here](https://github.com/GPUOpen-Drivers/AMDVLK/discussions/416). You are welcome to contribute to [RADV](https://gitlab.freedesktop.org/mesa/mesa/-/tree/main/src/amd/vulkan).

